PR #2, opened 2016-08-19
Misc and static field helper

j3parker:
FYI @travisMiehm 

Need to add tests, putting this up early so we can agree on a rough thing (+ pulling in deps that we'll both need and merge conflict on otherwise...)

Also, lunch time.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .vs
 bin/
 obj/
+packages/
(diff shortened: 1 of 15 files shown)


saquibmian:
This looks good! In terms of doing tests, there's a pretty lightweight way of doing it. It's kinda like so:

```csharp
var tree = CSharpSyntaxTree.ParseText(@"
public class RandomTest {
  // stuff in here
}");
 
 
var compilation = CSharpCompilation.Create(
    "TestCompilation",
    syntaxTrees: new[] { tree },
    references: new[] { MetadataReference.CreateFromFile( typeof( object ).Assembly.Location ) }
);

// now use the compilation unit and do stuffs...
```

We could easily make a `TestBase` that abstracts this later (Microsoft does this in their default templates for diagnostics), but I think just doing it vanilla is fine for now too.

j3parker:
Ok, a lot more tests to write but I at least think this is not totally wrong.

saquibmian:
Merging this so that I can make use of this stub in master. Obviously not quite ready for shipping just yet.

PR #3, opened 2016-08-22
Adding an analyzers project and an analyzers test project

saquibmian:
Adding an analyzers project, and including the one that I wrote previously for types that are not safe to use in tests not marked Isolated.

@j3parker FYI


src/D2L.CodeStyle.Analyzers/D2L.CodeStyle.Analyzers.nuspec
```diff
@@ -0,0 +1,29 @@
...
+    <requireLicenseAcceptance>false</requireLicenseAcceptance>
+    <description>$description$</description>
+    <copyright>$copyright$</copyright>
+    
+    <frameworkAssemblies>
+      <frameworkAssembly assemblyName="System" targetFramework="" />
+    </frameworkAssemblies>
+
+    <dependencies>
+      <dependency id="D2L.CodeStyle.Analysis" version="$version$" />
```
j3parker: It depends on itself?

src/D2L.CodeStyle.Analyzers/ParallelizableTestsAnalyzer.cs
```diff
@@ -0,0 +1,130 @@
...
+            m_offendingTypeSuggestions = ofendingTypeSuggestionsOverride;
+        }
+
+        public override void Initialize( AnalysisContext context ) {
+            context.RegisterCompilationStartAction( RegisterIfTestAssembly );
+        }
+
+        private void RegisterIfTestAssembly( CompilationStartAnalysisContext compilation ) {
+            var references = compilation.Compilation.ReferencedAssemblyNames;
+            if( !references.Any( r => r.Name.ToUpper().Contains( "NUNIT" ) ) ) {
```
j3parker: Nice. Worth validating this is legit in the monolith. I have vague memories that we have some prod assemblies dep'ing on this (super wrong)

src/D2L.CodeStyle.Analyzers/D2L.CodeStyle.Analyzers.nuspec
saquibmian: Analyzers depends on Analysis. Subtle :)

src/D2L.CodeStyle.Analyzers.Test/D2L.CodeStyle.Analyzers.Test.csproj
```diff
@@ -0,0 +1,134 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
j3parker: This (and friends) into `test/`? (and root namespace that matches what is being tested. Also `UnitTest` vs. `IntegrationTest`+categories but I'm not sure we need to cargo-cult that hard (maybe give a look at MS repos)

saquibmian: Yes, forgot about that. I'll move these around after I get the build working.

src/D2L.CodeStyle.Analyzers/D2L.CodeStyle.Analyzers.nuspec
j3parker: Ahhh I see :) So one assembly per analyzer. gotcha.

travisMiehm: Does it need to be a different project? The naming suggests they are related.

And from the names, you need something to analyze (ie analyzers) before you can have an analysis.  So this sounds backwards

saquibmian: The idea is that Analysis has the core analysis stuff (utilities, visitors, etc.), and that Analyzers is actually just Roslyn analyzers that wrap that stuff. That makes it very clear that "these have Roslyn analyzers" and "this has stuff that all roslyn analyzers use".

If the distinction ends up being bad in practice, I can squash the projects into just Analyzers later on.

j3parker:
NuGet :(

travisMiehm:
Looks good, :shipit:

PR #4, opened 2016-08-24
Implementing analyzer that assess immutability of statics.

saquibmian:
The analyzer now assess immutability of statics and reports unsafety there as an error.

Mutability is considered at the field/property level only; method purity is not considered (at least yet). Mutability is also only considered as it relates to the class externally (fields or properties with private accessors or private setters are considered immutable).

A type is considered immutable iff (1) all of its fields/properties are immutable and (2) all of the types of said fields/properties are also immutable.
A field/property is considered immutable iff (1) its declaration is either private or immutable (readonly, private setter) and (2) its type is also immutable.

We only care about statics right now, so static-ness is enforced by the analyzer, and everything else is delegated to the `MutabilityInspector`.


saquibmian:
To experiment with these analyzers locally, add this to a csproj file you want analyzed:

```xml
  <ItemGroup>
    <Analyzer Include="C:\dev\D2L.CodeStyle\src\D2L.CodeStyle.Analyzers\bin\Debug\D2L.CodeStyle.Analyzers.dll" />
    <Analyzer Include="C:\dev\D2L.CodeStyle\src\D2L.CodeStyle.Analysis\bin\Debug\D2L.CodeStyle.Analysis.dll" />
  </ItemGroup>
```

(alter the paths as needed)

src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
```diff
@@ -0,0 +1,86 @@
...
+
+		/// <summary>
+		/// A list of known non-valuetype immutable types
+		/// </summary>
+		private static readonly ImmutableHashSet<string> KnownImmutableTypes = new HashSet<string> {
+			nameof(String)
+		}.ToImmutableHashSet();
+
+		/// <summary>
+		/// Determine if a given type is immutable.
```
j3parker: immutable -> mutable for clarity

src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
```diff
@@ -0,0 +1,86 @@
...
+				return false;
+			}
+			if( prop.SetMethod.DeclaredAccessibility == Accessibility.Private ) {
+				return false;
+			}
+			return true;
+		}
+
+		/// <summary>
+		/// Determine if a field is externally immutable (private fields are considered immutable).
```
j3parker: Why are private fields considered immutable? I can see some value in analyzing the public interface separately from the internals of the functions but that's (`private static`) our most pressing concern at the moment... ditto for properties

saquibmian: I think you're right. I wanted to separate the public and private interfaces because we will have problems with interface types (where we can only inspect the public interface), but I agree that opens a particular class of problems. 

Interfaces can give us false positives like this:

```csharp
public class ClientNameFetcher {
    private readonly IDbFactory m_dbFactory;

    public ClientNameFetcher( IPluginsManger plugins ) {
         // as an example of something that is expensive, so this class should be cached, either as a singleton or as a static -- regardless its lifetime should be longer than the thing using it, and there is possibility of data leakage in general; remember that singletons and statics behave the same as far as leakage is concerned
        m_dbFactory = plugins.GetPlugin<IDbFactory>().First(); 
    }
    public string GetNameOfClient( Guid clientGuid ) {
        return m_dbFactory.GetDb( clientGuid, "Main" ).GetClientName();
    }
}
public class ClientNameController {
    private static readonly ClientNameFetcher = new ClientNameFetcher(); // or injected in, where its registered as a singleton
    public string Get( Guid clientGuid ) {
        return m_clientNameFetcher.Get( clientGuid );
    }
}
```
In this example (which is a problem if `ClientNameFetcher.m_dbFactory` holds client-specific data internally), we can't verify that `ClientNameController.ClientNameFetcher` is immutable, because the type can't be verified to be immutable, because `ClientNameFetcher.m_dbFactory` is an interface, and we can only reason about its public interface, not its private implementation. I wanted to look only at the public interface for this reason, but you're right, we can be smarter about private data where we have it. 

I'll fix this.

src/D2L.CodeStyle.Analyzers/UnsafeStaticsAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+
+        private const string Title = "Ensure that static field is safe in undifferentiated servers.";
+        private const string Description = "Static fields should not have client-specific or mutable data, otherwise they will not be safe in undifferentiated servers.";
+        internal const string MessageFormat = "This static is unsafe because: '{0}'.";
+
+        private static readonly DiagnosticDescriptor Rule = new DiagnosticDescriptor(
+            DiagnosticId, Title, MessageFormat, Category,
+            DiagnosticSeverity.Error,
+            isEnabledByDefault: true,
+            description: Description
```
travisMiehm: Gonna nit: I hate mixing formatting.  Either put it all on different lines, or together

saquibmian: Fair.

tests/D2L.CodeStyle.Analysis.UnitTests/D2L.CodeStyle.Analysis.UnitTests.csproj
```diff
@@ -41,12 +41,40 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
-    <Reference Include="Microsoft.CodeAnalysis, Version=1.3.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35, processorArchitecture=MSIL">
-      <HintPath>..\..\packages\Microsoft.CodeAnalysis.Common.1.3.2\lib\net45\Microsoft.CodeAnalysis.dll</HintPath>
+    <Reference Include="Castle.Core, Version=3.3.0.0, Culture=neutral, PublicKeyToken=407dd0808d44fbdc, processorArchitecture=MSIL">
```
travisMiehm: Where's this used?

saquibmian:
FYI Building D2L.LE.Content with this analyzer gives me 41 errors. 4-5 come from codegen'd files, but 1/2 are `StaticIsMutable`.

tests/D2L.CodeStyle.Analysis.UnitTests/D2L.CodeStyle.Analysis.UnitTests.csproj
saquibmian: Moq uses it to create proxy classes.

travisMiehm: (Moq appears to have added it in 4.5.7.  Possibly when it stopped using ILMerge?)

src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
```diff
@@ -0,0 +1,77 @@
...
+				}
+				// skip methods, events, ctors, etc.
+			}
+
+			return false;
+		}
+
+		/// <summary>
+		/// Determine if a property is mutable.
+		/// This does not check if the type of the property is also mutable; use <see cref="IsTypeMutable"/> for that.
```
travisMiehm: Would it make sense to check the type here?

saquibmian: I kept these two separate so that I can present better error messages to the user. If this method returns false, it's because the field/property is not readonly. If it returns true, the analyzer still needs to check if the type is also immutable.

travisMiehm: I guess that works.  The user then goes and fixes the type, and the property is fixed too.  Cool

travisMiehm:
I like it.  You may :shipit:

src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
```diff
@@ -0,0 +1,97 @@
...
+			}
+
+			if( KnownImmutableTypes.Contains( type.Name ) ) {
+				return false;
+			}
+
+			foreach( var member in type.GetMembers() ) {
+				if( member is IPropertySymbol ) {
+					var prop = member as IPropertySymbol;
+					if( IsPropertyMutable( prop ) || IsTypeMutable( prop.Type ) ) {
```
saquibmian: Fail, I think this logic is wrong...
if a field is not mutable, this will continue, even if the type is mutable.

saquibmian: NEvermind, Travis confirms this is cool.

saquibmian:
I think this is it for this PR. I've got immutability analyzer stuff that I will throw in a different branch, so that Jacob can take a stab at it.

tests/D2L.CodeStyle.Analysis.UnitTests/MutabilityInspectorTests.cs
```diff
@@ -0,0 +1,232 @@
...
+
+		[Test]
+		public void IsTypeMarkedImmutable_Yes_ReturnsTrue() {
+			var type = Type( "[Immutable] class Foo {}" );
+
+			Assert.IsTrue( m_inspector.IsTypeMarkedImmutable( type ) );
+		}
+
+		[Test]
+		public void IsTypeMarkedImmutable_InterfaceIs_ReturnsTrue() {
```
travisMiehm: Test for multiple interfaces?

src/D2L.CodeStyle.Analyzers/UnsafeStaticsAnalyzer.cs
```diff
@@ -0,0 +1,143 @@
...
+                }
+#pragma warning restore CS0618 // Type or member is obsolete
+
+                if( symbol.GetAttributes().Any( a => a.AttributeClass.MetadataName == nameof( Statics.Audited ) ) ) {
+                    // anything marked audited has been explicitly marked as a safe static
+                    return;
+                }
+
+                if( m_immutabilityInspector.IsFieldMutable( symbol ) ) {
+                    var diagnostic = Diagnostic.Create( Rule, variable.GetLocation(), variable.Identifier.ValueText, "it" );
```
travisMiehm: hahah

Why not
`private const string IT = "it";`

travisMiehm:
Reviewed commits as they came in.  Still looks good.  :ship: it!

PR #9, opened 2016-09-06
Adding D2L-specific types that are known to be 'safe', as well as D2L…

saquibmian:
… semantics around safe code-gen (i.e., DTOs/IFields).

Re. the code gen stuff, it's not nice, but that only other way I could think of to get around this was to either add attributes as part of codegen, or update the codegen to use immutable collections and go through and fix the non-immutable element types. I didn't do that because (1) I'd like to get something in the build and (2) that would require updating every project that uses codegen to reference either `System.Collections.Immutable` or `D2L.CodeStyle.Annotations` -- that's a fairly significant change.

FYI @travisMiehm


src/D2L.CodeStyle.Analyzers/UnsafeStaticsAnalyzer.cs
```diff
@@ -49,6 +49,11 @@ public sealed class UnsafeStaticsAnalyzer : DiagnosticAnalyzer {
         }
 
         private void AnalyzeField( SyntaxNodeAnalysisContext context ) {
+            if( context.Node.SyntaxTree.FilePath.Contains( ".generated" ) ) {
```
ableeck: Is it accurate enough to check for ".generated" anywhere in the path or should it be a component of the path ending in ".generated"? I assume the former is true, but wanted to make sure the other case was considered...

saquibmian: Yes, `.generated` is a folder name, not an extension, so it can be anywhere in the path.

ableeck: As chatted about offline: The code does not just exclude things that are in a folder called `.generated`, but also things with a partial name of `.generated`, say `d2l.automation.generated.contract.tests.sln` or similar.

saquibmian: Should be fixed in the latest PR.

tests/D2L.CodeStyle.Analysis.UnitTests/UtilsTests.cs
```diff
@@ -0,0 +1,41 @@
...
+
+namespace D2L.CodeStyle.Analysis {
+
+    internal sealed class UtilsTests {
+
+        private readonly Utils m_utils = new Utils();
+
+        [Test]
+        public void IsGeneratedCodefile_NotCSharpFile_False() {
+            const string file = @"proj\random.js";
```
ableeck: Would be a slightly more useful test if the path contained '.generated'.

src/D2L.CodeStyle.Analysis/Utils.cs
```diff
@@ -0,0 +1,26 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace D2L.CodeStyle.Analysis {
+
+    public sealed class Utils {
+
+        private static readonly string GeneratedFolderPathSegment = Path.DirectorySeparatorChar + ".generated" + Path.DirectorySeparatorChar;
```
ableeck: This should work as long as you wouldn't get a path like `.generated\random.cs`...

tests/D2L.CodeStyle.Analysis.UnitTests/UtilsTests.cs
saquibmian: This test is about the extension only. The path doesn't matter here, if it's not a C# file, it can't be generated (even if the path is in ".generated").

src/D2L.CodeStyle.Analysis/Utils.cs
saquibmian: That's why I did `path = Path.GetFullPath( path )` :)

j3parker:
:shipit:

tests/D2L.CodeStyle.Analysis.UnitTests/UtilsTests.cs
ableeck: Agreed, but having the '.generated' in the path means that the order of the check in code doesn't matter if the code changes.

PR #11, opened 2016-09-19
Fix collection type marking as Immutable

saquibmian:
Fixing collection types not respecting the `[Immutable]` attribute, and adding a test case for that.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/UnsafeStaticsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/UnsafeStaticsAnalyzer.cs
@@ -98,6 +98,21 @@ namespace D2L.CodeStyle.Analyzers {
                     var diagnostic = Diagnostic.Create( Rule, variable.GetLocation(), variable.Identifier.ValueText, symbol.Type.GetFullTypeName() );
                     context.ReportDiagnostic( diagnostic );
                 }
+
+                if( symbol.Type.IsImmutableCollectionType() ) {
+                    var elementType = symbol.Type.GetCollectionElementType();
+
+                    // non-generic collections
+                    if( elementType == null ) {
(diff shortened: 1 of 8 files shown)


tests/D2L.CodeStyle.Analyzers.Test/UnsafeStaticsAnalyzerTests.cs
```diff
@@ -172,6 +172,96 @@ internal class Foo {
...
+            const string test = @"
+    using System;
+
+    namespace test {
+        class Tests {
+            public static readonly System.Collections.IList bad;
+
+        }
+    }";
+            AssertSingleDiagnostic( test, 6, 61, "bad", "System.Collections.IList" );
```
saquibmian: line number and column number of where the diagnostic should be reported.

src/D2L.CodeStyle.Analyzers/UnsafeStaticsAnalyzer.cs
```diff
@@ -98,6 +98,21 @@ public sealed class UnsafeStaticsAnalyzer : DiagnosticAnalyzer {
                     var diagnostic = Diagnostic.Create( Rule, variable.GetLocation(), variable.Identifier.ValueText, symbol.Type.GetFullTypeName() );
                     context.ReportDiagnostic( diagnostic );
                 }
+
+                if( symbol.Type.IsImmutableCollectionType() ) {
```
saquibmian: I don't, because a non-immutable collection is flagged no matter what the element type is. 

Essentially the error message should look like this:
`List<foo>` --> List is mutable (don't bother even checking the element type)
`ImmutableList<foo>` --> foo is mutable
`ImmutableList<DateTime>` -- safe

saquibmian:
Note there will be merge conflicts after merge #13, because I put fixes in there to handle boundless recursion. You are free to review, and trust that I will resolve merge conflicts correctly...or you can wait, if you want.

PR #13, opened 2016-09-20
Updating the immutability checking to not be ultra conservative about

saquibmian:
methods in immutable types, and to ignore interfaces, abstract classes, or
non-sealed classes.

### DIFF
--- a/src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
@@ -89,6 +89,15 @@ namespace D2L.CodeStyle.Analysis {
 				return false;
 			}
 
+
+			// make sure we're dealing with the concrete implementation
+			if( type.TypeKind == TypeKind.Interface ) {
+				return true;
+			}
+			if( type.TypeKind == TypeKind.Class && ( !type.IsSealed || type.IsAbstract ) ) {
+				return true;
+			}
+
 			if( ImmutableCollectionTypes.Contains( type.GetFullTypeName() ) ) {
 				var namedType = type as INamedTypeSymbol;
 				if( namedType == null ) {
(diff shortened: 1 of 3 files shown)


src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
```diff
@@ -89,6 +89,15 @@ ITypeSymbol type
 				return false;
 			}
 
+
+			// make sure we're dealing with the concrete implementation
+			if( type.TypeKind == TypeKind.Interface ) {
+				return true;
+			}
+			if( type.TypeKind == TypeKind.Class && ( !type.IsSealed || type.IsAbstract ) ) {
```
JeffAshton: I think IsAbstract is irrelevant here.  Removing it, all the tests still pass.  I think we really just care that it isn't sealed

src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
```diff
@@ -128,14 +137,8 @@ ITypeSymbol type
 					continue;
 				}
 				if( member is IMethodSymbol ) {
```
JeffAshton: The contents of this foreach loop field like a method in itself.  That way we can have less control statements

src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
saquibmian: True. Removed.

src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
saquibmian: I wasn't able to remove this, because of the cycle detection changes (I'd have to pass the stack along to the helper method, which just seemed wrong). Can you think of a different way that I'm not seeing?

PR #14, opened 2016-09-21
Being smarter by checking concrete type's of fields/props where we can.

saquibmian:


saquibmian:
Note: I haven't accounted for conditionally checking for sealed based on left- or right-hand value.

saquibmian:
@JeffAshton I'm accounting for unsealed conditionally now. Not super sure about the design of my implementation...take a look.

PR #15, opened 2016-09-22
Ignore static members when inspecting for immutability.

saquibmian:


src/D2L.CodeStyle.Analysis/MutabilityInspector.cs
```diff
@@ -125,7 +125,7 @@ HashSet<ITypeSymbol> typeStack
 
 				{
 					foreach( ISymbol member in type.GetMembers() ) {
-						if( IsMemberMutableRecursive( member, typeStack ) ) {
+						if( !member.IsStatic && IsMemberMutableRecursive( member, typeStack ) ) {
```
travisMiehm: I do not follow how you can say a type with only static members is immutable. Can you explain?

saquibmian: A type with only static members has no state. My terminology probably conflates `type` with `instance of type`, but we really only care about the latter. 

Example:


```csharp
public class Foo {
    public static readonly Foo Default = new Foo();
}
```
`Foo` is safe here.

Why do we care? So that we can be smarter about something like this:

```csharp
public class PlainTextLangterm {

    public static readonly PlainTextLangterm Default = new PlainTextLangterm("Default");

    private readonly string _value;
    public PlainTextLangterm( string value ) { _value = value; }
}
```

travisMiehm: In the case of `Foo`, it is not just the `static` that makes it safe.  The `readonly` is also needed.

Consider your second case where `Default` is not `readonly`.  No longer safe.

What about 

```
public class Foo {
    public static readonly Bar Default = new Bar();
}
```

where `Bar` is mutable?

saquibmian: I definitely agree that the static itself also needs to be `readonly` -- the analyzer already picks that up, but not for types that we don't own, granted. I could probably push the "ignore statics" down to just before checking the type of the static itself.

In your second example, where `Bar` is mutable, that makes the static field `Default` illegal via the analyzer, but that shouldn't make `Foo` itself bad. Example:

```csharp
public class Foo {
    public static readonly Bar Default = new Bar(); // Bar is mutable, so this is not safe, analyzer will report error
}
public readonly static Foo GoodGuy = new Foo(); // this is safe, even if Foo.Default isn't
```
In this particular case, the problem is in the `GoodGuy` field, the problem is in the `Foo.Default` field. So we should report an error _there_ instead.

travisMiehm: Cool.  Misunderstood how we'd report this.

PR #16, opened 2016-09-23
Ignore properties that are really just functions.

saquibmian:
This ignores things like:
```csharp
public static IFoo FooBar { 
    get { 
        return new Foo(); 
    } 
}
```
...because they're not actually static variables.

### DIFF
--- a/src/D2L.CodeStyle.Analysis/TypeSymbolExtensions.cs
+++ b/src/D2L.CodeStyle.Analysis/TypeSymbolExtensions.cs
@@ -6,6 +6,7 @@ using System.Collections.Immutable;
 using System.Linq;
 
 namespace D2L.CodeStyle.Analysis {
+<<<<<<< HEAD
     public static class TypeSymbolExtensions {
         private static readonly SymbolDisplayFormat FullTypeDisplayFormat = new SymbolDisplayFormat(
             typeQualificationStyle: SymbolDisplayTypeQualificationStyle.NameAndContainingTypesAndNamespaces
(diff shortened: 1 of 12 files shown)


src/D2L.CodeStyle.Analysis/TypeSymbolExtensions.cs
```diff
@@ -41,4 +43,11 @@ public static class TypeSymbolExtensions {
             return args;
         }
     }
+
+    public static class SyntaxNodeExtension {
+        public static bool IsPropertyGetterImplemented(this PropertyDeclarationSyntax syntax) {
```
travisMiehm: Tests for this?

saquibmian:
Ugh stupid TortoiseGit. Hang on.

tests/D2L.CodeStyle.Analysis.UnitTests/D2L.CodeStyle.Analysis.UnitTests.csproj
```diff
@@ -119,6 +121,9 @@
     <Analyzer Include="..\..\packages\Microsoft.CodeAnalysis.Analyzers.1.1.0\analyzers\dotnet\cs\Microsoft.CodeAnalysis.Analyzers.dll" />
     <Analyzer Include="..\..\packages\Microsoft.CodeAnalysis.Analyzers.1.1.0\analyzers\dotnet\cs\Microsoft.CodeAnalysis.CSharp.Analyzers.dll" />
   </ItemGroup>
+  <ItemGroup>
+    <Service Include="{82A7F48D-3B50-4B1E-B82E-3ADA8210C358}" />
```
saquibmian: kill

src/D2L.CodeStyle.Analysis/TypeSymbolExtensions.cs
saquibmian: Added.

saquibmian:
@travisMiehm good?

tests/D2L.CodeStyle.Analysis.UnitTests/TestBase.cs
```diff
@@ -0,0 +1,59 @@
...
+using NUnit.Framework;
+using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using System.Text;
+
+namespace D2L.CodeStyle.Analysis {
+
+    public class TestBase {
```
travisMiehm: I don't like this as a base class at all.  Go with static methods.

saquibmian: Why? I prefer this base class over static methods...

PR #17, opened 2016-10-20
System.Collections.Generic.IReadOnlyDictionary is safe.

saquibmian:
IReadOnlyDictionary, althought itself potentially backed by a mutable
dictionary, is to be considered safe. Using ImmutableDictionary itself
is not enough, because it implements the (mutating) IDictionary interface
and simply throws on mutating methods. Grrr.


j3parker:
:shipit:

PR #19, opened 2016-10-27
Ignore resource generated files, and TypeConverter

saquibmian:


tests/D2L.CodeStyle.Analysis.UnitTests/TypeSymbolExtensionsTests.cs
```diff
@@ -1,33 +1,35 @@
-﻿using Microsoft.CodeAnalysis.CSharp.Syntax;
```
saquibmian: This file was renamed because the name was wrong...it shows up pretty poorly.

saquibmian:
It helps to add `?w=1` to the URL to kill the whitespace changes.

PR #20, opened 2016-11-01
Remove D2L.CodeStyle.Analysis, and package everything in a single ass…

saquibmian:
Remove D2L.CodeStyle.Analysis, and package everything in a single assembly.

On Monday night there were a few build failures caused by the following error:

```
"C:\d2l\instances\all\checkout\lp\framework\core\D2L.LP\D2L.LP.csproj" (default target) (1) ->
(CoreCompile target) ->
  CSC : error CS8032: An instance of analyzer D2L.CodeStyle.Analyzers.UnsafeStaticsAnalyzer cannot be created from C:\d
2l\instances\all\checkout\packages\D2L.CodeStyle.Analyzers.0.4.4\lib\net461\D2L.CodeStyle.Analyzers.dll : Exception has been thrown by the target of an invocation.. [C:\d2l\instances\all\checkout\lp\framework\core\D2L.LP\D2L.LP.csproj]
```

It turns out this is because there is some non-determinism in loading analyzers during the build process. If an analyzer assembly depends on another assembly, that second assembly must be in the GAC, or must be specified as a dummy analyzer. If the latter option is used, there is non-determinism in which analyzer reference gets loaded into the LoadFrom context first. In our case, if Analysis is loaded before Analyzer, all is well. But if Analyzer is loaded before Analysis, it fails. 

To fix this, we will merge all those two assemblies together. This also means that, to be safe, Analysis will no longer be able to depend on Annotations directly.


j3parker:
That was surprising

PR #23, opened 2016-11-30
Exclude projects with the name Test in them.

saquibmian:
In the event that references cannot be resolved, such as when a build has
not yet occurred and approot/bin does not exist, we can use another
heuristic to ignore test projects: the name of the project.
We need this heuristic only because a component-build should be a
requirement for this tool; only source code + references. Unfortunately,
there is no way to deploy references to the correct location without doing
a component-build.


j3parker:
:shipit:

PR #25, opened 2016-12-02
Removing System.IEqualityComparer, becuase it's an interface that is probably easily statefully implemented.

saquibmian:
`System.StringComparer` remains for now while I verify that we haven't
subclassed it anywhere; if we have, then I will remove that one too and
give the analyzer the ability to allow certain fields/properties that are
safe (e.g., `StringComparer.Ordinal`).


j3parker:
Lgtm

PR #26, opened 2016-12-08
Making all dependencies in the UnsafeStaticCounter a dev dependency.

saquibmian:
AppVeyor calls `nuget pack` in the csproj's context, which asigns all dependencies in the `packages.config` file as NuGet dependencies in the final NuGet package. This is not necessary, because the tool bundles all dependencies with it. Instead of overwriting AppVeyor's NuGet packing behaviour (and having to pass in all variables to `nuget pack` ourselves), we just mark those dependencies as development dependencies, to prevent them from being required for the final package.

### DIFF
--- a/src/D2L.CodeStyle.UnsafeStaticCounter/Properties/AssemblyInfo.cs
+++ b/src/D2L.CodeStyle.UnsafeStaticCounter/Properties/AssemblyInfo.cs
@@ -10,5 +10,5 @@ using System.Runtime.InteropServices;
 
 [assembly: ComVisible( false )]
 
-[assembly: AssemblyVersion( "0.5.4.0" )]
-[assembly: AssemblyFileVersion( "0.5.4.0" )]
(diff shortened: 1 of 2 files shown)


saquibmian:
I've checked that the package created has no dependencies.

src/D2L.CodeStyle.UnsafeStaticCounter/Properties/AssemblyInfo.cs
```diff
@@ -10,5 +10,5 @@
 
 [assembly: ComVisible( false )]
```
JeffAshton: Rather this was generated by app veyor actually.  ask me in person

saquibmian: Yeah I saw that recently (assuming you're talking about the`assembly_info` rule in the yaml). I can switch it after you do Core.

PR #29, opened 2017-01-11
Adding an Immutability analyzer

saquibmian:
Adding an Immutability analyzer to ensure that objects implementing immutable objects/interfaces are also immutable.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
@@ -17,7 +17,7 @@ namespace D2L.CodeStyle.Analyzers.Common {
 	public sealed class MutabilityInspector {
 
 		/// <summary>
-		/// A list of known non-valuetype immutable types
+		/// A list of known immutable types
 		/// </summary>
 		private static readonly ImmutableHashSet<string> KnownImmutableTypes = new HashSet<string> {
 			"Newtonsoft.Json.JsonSerializer",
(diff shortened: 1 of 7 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+
+		private readonly MutabilityInspector m_immutabilityInspector = new MutabilityInspector();
+
+		public override void Initialize( AnalysisContext context ) {
+			context.RegisterCompilationStartAction( RegisterIfNotTestAssembly );
+		}
+
+		private void RegisterIfNotTestAssembly( CompilationStartAnalysisContext compilation ) {
+			var references = compilation.Compilation.ReferencedAssemblyNames;
+			if( references.Any( r => r.Name.ToUpper().Contains( "NUNIT" ) ) ) {
```
j3parker: I think we're going to want to move away from this sorta thing in awhile but let's talk about that when we're both in the office

saquibmian:
There are a lot of tests that still need to be added, e.g., nested immutables, so not quite ready to merge yet.

src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
```diff
@@ -34,10 +35,20 @@ public sealed class MutabilityInspector {
...
 			"System.Uri",
 			"System.String",
-			"System.StringComparer",
 			"System.Xml.Serialization.XmlSerializer"
 		}.ToImmutableHashSet();
 
+		/// <summary>
+		/// A list of marked immutable types owned externally.
+		/// </summary>
+		private static readonly ImmutableHashSet<string> MarkedImmutableTypes = new HashSet<string> {
```
saquibmian: We call these out separately because it allows us to do things like this:

```
class Foo : StringComparer { ... } // Analyzer will ensure that `Foo` is immutable, because it's marked immutable
```
Then we can use `Foo` anywhere and be happy.

PR #30, opened 2017-01-11
Add some more immutable .NET types

j3parker:
This drops 51

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
@@ -24,7 +24,6 @@ namespace D2L.CodeStyle.Analyzers.Common {
 			"System.DateTime",
 			"System.Guid",
 			"System.Lazy",
-			"System.Net.IPAddress",
 			"System.Reflection.ConstructorInfo",
 			"System.Reflection.FieldInfo",
 			"System.Reflection.MemberInfo",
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
```diff
@@ -24,6 +24,7 @@ public sealed class MutabilityInspector {
 			"System.DateTime",
 			"System.Guid",
 			"System.Lazy",
+			"System.Net.IPAddress",
```
saquibmian: Definitely mutable: https://referencesource.microsoft.com/#System/net/System/Net/IPAddress.cs,341

j3parker: Ah I missed the obsolete method on MSDN :(

saquibmian: There's also this guy, but don't know what `Scope` is: https://referencesource.microsoft.com/#System/net/System/Net/IPAddress.cs,412

j3parker: Yep super broken, my bad

j3parker:
This fellas still good for 46

saquibmian:
`System.Workflow.ComponentModel.DependencyProperty` is technically mutable (non-readonly), but not through the public interface, so I'm cool with it.

:shipit: (after you've bumped the version number)

saquibmian:
fyi: if you want the counter to pick it up, you'll have to release a new version of it too.

j3parker:
yep

PR #31, opened 2017-01-12
Use a declarations initializers type via the semantic model for analysis

j3parker:
Will add some more tests.

I believe that `GetTypeInfo` doesn't return `null`: (after following some steps) http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/Compilation/CSharpSemanticModel.cs,e0a02c903a059c8f

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/UnsafeStatics/UnsafeStaticsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/UnsafeStatics/UnsafeStaticsAnalyzer.cs
@@ -169,11 +169,23 @@ namespace D2L.CodeStyle.Analyzers.UnsafeStatics {
 				return;
 			}
 
-			// try to use the concrete type if we have it (via a constructor)
 			var flags = MutabilityInspectionFlags.Default;
-			var constructedType = GetConstructedType( context, exp );
-			if( constructedType != null ) {
-				type = constructedType;
+
+			// Always prefer the type from the initializer if it exists because
+			// it may be more specific.
+			if( exp != null ) {
+				var typeInfo = context.SemanticModel.GetTypeInfo( exp );
+
+				if( typeInfo.Type is IErrorTypeSymbol ) {
+					return;
+				}
+
+				type = typeInfo.Type;
+			}
+
+			// When we know the concrete type as in "new T()" we don't have to
+			// be paranoid about mutable derived classes
+			if ( exp is ObjectCreationExpressionSyntax ) {
 				flags |= MutabilityInspectionFlags.AllowUnsealed;
 			}
 
@@ -182,28 +194,6 @@ namespace D2L.CodeStyle.Analyzers.UnsafeStatics {
 				context.ReportDiagnostic( diagnostic );
 			}
 		}
-
-		private ITypeSymbol GetConstructedType(
-			SyntaxNodeAnalysisContext context,
-			ExpressionSyntax exp
-		) {
-
-			if( exp == null ) {
-				return null;
-			}
-
-			var initializerSymbol = context.SemanticModel.GetSymbolInfo( exp ).Symbol as IMethodSymbol;
-			if( initializerSymbol == null ) {
-				return null;
-			}
-
-			if( initializerSymbol.MethodKind != MethodKind.Constructor ) {
-				return null;
-			}
-
-			return initializerSymbol.ContainingType;
-		}
-
 
 		private Diagnostic CreateDiagnostic( Location location, string fieldOrPropName, string offendingType ) {
 			var builder = ImmutableDictionary.CreateBuilder<string, string>();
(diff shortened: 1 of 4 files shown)


j3parker:
Got a bug to fix!

j3parker:
Ok I'm happy with these changes - the previous version was flipping the flag about sealed when it shouldn't have.

The diff is harder to read than it needs to be but oh well. Added a test specifically for that `StringComparer` case so we know it's good to go.

src/D2L.CodeStyle.Analyzers/UnsafeStatics/UnsafeStaticsAnalyzer.cs
```diff
@@ -169,42 +169,32 @@ string fieldOrPropName
...
+			// it may be more specific.
+			if( exp != null ) {
+				var typeInfo = context.SemanticModel.GetTypeInfo( exp );
 
-		private ITypeSymbol GetConstructedType(
-			SyntaxNodeAnalysisContext context,
-			ExpressionSyntax exp
-		) {
+				if( typeInfo.Type is IErrorTypeSymbol ) {
+					return;
```
j3parker: This can happen when you have a syntax error.

I'll do a follow-up PR that makes our tests always assert there aren't any types of non-our-analyzer errors because right now it's possible to create test cases that pass when they shouldn't if you make a typo in the test string (this confused me for awhile...)

saquibmian: Yeah I briefly went down that road too, but then I ran into legit compilation issues like "no main method found" and it wasn't hugely worth the effort to fix that. If you find a nice way, that's awesome.

tests/D2L.CodeStyle.Analyzers.Test/UnsafeStatics/UnsafeStaticsAnalyzerTests.cs
```diff
@@ -672,6 +674,61 @@ internal class Bar {
...
+				void IUnsafe.Magic() {} // looks safe to me
+				public static readonly Safe Instance { get; } = new Safe();
+			}
+
+			private readonly static IUnsafe foo = Safe.Instance;
+		}
+	}";
+
+			// Safe isn't sealed
+			AssertSingleDiagnostic( test, 12, 36, "foo", "test.Tests.Safe" );
```
saquibmian: This is actually safe, because we know that `Safe.Instance` is concretely `Safe` (doesn't need to be sealed). But that's one level down (inspect a field's declaration) that can come in another PR. It's good that there's a test for this though.

tests/D2L.CodeStyle.Analyzers.Test/UnsafeStatics/UnsafeStaticsAnalyzerTests.cs
```diff
@@ -672,6 +674,61 @@ internal class Bar {
...
+		}
+
+		[Test]
+		public void DocumentWithStaticField_ReadonlyUnsafeBaseClassWithSafeInitializer_NoDiag() {
```
saquibmian: I think this test and the one before it should be named something like `WithSealedSafe` and `WithUnsealedSafe`; they both actually call the same initializer.

saquibmian:
LGTM

tests/D2L.CodeStyle.Analyzers.Test/UnsafeStatics/UnsafeStaticsAnalyzerTests.cs
j3parker: That'd be a bit tricky in general because we'd need to consider control flow (a field declaration could have loops, if statements, multiple returns...)

Not saying it wouldn't be cool to do but it'd definitely be more advanced. A starting point could be to find all return points and consider the union of all their types... but that would even lose things so you'd have to go deeper.

tests/D2L.CodeStyle.Analyzers.Test/UnsafeStatics/UnsafeStaticsAnalyzerTests.cs
j3parker: gave em a more descriptive name, thanks

tests/D2L.CodeStyle.Analyzers.Test/UnsafeStatics/UnsafeStaticsAnalyzerTests.cs
saquibmian: Well all static fields need to be audited anyway, so they're either (1) deterministically safe, (2) annotated with `audited`, or (3) clear/unaudited. In the former two cases, we should be able to come to the same conclusion one level upstream (gets tricky if we don't have the source code, which we often won't). I've got some thoughts on how to do that in my head, will take a crack after the immutable analyzer.

src/D2L.CodeStyle.Analyzers/UnsafeStatics/UnsafeStaticsAnalyzer.cs
j3parker: Ahhh... :(

PR #36, opened 2017-01-13
Fix assembly resolution for the counter.

saquibmian:


src/D2L.CodeStyle.UnsafeStaticCounter/Program.cs
```diff
@@ -49,33 +51,46 @@ internal sealed class Program {
...
 					default:
 						throw new InvalidOperationException( $"unknown option: {enumerator.Current}" );
 				}
 			}
 
 			if( string.IsNullOrWhiteSpace( path ) ) {
-				throw new InvalidOperationException( "usage: UnsafeStaticsCounter.exe -d {rootDir} [-n {concurrency} -o {outputFile}]" );
+				throw new InvalidOperationException( "usage: UnsafeStaticsCounter.exe -d {rootDir} -b {binDir} [-n {concurrency} -o {outputFile}]" );
+			}
+			path = Path.GetFullPath( path );
```
saquibmian: this fixes a related bug where relative checkout paths don't form properly, so we get no results (no csprojs).

src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
```diff
@@ -83,6 +83,11 @@ public sealed class MutabilityInspector {
 			MutabilityInspectionFlags flags,
 			HashSet<ITypeSymbol> typeStack
 		) {
+			if( type is IErrorTypeSymbol  || type == null ) {
+				// type can't be resolved; the project is either not configured correctly, or there's a typo
+				return false;
```
j3parker: Should we be marking types we don't know about as immutable? That seems hazardous

saquibmian: I'm conflicted. On one hand, it's conservative, so we should make it unsafe. On the other hand, a seemingly safe type `ImmutableArray<LangTerm>` could be marked unsafe, because some type within `LangTerm` can't be resolved because the project isn't configured correctly. 

In the end I chose to not confuse the dev. There is a third option, which is `throw`, which will show the dev an `analyzer threw exception` message; the issue with that is the analyzer could run mid-word typing (e.g., briefly pause after writing `Immutable`), which would hit this case.

saquibmian: At the very least, this behaviour is consistent with what happens currently (type.GetMembers() returns empty list, type gets reported as safe).

j3parker: I'm worried about us marking unsafe code as safe and never knowing about it/creating headaches later when we realize the mistake but need to keep CI working. I think defaulting to unsafe makes the most sense.

saquibmian: I think I'd rather throw than mark is as unsafe, then. Unsafe can be very confusing if the project is misconfigured, which should be the vast majority of cases.

saquibmian:
Fixed some whitespace, use `?w=1` to ignore whitespace.

saquibmian:
@j3parker any more issues with this guy?

saquibmian:
This will require a major version update to the counter, because it requires a new `-b` (bindir) option.

j3parker:
:ship: :it:

PR #40, opened 2017-01-25
Disallowing ValueTypes, and checking for primitive types instead.

saquibmian:
ValueTypes are not immutable if they hold references to non-valuetypes.
We instead allow only primitive types instead. We expect this to
break the analyzer in some places.


j3parker:
:shipit:

PR #45, opened 2017-01-31
Bless log4net.ILog.

saquibmian:
Will merge after #44 to bump version.


j3parker:
For interfaces in this list, we'll still check their implementations when we find them right?

saquibmian:
I can't seem to reply directly to a comment...fail. 
No, we don't check for implementations of stuff in `KnownImmutableTypes`. We only check for implementations anything that is marked `[Immutable]` (either self or an ancestor).

j3parker:
:shipit: we should work on that

saquibmian:
Issue for that here: https://github.com/Brightspace/D2L.CodeStyle/issues/47

PR #52, opened 2017-02-08
Add Analyzers for Static Check of TestCaseSource and ValueSource

LiyueWang:


src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+						}
+					}
+				}
+			}
+
+			// For each Source Report Diagnostic
+			foreach( var method in methods ) {
+				if( memberNames.Contains( method.Identifier.ToString() ) ) {
+					if( !method.Modifiers.Any( SyntaxKind.StaticKeyword ) ) {
+						var diagnostic = Diagnostic.Create( Rule, method.GetLocation(), "method (and the called methods by it)" );
```
saquibmian: Is it a better idea to report the diagnostics at the usages of the attribute instead?
I.e., something like this:

```csharp
private readonly int[] items = { 1, 2, 3, 4 };

[Test, TestCaseSource( nameof(item) )] // this line errors out, saying "'items' need to be static."
public void Test( int item ) { }
```

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+						if( attribute.Name.ToString() == "TestCaseSource" ) {
+							var attributeArguments = attribute.ArgumentList.Arguments.ToImmutableArray();
+							foreach( var attributeArgument in attributeArguments ) {
+								var expression = attributeArgument.Expression;
+								//StringLiteralExpression: "a", invocationExpression: nameof(a)  
+								if( expression is LiteralExpressionSyntax ) {
+									var member = expression.ToString().Replace( "\"", "" ).Replace( "\'", "" ).Trim();
+									memberNames.Add( member );
+								} else if( expression is InvocationExpressionSyntax ) {
+									var member = expression.ToString().Trim().Substring( 7, expression.ToString().Trim().Length - 8 ).Trim();
```
saquibmian: Is there a better way to pull out the argument of this `nameof` or `typeof` expression?

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+				foreach( var attributeList in attributeLists ) {
+					var attributes = attributeList.Attributes.ToImmutableArray();
+					foreach( var attribute in attributes ) {
+						if( attribute.Name.ToString() == "TestCaseSource" ) {
+							var attributeArguments = attribute.ArgumentList.Arguments.ToImmutableArray();
+							foreach( var attributeArgument in attributeArguments ) {
+								var expression = attributeArgument.Expression;
+								//StringLiteralExpression: "a", invocationExpression: nameof(a)  
+								if( expression is LiteralExpressionSyntax ) {
+									var member = expression.ToString().Replace( "\"", "" ).Replace( "\'", "" ).Trim();
```
saquibmian: You should be able to use `string.Trim(char[])` here.

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+			Category,
+			DiagnosticSeverity.Warning,
+			isEnabledByDefault: true,
+			description: Description
+		);
+
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create( Rule );
+
+		public override void Initialize( AnalysisContext context ) {
+			context.RegisterSemanticModelAction(
```
saquibmian: When does this action typically run? 
I'm curious if registering an attribute declaration syntax action might be better? Find all usages of the attribute, pull out the thing being referenced by the source (can do something like `context.GetSymbolInfo( syntaxNode ).Symbol`), and verify that it's static. 
Thoughts?

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+								var expression = attributeArgument.Expression;
+								//StringLiteralExpression: "a", invocationExpression: nameof(a)  
+								if( expression is LiteralExpressionSyntax ) {
+									var member = expression.ToString().Replace( "\"", "" ).Replace( "\'", "" ).Trim();
+									memberNames.Add( member );
+								} else if( expression is InvocationExpressionSyntax ) {
+									var member = expression.ToString().Trim().Substring( 7, expression.ToString().Trim().Length - 8 ).Trim();
+									memberNames.Add( member );
+								}
+							}
```
saquibmian: I can't figure out how you handle this case?
```
[TestCaseSource( typeof( T ), "FieldOnT" )]
```
You read each argument at a time ... :/

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+			context.RegisterSemanticModelAction(
+				AnalyzeSemanticModel
+			);
+		}
+
+		private void AnalyzeSemanticModel( SemanticModelAnalysisContext semanticModelAnalysisContext ) {
+			var tree = semanticModelAnalysisContext.SemanticModel.SyntaxTree;
+			var fields = tree.GetRoot().DescendantNodes().OfType<FieldDeclarationSyntax>().ToImmutableArray();
+			var properties = tree.GetRoot().DescendantNodes().OfType<PropertyDeclarationSyntax>().ToImmutableArray();
+			var methods = tree.GetRoot().DescendantNodes().OfType<MethodDeclarationSyntax>().ToImmutableArray();
```
saquibmian: Are these loading _all_ methods in the entire compilation? The perf for this might not be great...

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
LiyueWang: make sense to me, will do

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
LiyueWang: for `typeof( T )` the expression is `TypeofExpressionSyntax` and will be ignore
only `LiteralExpressionSyntax` for string and `InvocationExpressionSyntax` for nameof() will be considered

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
LiyueWang: I see, will do, thanks

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
saquibmian: We need to consider the type though, no? Also, I just noticed though that `memberNames` is not a list of fully-qualified names. I.e., I can have collisions in this case:

```csharp
class Foo {
    private static readonly int[] items = { 1,2,3,4 };
}
class FooTests {
    private readonly int[] items = { 1,2,3,4 };
    
    [Test, TestCaseSource( typeof( FooTest ), nameof(items) ]
    public void Test() {}
}
```
... where `items` exists twice; this is valid code, but the analyzer might not be deterministic here.

LiyueWang: Yes, you are right, I didn't consider this case, will update, thanks!

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
LiyueWang: > Register an action to be executed at completion of semantic analysis of a document

haven't done in this way before, will try

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
LiyueWang: Yes, pull out from the expression instead of the string, i.e.
`invocationExpression.ArgumentList.Arguments.First().Expression.ToString()`

this way is better, will do

src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
```diff
@@ -0,0 +1,90 @@
...
+						}
+
+						if( nameExpression is LiteralExpressionSyntax ) {
+							memberName = nameExpression.ToString().Trim( new char[] { '\"', '\'', ' ' } );
+						} else if( nameExpression is InvocationExpressionSyntax ) {
+							InvocationExpressionSyntax invocationExpression = nameExpression as InvocationExpressionSyntax;
+							memberName = invocationExpression.ArgumentList.Arguments.First().Expression.ToString();
+						}
+
+						var ns = context.SemanticModel.LookupNamespacesAndTypes( 0 )[0];
```
saquibmian: This might have bad perf. Try something like this:

```
diff --git a/src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs b/src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
index 3f7a887..eebe6f4 100644
--- a/src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/SourceAttribute/TestCaseSourceAttributeAnalyzer.cs
@@ -54,17 +54,17 @@ namespace D2L.CodeStyle.TestAnalyzers.SourceAttribute {
 				foreach( var attribute in attributes ) {
 					if( attribute.Name.ToString() == "TestCaseSource" ) {
 						var attributeArguments = attribute.ArgumentList.Arguments.ToImmutableArray();
-						String memberType = null;
+						ITypeSymbol typeContainingMember = null;
 						String memberName = null;
 						ExpressionSyntax nameExpression;
 
 						if( attributeArguments.Length == 2 ) {
 							var typeExpression = attributeArguments[0].Expression as TypeOfExpressionSyntax;
-							memberType = typeExpression.Type.ToString();
+							typeContainingMember = context.SemanticModel.GetTypeInfo( typeExpression.Type ).Type;
 							nameExpression = attributeArguments[1].Expression;
 						} else {
 							var methodSymbol = context.SemanticModel.GetDeclaredSymbol( method );
-							memberType = methodSymbol.ContainingType.MetadataName;
+							typeContainingMember = methodSymbol.ContainingType;
 							nameExpression = attributeArguments[0].Expression;
 						}
 
@@ -75,9 +75,7 @@ namespace D2L.CodeStyle.TestAnalyzers.SourceAttribute {
 							memberName = invocationExpression.ArgumentList.Arguments.First().Expression.ToString();
 						}
 
-						var ns = context.SemanticModel.LookupNamespacesAndTypes( 0 )[0];
-						var namedTypeSymbol = context.SemanticModel.Compilation.GetTypeByMetadataName( string.Format( "{0}.{1}", ns, memberType ) );
-						var source = context.SemanticModel.LookupSymbols( attribute.GetLocation().SourceSpan.Start, namedTypeSymbol, memberName ).Single();
+						var source = typeContainingMember.GetMembers( memberName ).First();
 						if( !source.IsStatic ) {
 							var diagnostic = Diagnostic.Create( Rule, attribute.GetLocation(), memberName );

```

saquibmian: (same idea with the valuesource analyzer)

PR #53, opened 2017-02-09
Unaudited reasons

j3parker:
Should look like this:

* `[Statics.Unaudited( Because.ItHasntBeenLookedAt )]` : what we'll do for the automated flagging
* `[Statics.Unaudited( Because.ItsSketchy )]` : this code is fucked in the head and if its safe that's its only saving grace; should improve it
* `[Statics.Unaudited( Because.ItsStickyDataOhNooo )]` : yup
* `[Statics.Unaudited( Because.WeNeedToMakeTheAnalyzerConsiderThisSafe )]` : we have a bunch of these, e.g. `ILocation` needs to be fixed, so we can use this in areas that we expect will be safe when the underlying type is fixed for example

### DIFF
--- a/src/D2L.CodeStyle.Annotations/Statics/Unaudited.cs
+++ b/src/D2L.CodeStyle.Annotations/Statics/Unaudited.cs
@@ -2,10 +2,27 @@
 
 // ReSharper disable once CheckNamespace
 namespace D2L.CodeStyle.Annotations {
+	public enum Because {
+		ItHasntBeenLookedAt,
+		ItsSketchy,
+		ItsStickyDataOhNooo,
+		WeNeedToMakeTheAnalyzerConsiderThisSafe
+	}
+
 	public static partial class Statics {
 		[Obsolete( "Static variables marked as unaudited require auditing. Only use this attribute as a temporary measure in assemblies." )]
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Annotations/Statics/Unaudited.cs
```diff
@@ -2,10 +2,27 @@
...
+
 	public static partial class Statics {
 		[Obsolete( "Static variables marked as unaudited require auditing. Only use this attribute as a temporary measure in assemblies." )]
 		[AttributeUsage( validOn: AttributeTargets.Field | AttributeTargets.Property )]
 		public sealed class Unaudited : Attribute {
+			public readonly Because m_cuz;
+
+			// extra-legacy parameterless constructor
+			public Unaudited() {
+				m_cuz = Because.ItHasntBeenLookedAt;
```
saquibmian: +1 backwards compatible ABI

j3parker:
I posted some usage guidelines, @saquibmian LMK if you think any are missing

saquibmian:
My only concern is the "global" `Because`, but this is temporary so I don't care enough.

j3parker:
Yeah I was being cheeky. Definitely temporary.

PR #55, opened 2017-02-14
Add Analyzer for named preperty 'Throws' of TestCaseData

LiyueWang:


src/D2L.CodeStyle.TestAnalyzers/TestCaseData/TestCaseDataAnalyzer.cs
```diff
@@ -0,0 +1,57 @@
...
+
+		private void AnalyzeSyntaxNode( SyntaxNodeAnalysisContext context ) {
+			var root = context.Node as InvocationExpressionSyntax;
+			if( root == null ) {
+				return;
+			}
+
+			if( root.Expression is MemberAccessExpressionSyntax ) {
+				var memberAccessExpression = root.Expression as MemberAccessExpressionSyntax;
+				var objectCreationExpressions = memberAccessExpression.DescendantNodes().OfType<ObjectCreationExpressionSyntax>().ToImmutableArray();
```
saquibmian: I don't understand how `ObjectCreationExpressionSyntax` is a descendent of `MemberAccessExpressionSyntax ` :/

I wonder if you should ignore the `ObjectCreationExpressionSyntax` at all, and just look to see if the method that's being called is one of the bad ones. I think you can do `SemanticModel.GetSymbolInfo( invocationExpression )` to pull out the symbol for the method being called, and then just make sure it's the one you care about. 

That will also handle the case where someone calls `.Throws` as part of a different expression.

saquibmian: (if that is super difficult, I don't care too much about it, these analyzers are temporary -- but I'm still confused about why the creation is a child of the member access).

tests/D2L.CodeStyle.TestAnalyzers.Tests/TestCaseData/TestCaseDataAnalyzerTests.cs
```diff
@@ -0,0 +1,121 @@
...
+		[Test]
+		public void DocumentWithThrowsCase2_Diag() {
+			const string test = @"
+	using System;
+
+	namespace test {
+		class Test {
+			static IEnumerable GetTranslateTestCases() {
+				yield return new TestCaseData( 0, 0 ).SetName( 'NoExistingResult_EmptyIdList_ExpectInvalidRequestDataException' ).Throws( typeof( Exception ) );
+				yield return new TestCaseData( 0, 500 ).SetCategory('').Throws( 'Exception' );
```
saquibmian: Do we have any places that call `.Throws` as part of a different expression? I.e., like so:

```csharp
var case = new TestCaseData(); 
case.Throws( typeof( Expression ) );
yield return case;
```

LiyueWang: Good point! I didn't check these cases.  will do, thanks!

src/D2L.CodeStyle.TestAnalyzers/TestCaseData/TestCaseDataAnalyzer.cs
LiyueWang: will try your suggestion, thanks!

saquibmian: Ahhh that makes a lot of sense. It's all one expression. I was expecting them to be siblings. 👍

tests/D2L.CodeStyle.TestAnalyzers.Tests/TestCaseData/TestCaseDataAnalyzerTests.cs
LiyueWang: Check that we don't have these cases in Lms, but still add to consider these cases in Analyzer

src/D2L.CodeStyle.TestAnalyzers/TestCaseData/TestCaseDataAnalyzer.cs
```diff
@@ -42,14 +42,29 @@ internal sealed class TestCaseDataAnalyzer : DiagnosticAnalyzer {
...
+					if( objectCreationExpressions.Length > 0 ) {
+						if( objectCreationExpressions[0].Type.ToString().Equals( "TestCaseData" ) ) {
+							var diagnostic = Diagnostic.Create( Rule, memberAccessExpression.Name.GetLocation() );
+							context.ReportDiagnostic( diagnostic );
+						}
+					} else {
+						var identifierNames = memberAccessExpression.DescendantNodes().OfType<IdentifierNameSyntax>().ToImmutableArray();
+						foreach( var identifierName in identifierNames ) {
+							var symbol = context.SemanticModel.GetSymbolInfo( identifierName ).Symbol;
+							if( symbol != null && symbol.DeclaringSyntaxReferences != null ) {
```
saquibmian: Does `symbol.Type.Name == "TestCaseSource"` work?

LiyueWang: No, `symbol` doesn't have the property `Type`

src/D2L.CodeStyle.TestAnalyzers/TestCaseData/TestCaseDataAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create( Rule );
+
+		public override void Initialize( AnalysisContext context ) {
+			context.RegisterSyntaxNodeAction(
+				AnalyzeSyntaxNode,
+				SyntaxKind.InvocationExpression
+			);
+		}
+
+		private void AnalyzeSyntaxNode( SyntaxNodeAnalysisContext context ) {
```
saquibmian: I think this method could be better written like this:

```csharp
private void AnalyzeSyntaxNode( SyntaxNodeAnalysisContext context ) {
	var root = context.Node as InvocationExpressionSyntax;
	if( root == null ) {
		return;
	}

	var symbol = context.SemanticModel.GetSymbolInfo( root );
	if( symbol.Symbol == null ) {
		return;
	}

	var method = symbol.Symbol as IMethodSymbol;
	if( method == null ) {
		return;
	}

	if( method.Name == "Throws" && method.ContainingType.Name == "TestCaseData" ) {
		var diagnostic = Diagnostic.Create( Rule, root.GetLocation() );
	}
}
```

This handles all cases of a method invocation expression and doesn't care so much about how the syntax is laid out. I tested it and it works locally, with some changes to the tests (to add a reference to NUnit and make the tes code actually compile). What do you think?

tests/D2L.CodeStyle.TestAnalyzers.Tests/TestCaseData/TestCaseDataAnalyzerTests.cs
```diff
@@ -0,0 +1,152 @@
...
+
+		[Test]
+		public void DocumentWithThrowsCase1_Diag() {
+			const string test = @"
+	using System;
+
+	namespace test {
+		class Test {
+			private TestCaseData[] UserDetailsTestSource = {
+				new TestCaseData(0,0 ).SetCategory('').SetName( 'NoExistingResult_500Ids_Success' ).Throws( typeof( Exception ) ),
```
saquibmian: The `'` is not valid here; those are characters, not strings.
- `SetCategory('')` no such method
- `SetName( 'NoExistingResult_500Ids_Success' )` not valid C# code (multicharacter character)
- `Throws('')` no such method

src/D2L.CodeStyle.TestAnalyzers/TestCaseData/TestCaseDataAnalyzer.cs
LiyueWang: Tested a little bit, for `var symbol = context.SemanticModel.GetSymbolInfo( root )`, the symbols for the InvocationExpressions of the tests in `TestCaseDataAnalyzerTests.cs` are always `null`?

saquibmian: Yup, you need to:
- add `using NUnit.Framework;` to each code snippet.
- add this to to class:
    ```csharp
		protected override MetadataReference[] GetAdditionalReferences() {
			return new MetadataReference[] { MetadataReference.CreateFromFile( @"..\..\..\..\packages\NUnit.2.6.4\lib\nunit.framework.dll" ) };
		}
    ````
- and modify `packages.config` in that project to also download NUnit 2.6.4.

LiyueWang: I see, thanks!

tests/D2L.CodeStyle.TestAnalyzers.Tests/D2L.CodeStyle.TestAnalyzers.Tests.csproj
```diff
@@ -56,6 +56,10 @@
       <Private>True</Private>
     </Reference>
     <Reference Include="Microsoft.CSharp" />
+    <Reference Include="nunit.framework, Version=2.6.4.14350, Culture=neutral, PublicKeyToken=96d09a1eb7f44a77, processorArchitecture=MSIL">
```
saquibmian: You still want to reference nunit 3, but keep both versions in `packages.config`.

tests/D2L.CodeStyle.TestAnalyzers.Tests/packages.config
```diff
@@ -1,13 +1,13 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.CodeAnalysis.Analyzers" version="1.1.0" targetFramework="net461" />
+  <package id="Microsoft.CodeAnalysis.Analyzers" version="1.0.0" targetFramework="net461" />
```
saquibmian: why the downgrade?

tests/D2L.CodeStyle.TestAnalyzers.Tests/D2L.CodeStyle.TestAnalyzers.Tests.csproj
LiyueWang: But one project cannot reference both nunit 3 versions at the same time?

tests/D2L.CodeStyle.TestAnalyzers.Tests/packages.config
LiyueWang: When I updated Nunit version to 2.6.4 it comes out a error to ask me to downgrade the version, but it should work with `1.1.0`, will change it back to `1.1.0`

saquibmian:
AppVeyor doesn't like the path stuff. This might work (but is ugly, but temporary so I'm okay with it):
```csharp
protected override MetadataReference[] GetAdditionalReferences() {
	return new MetadataReference[] { MetadataReference.CreateFromFile( Path.Combine(
		Path.GetDirectoryName( this.GetType().Assembly.Location ), @"..\..\..\..\packages\NUnit.2.6.4\lib\nunit.framework.dll"
	) ) };
}
```

PR #58, opened 2017-02-15
Switching counter to count usages of `[Unaudited]` instead.

saquibmian:
Counter will output usages and causes, but drops knowledge of the old causes of `TypeIsMutable` vs. `DeclarationIsMutable`. I don't think those were super valuable, but if we need them, we can easily add them back (e.g., `IPropertySymbol.IsReadOnly` in `AnalyzedStatic`'s constructor).

### DIFF
--- a/src/D2L.CodeStyle.UnsafeStaticCounter/CountingVisitor.cs
+++ b/src/D2L.CodeStyle.UnsafeStaticCounter/CountingVisitor.cs
@@ -1,0 +1,46 @@
+﻿using System;
+using System.Collections.Concurrent;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+
+namespace D2L.CodeStyle.UnsafeStaticCounter {
+	internal class CountingVisitor : SymbolVisitor {
+
+		private const string UnauditedAttributeName = "Unaudited";
+		private const string DefaultCause = "none";
+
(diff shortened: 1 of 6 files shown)


saquibmian:
@j3parker bump

src/D2L.CodeStyle.UnsafeStaticCounter/CountingVisitor.cs
```diff
@@ -0,0 +1,46 @@
...
+			if( unauditedAttribute == null ) {
+				return;
+			}
+
+			var cause = DefaultCause;
+			if( unauditedAttribute.ConstructorArguments.Length > 0 ) {
+				cause = unauditedAttribute.ConstructorArguments[ 0 ].Value.ToString();
+			}
+			AnalyzedStatics.Add( new AnalyzedStatic( symbol, cause ) );
+			Console.WriteLine( $"found one: {symbol}" );
```
j3parker: Will this show up in the Jenkins output?

j3parker: If it does we probably want to remove it.

saquibmian:
@j3parker wanna take a look at the refactor around aggregations again?

saquibmian:
@j3parker I think this is done, but sanity check?

PR #59, opened 2017-02-17
Add Analyzer for TestContext, check two fields 'Status' and 'State' changed in NUnit3

LiyueWang:


saquibmian:
Remind me again, is this because `Result` has been renamed?

LiyueWang:
- TestContext.CurrentContext.Result.State  => TestContext.CurrentContext.Result.Outcome
- TestContext.CurrentContext.Result.Status  =>  TestContext.CurrentContext.Result.Outcome.Status

saquibmian:
Awesome, thanks. Do you think we need to be concerned about this case?

```csharp
var context = TestContext.Current;
if ( context.Result.State == blah ) // ...
```

LiyueWang:
searched  "TestContext.CurrentContext", we don't have this case in Lms.  So I would not be much concerned about this

PR #62, opened 2017-02-27
Littlefoot/statics/fix test analyzers

LiyueWang:
- temporarily set ParallelizableTestsAnalyer as Warning instead of Error
- fix a few test analyzers bugs
    -- fix bug of null object reference in ExpectedExceptionAnalyzer and TestCaseAnalyer
 -- fix  IgnoreAttributeAnalyzer: check if the Ignore is under [ReflectionSerializer]
 -- fix TestCaseSourceAttributeAnalyer: consider the case that the source is the base class in another doc

### DIFF
--- a/src/D2L.CodeStyle.TestAnalyzers/ExpectedException/ExpectedExceptionAnalyzer.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/ExpectedException/ExpectedExceptionAnalyzer.cs
@@ -51,6 +51,9 @@ namespace D2L.CodeStyle.TestAnalyzers.ExpectedException {
 					isExpectedException = true;
 					break;
 				} else if( attribute.Name.ToString().Equals( "TestCase" ) ) {
+					if( attribute.ArgumentList == null ) {
+						continue;
+					}
 					foreach( var argument in attribute.ArgumentList.Arguments ) {
 						if( argument.NameEquals != null && argument.NameEquals.Name.ToString().Equals( "ExpectedException" ) ) {
 							location = argument.GetLocation();
--- a/src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
@@ -39,7 +39,7 @@ namespace D2L.CodeStyle.TestAnalyzers.IgnoreAttribute {
 			if( root == null ) {
 				return;
 			}
-			
+
 			foreach( var attribute in root.Attributes ) {
 				if( attribute.Name.ToString().Equals( "Ignore" ) ) {
 					var parentClassDeclaration = root.AncestorsAndSelf().OfType<ClassDeclarationSyntax>().ToImmutableArray().Single();
@@ -50,7 +50,7 @@ namespace D2L.CodeStyle.TestAnalyzers.IgnoreAttribute {
 							}
 						}
 					}
-					
+
 					if( attribute.ArgumentList == null ) {
 						var diagnostic = Diagnostic.Create( Rule, attribute.GetLocation() );
 						context.ReportDiagnostic( diagnostic );
--- a/src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
@@ -42,16 +42,12 @@ namespace D2L.CodeStyle.TestAnalyzers.IgnoreAttribute {
 
 			foreach( var attribute in root.Attributes ) {
 				if( attribute.Name.ToString().Equals( "Ignore" ) ) {
-					var parentClassDeclaration = root.AncestorsAndSelf().OfType<ClassDeclarationSyntax>().ToImmutableArray().Single();
-					foreach( var parentAttributeList in parentClassDeclaration.AttributeLists ) {
-						foreach( var parentAttribute in parentAttributeList.Attributes ) {
-							if( parentAttribute.Name.ToString().Equals( "ReflectionSerializer" ) ) {
-								return;
-							}
-						}
+					var memberGroups = context.SemanticModel.GetMemberGroup( attribute );
+					if( memberGroups.Length == 0 ) {
+						return;
 					}
 
-					if( attribute.ArgumentList == null ) {
+					if( memberGroups.First().ContainingType.ToString().Equals( "NUnit.Framework.IgnoreAttribute" ) && attribute.ArgumentList == null ) {
 						var diagnostic = Diagnostic.Create( Rule, attribute.GetLocation() );
 						context.ReportDiagnostic( diagnostic );
 					}
--- a/src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
@@ -1,4 +1,5 @@
 ﻿using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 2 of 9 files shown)


src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
```diff
@@ -41,6 +42,15 @@ internal sealed class IgnoreAttributeAnalyzer : DiagnosticAnalyzer {
 
 			foreach( var attribute in root.Attributes ) {
 				if( attribute.Name.ToString().Equals( "Ignore" ) ) {
+					var parentClassDeclaration = root.AncestorsAndSelf().OfType<ClassDeclarationSyntax>().ToImmutableArray().Single();
```
saquibmian: I think a better approach might be to check what the ignore attribute is on, and make sure it's a method. 
And to check to make sure that the attribute class's namespace is the nunit one.

saquibmian: Or check that the ignore attribute is accompanied by a `Test` attribute too.

tests/D2L.CodeStyle.TestAnalyzers.Tests/IgnoreAttribute/IgnoreAnalyzerTests.cs
```diff
@@ -37,6 +37,24 @@ class Test {
 		}
 
 		[Test]
+		public void DocumentWithoutTestIgnore_NoDiag() {
+			const string test = @"
+	using System;
+
+	namespace test {
+		[ReflectionSerializer]
```
saquibmian: You shouldn't need this. `ReflectionSerializer` is only for DTOs. See my comments above in terms of making sure `Ignore` is on a test method, and making sure the correct `Ignore` is being checked.

tests/D2L.CodeStyle.TestAnalyzers.Tests/SourceAttribute/TestCaseSourceAttributeAnalyzerTests.cs
```diff
@@ -190,6 +190,36 @@ class Test {
 			VerifyCSharpDiagnostic( test, diag1, diag2 );
 		}
 
+		[Test]
+		public void DocumentWithTestCaseSource_WithoutStatic_Case6_Diag() {
```
saquibmian: nice test!

src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
LiyueWang: Sorry, didn't get that, what do you mean *check what the ignore attribute is on*?

The ignore sometimes is not accompanied by a Test, i.e:
```
       [TestFixture]  
	[Ignore]   
	internal sealed class IgnoreSample {}
```

saquibmian: Sorry, I was confusing. I think just make sure that the correct attribute is being inspected. 
I.e., `NUnit.Framework.IgnoreAttribute`. The conflict you're getting with `ReflectionSerializer` is a different `Ignore`.

LiyueWang: gotcha, will do, thanks!

tests/D2L.CodeStyle.TestAnalyzers.Tests/IgnoreAttribute/IgnoreAnalyzerTests.cs
LiyueWang: I update the analyzer as your comments. Leaving this one as a test case that the ignore is not on a test method still makes sense?

saquibmian: Did you forget to push? :/
I think a test to make sure ignore doens't matter for non-test methods is cool, because it won't compile with nunit3 (reason is required), but the `ReflectionSerializer` part shouldn't matter.

LiyueWang: Pushed :)

src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
```diff
@@ -42,16 +42,12 @@ internal sealed class IgnoreAttributeAnalyzer : DiagnosticAnalyzer {
...
 			foreach( var attribute in root.Attributes ) {
 				if( attribute.Name.ToString().Equals( "Ignore" ) ) {
-					var parentClassDeclaration = root.AncestorsAndSelf().OfType<ClassDeclarationSyntax>().ToImmutableArray().Single();
-					foreach( var parentAttributeList in parentClassDeclaration.AttributeLists ) {
-						foreach( var parentAttribute in parentAttributeList.Attributes ) {
-							if( parentAttribute.Name.ToString().Equals( "ReflectionSerializer" ) ) {
-								return;
-							}
-						}
+					var memberGroups = context.SemanticModel.GetMemberGroup( attribute );
```
saquibmian: Interesting. I'm not super familiar with `GetMemberGroup`, but it seems to operate on method/property groups only. Why did you chose this one over `GetSymbolInfo`? (not saying you have to use the other one, just curious around the choice here).

LiyueWang: No special reason, just want to try something new to see if it works. does it make sense here?

src/D2L.CodeStyle.TestAnalyzers/ExpectedException/ExpectedExceptionAnalyzer.cs
```diff
@@ -51,6 +51,9 @@ public sealed class ExpectedExceptionAnalyzer : DiagnosticAnalyzer {
 					isExpectedException = true;
 					break;
 				} else if( attribute.Name.ToString().Equals( "TestCase" ) ) {
+					if( attribute.ArgumentList == null ) {
```
saquibmian: When does this actually happen?

tests/D2L.CodeStyle.TestAnalyzers.Tests/IgnoreAttribute/IgnoreAnalyzerTests.cs
saquibmian: I think we can make this test better like so:

`csharp
public class TestAttribute : Attribute { }
public class NotTest {
	[Test]
	public void NotTestFoo() { }
}
```
... should not report diagostic

src/D2L.CodeStyle.TestAnalyzers/IgnoreAttribute/IgnoreAttributeAnalyzer.cs
saquibmian: I think so, because the attribute is a method for the constructor. 👍

src/D2L.CodeStyle.TestAnalyzers/ExpectedException/ExpectedExceptionAnalyzer.cs
LiyueWang: When I built locally, I got the error message for LP. after I fix like this, the errors gone, I didn't track down to the specific code as the error message didn't show that...
>  [exec] CSC(0,0) : error AD0001: Analyzer 'D2L.CodeStyle.TestAnalyzers.TestCase.TestCaseAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.
> [exec] CSC(0,0) : error AD0001: Analyzer 'D2L.CodeStyle.TestAnalyzers.ExpectedException.ExpectedExceptionAnalyzer' threw an exception of type 'System.NullReferenceException' with message 'Object reference not set to an instance of an object.'.

LiyueWang: some case like `[TestCase]` ?

tests/D2L.CodeStyle.TestAnalyzers.Tests/IgnoreAttribute/IgnoreAnalyzerTests.cs
LiyueWang: make sense, will do!

src/D2L.CodeStyle.TestAnalyzers/ExpectedException/ExpectedExceptionAnalyzer.cs
saquibmian: Is that actually valid code? If so, you're right, we have to handle that :(

tests/D2L.CodeStyle.TestAnalyzers.Tests/IgnoreAttribute/IgnoreAnalyzerTests.cs
saquibmian: Oh sorry, `Ignore` instead of `Test`.

src/D2L.CodeStyle.TestAnalyzers/ExpectedException/ExpectedExceptionAnalyzer.cs
LiyueWang: test locally, hmm, it seems work with only `TestCase`

saquibmian: Yeah, grrr: http://livegrep.proddev.d2l:6080/?q=%5C%5BTestCase%5C%5D&i=nope&files=&repos=

PR #64, opened 2017-03-02
Upgrade everything to same version of Roslyn (1.2).

saquibmian:


saquibmian:
@LiyueWang there's a test broken here that I just realized is invalid C# code. I can't figure out what the test case is _supposed_ to be, could you take a look and fix in a different PR? Here's the test:

```csharp
	[Test]
	public void DocumentWithValueSource_WithoutStatic_Case2_Diag() {
		const string test = @"
using System;

namespace test {
	class Test {
		private String ValidCases {
			return new String(""test"");
		}

		public void test1( [ValueSource( ""ValidCases"" )] String s ) {

		}
	}
}";
		AssertSingleDiagnostic( test, 10, 24, "ValidCases" );
	}
```

saquibmian:
@LiyueWang @drose-d2l pointed out that Roslyn still emits an error because it seems to emit `<invalid-global-code>` for the symbol.
This _used_ to fail silently in 1.1, but we're updating all projects in the sln to use the same version (1.2).

PR #66, opened 2017-03-19
UnsafeStaticsAnalyzer now exposes better messages for why something is mutable.

saquibmian:
Examples of messages include:

```
The static field or property 'foo' is unsafe because 'foo.Bar''s type ('test.Tests.Bar') is not sealed.
The static field or property 'foo' is unsafe because 'foo.Bar''s type argument ('System.Object') is not sealed.
The static field or property 'foo' is unsafe because 'foo.Bar.Baz' is not readonly.
```

Sorry for the large PR, I could have broken this up into steps, but they wouldn't have made sense that way.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
@@ -51,24 +51,24 @@ namespace D2L.CodeStyle.Analyzers.Common {
 		/// <summary>
 		/// A list of immutable container types (i.e., types that hold other types)
 		/// </summary>
-		private static readonly ImmutableHashSet<string> ImmutableContainerTypes = new HashSet<string> {
-			"D2L.LP.Utilities.DeferredInitializer",
-			"System.Collections.Immutable.ImmutableArray",
-			"System.Collections.Immutable.ImmutableDictionary",
-			"System.Collections.Immutable.ImmutableHashSet",
-			"System.Collections.Immutable.ImmutableList",
-			"System.Collections.Immutable.ImmutableQueue",
-			"System.Collections.Immutable.ImmutableSortedDictionary",
-			"System.Collections.Immutable.ImmutableSortedSet",
-			"System.Collections.Immutable.ImmutableStack",
-			"System.Collections.Generic.IReadOnlyCollection",
-			"System.Collections.Generic.IReadOnlyList",
-			"System.Collections.Generic.IReadOnlyDictionary",
-			"System.Collections.Generic.IEnumerable",
-			"System.Lazy",
-			"System.Nullable",
-			"System.Tuple"
-		}.ToImmutableHashSet();
+		private static readonly ImmutableDictionary<string, string[]> ImmutableContainerTypes = new Dictionary<string, string[]> {
+			[ "D2L.LP.Utilities.DeferredInitializer" ] = new[] { "Value" },
+			[ "System.Collections.Immutable.ImmutableArray" ] = new[] { "[]" },
+			[ "System.Collections.Immutable.ImmutableDictionary" ] = new[] { "[].Key", "[].Value" },
+			[ "System.Collections.Immutable.ImmutableHashSet" ] = new[] { "[]" },
+			[ "System.Collections.Immutable.ImmutableList" ] = new[] { "[]" },
+			[ "System.Collections.Immutable.ImmutableQueue" ] = new[] { "[]" },
+			[ "System.Collections.Immutable.ImmutableSortedDictionary" ] = new[] { "[].Key", "[].Value" },
+			[ "System.Collections.Immutable.ImmutableSortedSet" ] = new[] { "[]" },
+			[ "System.Collections.Immutable.ImmutableStack" ] = new[] { "[]" },
+			[ "System.Collections.Generic.IReadOnlyCollection" ] = new[] { "[]" },
+			[ "System.Collections.Generic.IReadOnlyList" ] = new[] { "[]" },
+			[ "System.Collections.Generic.IReadOnlyDictionary" ] = new[] { "[].Key", "[].Value" },
+			[ "System.Collections.Generic.IEnumerable" ] = new[] { "[]" },
+			[ "System.Lazy" ] = new[] { "Value" },
+			[ "System.Nullable" ] = new[] { "Value" },
+			[ "System.Tuple" ] = new[] { "Item1", "Item2", "Item3", "Item4", "Item5", "Item6" }
+		}.ToImmutableDictionary();
 
 		/// <summary>
 		/// Determine if a given type is mutable.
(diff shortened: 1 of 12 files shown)


src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
```diff
@@ -75,84 +75,94 @@ public sealed class MutabilityInspector {
...
 						&& !type.IsSealed
 					) {
-					return true;
+					return MutabilityInspectionResult.Mutable( null, type.GetFullTypeName(), MutabilityTarget.Type, MutabilityCause.IsNotSealed );
 				}
 
-				foreach( ISymbol member in type.GetNonStaticMembers() ) {
-					if( IsMemberMutableRecursive( member, MutabilityInspectionFlags.Default, typeStack ) ) {
-						return true;
+				foreach( ISymbol member in type.GetExplicitNonStaticMembers() ) {
```
saquibmian: This change was required because when I started asserting causes in the tests, one of the tests was raising a diagnostic on a different cause than was expected (it was looking at backing fields for properties, where only the property really matters).

src/D2L.CodeStyle.Analyzers/Common/MutabilityInspector.cs
```diff
@@ -75,84 +75,94 @@ public sealed class MutabilityInspector {
...
 			}
 
 			if( IsTypeKnownImmutable( type ) ) {
-				return false;
+				return MutabilityInspectionResult.NotMutable();
 			}
 
 			if( type.TypeKind == TypeKind.Array ) {
-				return true;
+				return MutabilityInspectionResult.Mutable( null, type.GetFullTypeName(), MutabilityTarget.Type, MutabilityCause.IsAnArray );
```
saquibmian: The `null` names here are properly set by their the analyzer or by `InspectMemberRecursive`. We don't have the information here, but one level up the stack has it.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -67,7 +67,7 @@ public sealed class ImmutabilityAnalyzer : DiagnosticAnalyzer {
 				| MutabilityInspectionFlags.AllowUnsealed // `symbol` is the concrete type
 				| MutabilityInspectionFlags.IgnoreImmutabilityAttribute; // we're _validating_ the attribute
 
-			if( m_immutabilityInspector.IsTypeMutable( symbol, flags ) ) {
+			if( m_immutabilityInspector.InspectType( symbol, flags ).IsMutable ) {
```
saquibmian: We should expose the cause here too, but I will do that in a different PR.

tests/D2L.CodeStyle.Analyzers.Test/Common/MutabilityInspectionResultFormatterTests.cs
```diff
@@ -0,0 +1,79 @@
...
+
+		private readonly MutabilityInspectionResultFormatter m_formatter = new MutabilityInspectionResultFormatter();
+
+		private static readonly IEnumerable<TestCaseData> m_testCases = new[] {
+			new TestCaseData(
+				MutabilityInspectionResult.NotMutable(),
+				string.Empty
+			).SetName("not mutable"),
+
+            // field cases
```
saquibmian: formatting...I shall fix

saquibmian: I didn't test cases that are not "possible" to occur (e.g., a member that is unsealed is not possible, only a type is). I can test them, if you guys feel that's valuable.

tests/D2L.CodeStyle.Analyzers.Test/Common/MutabilityInspectorTests.cs
```diff
@@ -67,120 +67,190 @@ public class MutabilityInspectorTests {
...
+
+			var actual = m_inspector.InspectType( type );
 
-			Assert.IsFalse( m_inspector.IsTypeMutable( type ) );
+			AssertResultsAreEqual( expected, actual );
 		}
 
 		[Test]
-		public void IsTypeMutable_NullableNonPrimitiveType_True() {
+		public void InspectType_NullableNonPrimitiveType_NotMutable() {
```
saquibmian: I changed this test because (1) the behaviour of deep checking is already tested by other tests and (2) this test actually really cares about making sure that _nullable_ structs are inspected (`System.Nullable` is itself a struct).

tests/D2L.CodeStyle.Analyzers.Test/Common/MutabilityInspectorTests.cs
```diff
@@ -67,120 +67,190 @@ public class MutabilityInspectorTests {
...
 
-			Assert.IsTrue( m_inspector.IsTypeMutable( field.ContainingType ) );
+			var actual = m_inspector.InspectType( field.ContainingType );
+
+			AssertResultsAreEqual( expected, actual );
 		}
 
 		[Test]
-		public void IsTypeMutable_LooksAtMutabilityOfTypeOfField() {
+		public void InspectType_LooksAtMembersInExternalType() {
```
saquibmian: Changed the names of this test and the one above it because they were a little confusing/overlapped.

tests/D2L.CodeStyle.Analyzers.Test/Common/MutabilityInspectorTests.cs
```diff
@@ -67,120 +67,190 @@ public class MutabilityInspectorTests {
...
+			);
 
-			Assert.IsTrue( m_inspector.IsTypeMutable( property.Type ) );
+			var actual = m_inspector.InspectType( field.ContainingType );
+
+			AssertResultsAreEqual( expected, actual );
 		}
 
 		[Test]
-		public void IsTypeMutable_NonExistentType_ThrowsException() {
```
saquibmian: Killed this test as it was implemented incorrectly. I will add it back and implement it correctly, as we don't currently test that anywhere (I thought we did in `UnsafeStaticsAnalyzerTests`).

tests/D2L.CodeStyle.Analyzers.Test/UnsafeStatics/UnsafeStaticsAnalyzerTests.cs
```diff
@@ -92,7 +105,12 @@ internal sealed class Foo {
 
         }
     }";
-			AssertSingleDiagnostic( test, 11, 31, "bad", "it" );
+			AssertSingleDiagnostic( test, 11, 31, "bad", MutabilityInspectionResult.Mutable(
```
saquibmian: formatting...will fix

tests/D2L.CodeStyle.Analyzers.Test/UnsafeStatics/UnsafeStaticsAnalyzerTests.cs
```diff
@@ -444,11 +497,16 @@ internal class Foo {
                 public string ClientsName = ""YOLO"";
             }
 
-            public static Foo bad { get; }
+            public static Foo bad { get; } = new Foo()
```
saquibmian: This was another case where, once the reasons are exposed, the actual test is slightly different than expected. While the test is not wrong, I changed is to that it matches the same reason as its field equivalent (`DocumentWithStaticProperty_ReadonlyButMutable_Diag`).

tests/D2L.CodeStyle.Analyzers.Test/UnsafeStatics/UnsafeStaticsAnalyzerTests.cs
```diff
@@ -687,12 +765,16 @@ class Safe : IUnsafe {
 				public static readonly Safe Instance { get; } = new Safe();
 			}
 
-			private readonly static IUnsafe foo = Safe.Instance;
+			private readonly static IUnsafe foo = Safe.Instance; // bad, Safe is not sealed
```
saquibmian: note: `foo` is actually safe here, _because_ `Safe.Instance` is safe. But we don't currently expose other members' inspections (maybe we could via Roslyn annotations), soooo we don't allow it.

If you're thinking that adding this feature would make some types of errors go away (`ConstructorBinder<T>.AllFields`), unfortunately it won't, because we need source code for that level of analysis, which we don't have because of non-project references.

saquibmian:
@JeffAshton @j3parker you guys have time to look at this today?

PR #77, opened 2017-05-15
Disallow `System.Action` and `System.Func`.

saquibmian:
Because they hold on to state that we cannot analyze.


saquibmian:
I didn't, but good idea, I'll do that.

PR #296, opened 2018-02-14
Adding a new `ImmutableBaseClass` attribute.

saquibmian:
This attribute applies the rules of immutability to a base class, without requiring its implementations to also be immutable.
As such, a member of said type is not safe, unless it is known to be the concrete type.


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -118,10 +127,14 @@ HashSet<ITypeSymbol> typeStack
...
 
-			// If we're verifying immutability, then carry on; otherwise, bailout
-			if( !flags.HasFlag( MutabilityInspectionFlags.IgnoreImmutabilityAttribute ) && type.IsTypeMarkedImmutable() ) {
-				ImmutableHashSet<string> immutableExceptions = type.GetAllImmutableExceptions();
-				return MutabilityInspectionResult.NotMutable( immutableExceptions );
+			// If we're not verifying immutability, we might be able to bail out early
+			var scope = type.GetImmutabilityScope();
+			if( !flags.HasFlag( MutabilityInspectionFlags.IgnoreImmutabilityAttribute ) && scope != ImmutabilityScope.None ) {
+				// if we're fully immutable or we allow unsealed (i.e., base classes), then bailout
+				if( scope == ImmutabilityScope.SelfAndChildren || flags.HasFlag( MutabilityInspectionFlags.AllowUnsealed ) && scope == ImmutabilityScope.Self ) {
```
j3parker: I think this should be rewritten to not rely on implicit associativity rules.

```cs
if ( flags.HasFlag( MutabilityInspectionFlags.AllowUnsealed ) ) {
  if ( scope != ImmutabilityScope.None ) {
    // not mutable
  } 
} else if ( scope == ImmutablityScope.SelfAndChildern ) {
  // not mutable
}

// mutable
```
I'm going to submit a PR today that will start the ball rolling on removing `MutabilityInspectionFlags.AllowUnsealed`, so each of those branches are going to go to different functions anyway.

saquibmian: Sure, I can do that.

PR #301, opened 2018-03-05
Singleton registrations should require [Immutable], not [Singleton].

saquibmian:
I need some feedback on the test coverage...non-trivial change.

### DIFF
--- a/tests/D2L.CodeStyle.Analyzers.Test/Specs/DependencyRegistrationsAnalyzer.cs
+++ b/tests/D2L.CodeStyle.Analyzers.Test/Specs/DependencyRegistrationsAnalyzer.cs
@@ -121,87 +121,66 @@ namespace SpecTests {
 		public void DoesntMatter( IDependencyRegistry reg ) {
 
 			// Marked Singletons are not flagged.
-			reg.Register<IMarkedSingleton>( new MarkedSingleton() );
-			reg.Register( new MarkedSingleton() ); // inferred generic argument of above
-			reg.RegisterPlugin<IMarkedSingleton>( new MarkedSingleton() );
-			reg.RegisterPlugin( new MarkedSingleton() ); // inferred generic argument of above
-			reg.Register<ISingleton, MarkedSingleton>( ObjectScope.Singleton );
-			reg.RegisterFactory<ISingleton, ConcreteSingletonFactory>( ObjectScope.Singleton );
-			reg.RegisterPluginFactory<ISingleton, ConcreteSingletonFactory>( ObjectScope.Singleton );
-			reg.Register( typeof( IMarkedSingleton ), typeof( MarkedSingleton ), ObjectScope.Singleton );
-			reg.ConfigurePlugins<MarkedSingleton>( ObjectScope.Singleton );
-			reg.ConfigureOrderedPlugins<MarkedSingleton, SomeComparer<MarkedSingleton>>( ObjectScope.Singleton );
-			reg.ConfigureInstancePlugins<MarkedSingleton>( ObjectScope.Singleton );
-			reg.ConfigureInstancePlugins<MarkedSingleton, DefaultExtensionPoint<MarkedSingleton>>( ObjectScope.Singleton );
-			reg.RegisterPluginExtensionPoint<DefaultExtensionPoint<MarkedSingleton>, MarkedSingleton>( ObjectScope.Singleton );
-			reg.RegisterPlugin<DefaultExtensionPoint<MarkedSingleton>, IMarkedSingleton, MarkedSingleton>( ObjectScope.Singleton );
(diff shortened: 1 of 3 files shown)


JeffAshton:
Uhhh, when you register anything, 

```Register<T,...>```
```RegisterFactory<T,...>```
```RegisterPlugin<T,...>```

As singleton, the T should have ```[Singleton]```.  The Singleton attribute shouldn't be used for anything but these.  It's meant to signal that if I go ```IServiceLocator.Get<T>```, there is a corresponding registration of T for Singleton scope.

**Comment by saquibmian** [comment]
@JeffAshton that's what this PR is trying to do, but we are not going to require `[Singleton]` until someone calls `OldAndBrokenSingletonLocator.Get<T>`. Because there are many nuances where we don't want that attribute even if they register as singleton (mixed mode registrations). 

Also,
> It's meant to signal that if I go OldAndBrokentSingletonLocator.Get<T>, there is a corresponding registration of T for Singleton scope.

This is not true. You can put `[Singleton]` on things that are not registered at all. Eventually when Roslyn ships compile-time codegen, we will be able to enforce that as an analyzer, but for now if we wanted to do that, it would have to be an integration test.

**Comment by ChrisAtD2L** [comment]
From a purity perspective, I like the idea that every interface has a consistent scope (never mixing plugins of Singleton and WebRequest scopes).  I also like the idea of making it very clear on the interface itself when the type is/is not safe to inject into a Singleton.

But that is less important than finding the really bad abuses of our DI framework and fixing them.  So I agree with Saquib, that for now, all we need is "can't mark a WebRequest/Thread scoped object as [Singleton]" and "can't use a serviceLocator to get a non-Singleton".  That is enough to catch singletons relying on webrequests, which is our main concern right now.

**Comment by JeffAshton** [comment]
@saquibmian .  If and when Roslyn ships compile-time codegen, but for now, my statement holds true.

**Comment by JeffAshton** [comment]
@ChrisAtD2L   Never mixing plugins of Singleton and WebRequest scopes is an argument I don't get.  The person consuming the plugins doesn't care how long the object lives for, so long as it's safe.  Expecting anything else is violating the abstraction.  Just as when I write a webpage, I don't require all my dependencies to be WebRequest scoped.

**Comment by ChrisAtD2L** [comment]
@JeffAshton The problem with mixing plugin scopes comes from the insistence that:
1) [Singleton] attribute goes on interfaces, not concrete classes
2) Everything registered at Singleton scope should have the [Singleton] attribute
3) Anything registered at non-Singleton scope should not have the attribute

That leads to the situation where a mixed bag of plugins results in an interface that both must have the attribute, and cannot have the attribute.  A fancier solution might need to look at all of the DI registrations (across all assemblies) and choose the minimum lifecycle for that plugin type.  But that is beyond the scope of these one-project-at-a-time analyzers.  That's why having a consistent scope for plugins implementing the same interface seemed like a good goal.

Is there a better solution that I don't see?

**Comment by JeffAshton** [comment]
@ChrisAtD2L :  I think the problem is disagreement between these statements:

1.  [Singleton] attribute goes on interfaces, not concrete classes
2.  Everything registered at Singleton scope should have the [Singleton] attribute

I agree with statement Statement 1.  I disagree with Statement 2, and see things as

2.  Every implementation registered at Singleton should be immutable.  i.e. [Immutable]

This can easily be achieved with the current analyzers, and ensure no sticky data.

**Comment by saquibmian** [comment]
@JeffAshton that's exactly what this PR is doing, unless I'm confused.

**Comment by JeffAshton** [comment]
@saquibmian :  You're good.  I'm just communicating to Chris how I see the 2 different problem spaces.  We had an in-person conversation that I think helped.

File: tests/D2L.CodeStyle.Analyzers.Test/Specs/DependencyRegistrationsAnalyzer.cs [hunk 1]
```diff
@@ -232,42 +211,43 @@ public static class SomeExtensionMethods {
...
-	public sealed class UnmarkedSingleton : IUnmarkedSingleton {}
+	public interface IImmutableSubSingleton : IImmutableSingleton { }
+
+	[Objects.Immutable]
+	public sealed class ImmutableThing : ISingleton, INotSingleton { }
 
 	public interface ICreatedByDynamicFactory { }
 
 	public class ThingThatIsCreatedByDynamicObjectFactoryViaMarkedThing : ICreatedByDynamicFactory {
 		public ThingThatIsCreatedByDynamicObjectFactoryViaMarkedThing(
```
ChrisAtD2L: You changed the other names in this class to use "Immutable" instead of "Marked"/"Unmarked".  Similarly, we might want to change this to `CreatedByDynamicFactoryWithImmutableDependency" or something like that

saquibmian: Yes, I'll do that now.

saquibmian: Done.

PR #302, opened 2018-03-07
Create MutabilityInspector.InspectConcreteType

j3parker:
This moves all usage of AllowUnsealed into the mutability inspector. The
next step is to move the logic for AllowUnsealed "up" to
InspectConcreteType from InspectType and remove the flag entirely.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -231,21 +231,27 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			}
 		}
 
-		private MutabilityInspectionResult InspectBaseType(
+		public MutabilityInspectionResult InspectConcreteType(
 			ITypeSymbol type,
-			HashSet<ITypeSymbol> typeStack
+			MutabilityInspectionFlags flags
 		) {
-			if( type.BaseType == null ) {
-				return MutabilityInspectionResult.NotMutable();
-			}
+			var typesInCurrentCycle = new HashSet<ITypeSymbol>();
 
-			if( type.BaseType is IErrorTypeSymbol ) {
+			return InspectConcreteType( type, typesInCurrentCycle, flags );
+		}
+
+		private MutabilityInspectionResult InspectConcreteType(
+			ITypeSymbol type,
+			HashSet<ITypeSymbol> typeStack,
+			MutabilityInspectionFlags flags = MutabilityInspectionFlags.Default
+		) {
+			if( type is IErrorTypeSymbol ) {
 				return MutabilityInspectionResult.NotMutable();
 			}
 
 			return InspectType(
-				type.BaseType,
-				MutabilityInspectionFlags.AllowUnsealed,
+				type,
+				flags | MutabilityInspectionFlags.AllowUnsealed,
 				typeStack
 			);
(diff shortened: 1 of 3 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -231,21 +231,27 @@ HashSet<ITypeSymbol> typeStack
...
+			var typesInCurrentCycle = new HashSet<ITypeSymbol>();
 
-			if( type.BaseType is IErrorTypeSymbol ) {
+			return InspectConcreteType( type, typesInCurrentCycle, flags );
+		}
+
+		private MutabilityInspectionResult InspectConcreteType(
+			ITypeSymbol type,
+			HashSet<ITypeSymbol> typeStack,
+			MutabilityInspectionFlags flags = MutabilityInspectionFlags.Default
```
j3parker: This exists to support ImmutabilityAnalyzer passing `IgnoreImmutable`. I want to kill that flag too but that's a bit more work that makes sense after a bunch of other changes. The default means that internally things look mostly sane.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -281,11 +287,14 @@ HashSet<ITypeSymbol> typeStack
 				}
 
 				// We descend into the base class last
-				var baseResult = InspectBaseType( type, typeStack );
-				if( baseResult.IsMutable ) {
-					return baseResult;
+				if ( type.BaseType != null ) {
+					var baseResult = InspectConcreteType( type.BaseType, typeStack );
```
j3parker: I inlined a bit of logic here: checking that `type.BaseType` isn't `null` and pulling out `type.BaseType` here rather than inside the function. If you think this would look better if there was still a `InspectBaseType` helper (that calls `InspectConcreteType`) I can make that change.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
saquibmian: Might be nice at some point to make the private recursive ones named differently, like `InspectConcreteTypeRecursive`.

j3parker:
Opps I forgot to wait for your review @mthjones

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
mthjones: `type.BaseType` could still be `abstract`, I believe, but I don't think there's any functional issue if it was. I don't know if there's a more appropriate name for the function though.

j3parker: That's true. "Concrete type" is a D2L-ism that kinda bugs me but I haven't found a more suitable term looking at the literature.

j3parker: If you can think of a better name I'd be happy to switch stuff. Things I've considered (none:

* BaseType (already has a meaning in C# that doesn't precisely match)
* BaseClass (doesn't really make sense generally e.g. structs )
* SpecificType (name is... unspecific, no better than concrete type IMO)

PR #303, opened 2018-03-07
Move some checks from InspectType to InspectConcreteType (and remove AllowUnsealed guards)

j3parker:


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -94,6 +94,15 @@ IPropertySymbol property
 			return InspectProperty( property, typesInCurrentCycle );
 		}
 
+		public MutabilityInspectionResult InspectConcreteType(
```
j3parker: Moved up to live with the other `public` functions

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -231,16 +228,7 @@ HashSet<ITypeSymbol> typeStack
...
-			ITypeSymbol type,
-			MutabilityInspectionFlags flags
-		) {
-			var typesInCurrentCycle = new HashSet<ITypeSymbol>();
-
-			return InspectConcreteType( type, typesInCurrentCycle, flags );
-		}
-
-		private MutabilityInspectionResult InspectConcreteType(
+		private MutabilityInspectionResult InspectConcreteTypeImpl(
```
j3parker: @saquibmian looks like the current idiom in MutabilityInspector is to suffix the private versions with `Impl` (rather than `Recursive`... maybe I changed that awhile ago.) Anyway, made the name change.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -249,6 +237,17 @@ MutabilityInspectionFlags flags
 				return MutabilityInspectionResult.NotMutable();
 			}
 
+			// The concrete type System.Object is empty (and therefore safe.)
+			// For example, `private readonly object m_lock = new object();` is fine. 
+			if( type.SpecialType == SpecialType.System_Object ) {
```
j3parker: Note: the if statements no longer check for the `AllowUnsealed` flag

j3parker:
There are two remaining usages of `AllowUnsealed`:

* The `ImmutableBaseClass` logic: no biggy, just need to untangle that a bit.
* Wether to care about a `class` being `sealed`: this is going to require making `InspectConcreteType` not actually use `InspectType`. There will be a small amount of duplication (of the simple `switch` from `InspectType`) but I think it makes sense. The other thing that needs to be done is tweak how caching works (currently it's based on flags as the cache key. I think we want two caches (for just concrete analysis and for type analysis) and some smarts about one implying the other

j3parker:
Why this is safe: the only caller that passes `AllowUnsealed` down to `InspectType` is `InspectConcreteType`.

PR #304, opened 2018-03-08
Add analyzer to prevent accidental usage of ImmutableArray<T>'s defau…

j3parker:
There are more things I'd like to add to this analyzer later, but this is a useful start.


j3parker:
(p.s. this came up on a PR Owen was on. It resulted in a runtime error during the tests, though.)

j3parker:
I'm going to bump/update the LMS before merging this so we don't block getting `ImmutableBaseClass` in.

j3parker:
I'll add a fixit tomorrow

PR #307, opened 2018-03-09
Untangle/split some logic that was in InspectType into InspectConcreteType

j3parker:
This one requires a bit more lookin'. It might help to ignore what was there before an just look at what's there now.

Key fact: `AllowUnsealed` is only ever passed into `InspectType` by way of `InspectConcreteType`. So if we move logic predicated on `AllowUnsealed` into `InspectConcreteType` then the checks in `InspectType` become completely redundant.

The only concern is what we're commuting the check over in `InspectType` to get "up" to `InspectConcreteType` (i.e. what checks came before the thing we're moving, inside `InspectType`, that will now happen after... if the checks aren't commutative it isn't safe.)

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -138,15 +138,9 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 
 			// If we're not verifying immutability, we might be able to bail out early
 			var scope = type.GetImmutabilityScope();
-			if( !flags.HasFlag( MutabilityInspectionFlags.IgnoreImmutabilityAttribute ) && scope != ImmutabilityScope.None ) {
-				// if we're fully immutable or we allow unsealed (i.e., base classes), then bailout
-				if( scope == ImmutabilityScope.SelfAndChildren ) {
-					ImmutableHashSet<string> immutableExceptions = type.GetAllImmutableExceptions();
-					return MutabilityInspectionResult.NotMutable( immutableExceptions );
-				} else if( scope == ImmutabilityScope.Self && flags.HasFlag( MutabilityInspectionFlags.AllowUnsealed ) ) {
-					ImmutableHashSet<string> immutableExceptions = type.GetAllImmutableExceptions();
-					return MutabilityInspectionResult.NotMutable( immutableExceptions );
-				}
+			if( !flags.HasFlag( MutabilityInspectionFlags.IgnoreImmutabilityAttribute ) && scope == ImmutabilityScope.SelfAndChildren ) {
+				ImmutableHashSet<string> immutableExceptions = type.GetAllImmutableExceptions();
+				return MutabilityInspectionResult.NotMutable( immutableExceptions );
 			}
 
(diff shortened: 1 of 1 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -248,6 +242,12 @@ HashSet<ITypeSymbol> typeStack
 				return MutabilityInspectionResult.NotMutable();
 			}
 
+			var scope = type.GetImmutabilityScope();
+			if( !flags.HasFlag( MutabilityInspectionFlags.IgnoreImmutabilityAttribute ) && scope != ImmutabilityScope.None ) {
```
j3parker: If you're looking at a concrete type and its definition has either `[ImmutableBaseClass]` or `[Immutable]` then we're good (i.e. `[Immutable]` => `[ImmutableBaseClass]`)

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -138,15 +138,9 @@ HashSet<ITypeSymbol> typeStack
 
 			// If we're not verifying immutability, we might be able to bail out early
 			var scope = type.GetImmutabilityScope();
-			if( !flags.HasFlag( MutabilityInspectionFlags.IgnoreImmutabilityAttribute ) && scope != ImmutabilityScope.None ) {
-				// if we're fully immutable or we allow unsealed (i.e., base classes), then bailout
-				if( scope == ImmutabilityScope.SelfAndChildren ) {
-					ImmutableHashSet<string> immutableExceptions = type.GetAllImmutableExceptions();
-					return MutabilityInspectionResult.NotMutable( immutableExceptions );
-				} else if( scope == ImmutabilityScope.Self && flags.HasFlag( MutabilityInspectionFlags.AllowUnsealed ) ) {
```
mthjones: IIUC from the PR OP, this predicate would only ever be true when getting called from `InspectConcreteType`, so it's moved into `InspectConcreteType` directly (with the modification of also checking `SelfAndChildren` as a short-circuit). Makes sense.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
mthjones: Safe to assume `!= Immutability.None` will always be a valid scope here? Or better to be explicit about which scopes you want to test? I'm fine with either way.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
j3parker: Yep!

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
j3parker: Good point. I'm very confident that enum won't ever gain/lose members. I'm going to go with this for now. I might change how that stuff is structured 9the enum/helpers) after the flags are gone but still undecided.

PR #309, opened 2018-03-15
Temporarily disable new analyzer

j3parker:


src/D2L.CodeStyle.Analyzers/ApiUsage/System.Collections.Immutable/ImmutableCollectionsAnalyzer.cs
```diff
@@ -14,9 +14,10 @@ public sealed class ImmutableCollectionsAnalyzer : DiagnosticAnalyzer {
 
 		public override void Initialize( AnalysisContext context ) {
 			context.EnableConcurrentExecution();
-			context.RegisterCompilationStartAction( RegisterAnalysis );
-
+			// TODO: fix analyzer
+			//context.RegisterCompilationStartAction( RegisterAnalysis );
```
ChrisAtD2L: Looks like you changed the analyzer behaviour but not the test expectations, which caused CI to fail.

Is there a way to have an analyzer exist in the DLL, but not run it against the LMS?  Otherwise, we might have to turn off the tests as well.

j3parker: I switched to a straight revert. This analyzer is interesting but not critical at the moment.

j3parker:
Closing in favour of https://github.com/Brightspace/D2L.CodeStyle/pull/310

PR #310, opened 2018-03-15
Fix NullReferenceException in ImmutableCollectionsAnalyzer

j3parker:
This was due to assuming all type symbols that come out of an
`ObjectCreationSyntax` are `INamedTypeSymbol`s. An obvious other case is
`ITypeParameterSymbol` for generic things. I don't think the other cases
apply by inspection but regardless we'll ignore them.

These are the subclasses of `ITypeSymbol` today:
- **IArrayTypeSymbol**: not relevant because it doesn't come up for
  `ObjectCreationSyntax`, only `ArrayCreationSyntax`.
- **IDynamicTypeSymbol**: can't see how this could happen in
  `ObjectCreationSyntax`.
- **IPointerTypeSymbol**: not something D2L tends to use.

Source:
http://source.roslyn.io/#Microsoft.CodeAnalysis/Symbols/ITypeSymbol.cs,3dff140bf09f1a2d,references

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/System.Collections.Immutable/ImmutableCollectionsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/System.Collections.Immutable/ImmutableCollectionsAnalyzer.cs
@@ -48,6 +48,11 @@ namespace D2L.CodeStyle.Analyzers.ApiUsage.SystemCollectionsImmutable {
 			var specificType = context.SemanticModel.GetTypeInfo( context.Node ).Type
 				as INamedTypeSymbol;
 
+			// This happens for generic types
+			if ( specificType == null ) {
+				return;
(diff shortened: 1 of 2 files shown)


j3parker:
The parameterless constructor is the "default" one that all `struct`s get that can't be overridden (`ImmutableArray<T>` is a `struct` not a `class`.) It doesn't show up in our code base because any usage of it is a bug (this has caught a few people) because the `ImmutableArray<T>` it creates is in a permanently broken state (e.g. calling `.Length` would throw a NRE.)

Yeah probably, but I'm only trying to catch the most obvious cases. I could add that as a test case for something we knowingly don't catch.

ChrisAtD2L:
Thanks, I was wondering why I didn't see it in the code, but Visual Studio allowed it.  Makes more sense now.

j3parker:
Added your test case

PR #311, opened 2018-03-16
Fix a NRE in ImmutabilityAnalyzer caused by generics

j3parker:


src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -77,7 +77,14 @@ internal static partial class RoslynExtensions {
 			if( type.ContainingAssembly.TryGetImmutableGenericAnnotation( type, out ignore ) ) {
 				return true;
 			}
+
 			foreach( var typeArgument in type.TypeArguments ) {
+				// Can happen with generics, but I don't fully understand when
+				// it does, exactly.
```
j3parker: Left a comment because at some point I'll probably figure it out. the impact of `continue` here is that we don't consider something immutable which is safe at least.

j3parker:
Whoops forgot to commit the assertion.

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
ChrisAtD2L: Ok, that's probably the better way to fail (too restrictive instead of too permissive)

j3parker:
This may imply some limitations on the usefulness of `ImmutableGeneric` that will be a bit surprising but I need to do some tests.

PR #313, opened 2018-03-29
Fixing immutable generic

ChrisAtD2L:
The fix is small, but it was super annoying to find.

Basically, there was already an attempt to check for `ImmutableGeneric` in the `Except` analyzing code.  However, it used the wrong helper method to do the check, and ended up only checking the assembly containing the generic type, not the assembly containing the type param.  In practice, this meant that we could only do `ImmutableGeneric( type: typeof(IFactory<IFoo>))` in LP.Foundation where IFactory is defined.

On the spec tests, though, everything was in the same assembly, so the problem did not appear.

Further compounding this, when I tried to write tests to show the problem, the `Equals` method on `ITypeSymbol` was not behaving the way I expected it to.  When running against a real DLL in Visual Studio, it would correctly detect that the types were the same (between the inspected class and the ImmutableGeneric attribute on the class).  But in the tests, Equals was returning false, and I didn't see a reason why.  `ToString` returns a qualified type name, such as "SpecTests.GenericsTests.IFactory<System.Version>", which works in both tests and real code.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
+++ b/src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
@@ -60,6 +60,12 @@ namespace D2L.CodeStyle.Analyzers.Extensions {
 		}
 
 		private static bool IsMarkedImmutableGeneric( this ITypeSymbol symbol ) {
+			AttributeData ignore;
+			return TryGetImmutableGenericAnnotation( symbol, out ignore );
+		}
+
+		private static bool TryGetImmutableGenericAnnotation( this ITypeSymbol symbol, out AttributeData attribute ) {
+			attribute = null;
 			var type = symbol as INamedTypeSymbol;
 			if( type == null ) {
 				return false;
@@ -73,8 +79,7 @@ namespace D2L.CodeStyle.Analyzers.Extensions {
 			 *  (1) symbol's assembly,
 			 *  (2) any of symbol's type arguments's assemblies
 			 */
-			AttributeData ignore;
-			if( type.ContainingAssembly.TryGetImmutableGenericAnnotation( type, out ignore ) ) {
+			if( type.ContainingAssembly.TryGetImmutableGenericAnnotation( type, out attribute ) ) {
 				return true;
 			}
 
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -104,7 +109,8 @@ internal static partial class RoslynExtensions {
 				}
 
 				var arg = attr.ConstructorArguments[0];
-				if( arg.Value.Equals( type ) ) {
+				//Using ToString, otherwise it sometimes fails to match, and the test behaviour does not match the real behaviour
+				if( arg.Value.ToString().Equals( type.ToString() ) ) {
```
ChrisAtD2L: Not ideal, but without this I couldn't get the `ImplementsImmutableGeneric` test to work.  The ToString output was the same for the attribute and interface `ITypeSymbol`s ("System.IComparable<SpecTests.GenericsTests.ILocallyDefined>") but this line was returning false.

j3parker: What's the actual type of `Value` for that test? Trying to understand this.

j3parker:
nit: you don't need to fix it (particularly in this PR) but this repo sticks to 80 columns so that diffs are readable in GitHub (it's a very narrow viewport and horizontally scrolling sucks.)

I use https://marketplace.visualstudio.com/items?itemName=PaulHarrington.EditorGuidelines to draw a line in VS so that I can spot when to (manually) reformat my code.

j3parker:
Did you test this in a local LE build, using this new feature?

ChrisAtD2L:
@j3parker Yes, I discovered the problem by testing locally.  The first commit creates a test that demonstrates the same problem, and the second commit fixes the issue in both the test and a real DLL.  I was testing with D2L.LP.Diagnostics.Web (one of the PlatformTools DLLs)

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
ChrisAtD2L: When the test is running, it looks like `type` is of type SourceNamedTypeSymbol, and `arg.Value` is of type ConstructedNamedTypeSymbol

ChrisAtD2L: When running in the real DLL, both are of type ConstructedNamedTypeSymbol

j3parker: Hmmm....

PR #314, opened 2018-04-05
Adding WebRequest attribute analyzer

toddlang:


src/D2L.CodeStyle.Analyzers/Attributes.cs
```diff
@@ -24,6 +24,7 @@ internal static class Mutability {
 			internal static readonly RoslynAttribute Unaudited = new RoslynAttribute( "D2L.CodeStyle.Annotations.Mutability.UnauditedAttribute" );
 		}
 		internal static readonly RoslynAttribute Singleton = new RoslynAttribute( "D2L.LP.Extensibility.Activation.Domain.SingletonAttribute" );
+		internal static readonly RoslynAttribute WebRequest = new RoslynAttribute( "D2L.LP.Extensibility.Activation.Domain.WebRequestAttribute" );
```
j3parker: RE: the name
Maybe `[OnlyWebRequest]`? There are some types shared in the singleton and webrequest graph that are fine. `[WebRequest]` looks a lot like `[Singleton]` but we're adding semantics to it that I don't think would apply to all `[WebRequest]` types. I could be convinced otherwise though, I'll come over IRL.

RE: this attributes file
I think the stuff in this file and `RoslynAttribute` etc. are not very good and are perpetuating some bad behaviours. I'll bug you IRL about an alternative (which is simpler and doesn't involve any custom stuff.)

src/D2L.CodeStyle.Analyzers/D2L.CodeStyle.Analyzers.csproj
```diff
@@ -31,6 +31,7 @@
     <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup>
+    <Compile Include="DependencyScope\DependencyScopeAnalyzer.cs" />
```
j3parker: I'm going to be a jerk and come over and nit-pick about the name `DependencyScope` (the word scope) because I have some other analyzers in mind that I think could go into this category if the name is tweaked (and if they existed - there are some blockers but we have non-analyzers that check some properties of the graph that long-term could be awesome as analyzers.)

src/D2L.CodeStyle.Analyzers/DependencyScope/DependencyScopeAnalyzer.cs
```diff
@@ -0,0 +1,142 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+
+namespace D2L.CodeStyle.Analyzers.DependencyScope {
+
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	public sealed class DependencyScopeAnalyzer : DiagnosticAnalyzer {
```
j3parker: If this is the only purpose of the analyzer long-term (which would be fine) I'd go for a more specific name, like `OnlyWebRequestAnalyzer`

src/D2L.CodeStyle.Analyzers/DependencyScope/DependencyScopeAnalyzer.cs
```diff
@@ -0,0 +1,142 @@
...
+			// skip classes not marked singleton
+			if( !symbol.IsTypeMarkedSingleton() ) {
+				return;
+			}
+
+			AnalyzeType( context, root, symbol );
+		}
+
+		private void AnalyzeType( SyntaxNodeAnalysisContext context, ClassDeclarationSyntax root, ITypeSymbol type ) {
+			if( type.IsTypeMarkedWebRequest() ) {
```
j3parker: Nice

src/D2L.CodeStyle.Analyzers/DependencyScope/DependencyScopeAnalyzer.cs
```diff
@@ -0,0 +1,142 @@
...
+				var location = GetLocationOfClassIdentifierAndGenericParameters( root );
+				context.ReportDiagnostic( Diagnostic.Create(
+					Diagnostics.SingletonDependencyIsWebRequest,
+					location
+				) );
+			}
+
+			foreach( ISymbol member in type.GetExplicitNonStaticMembers() ) {
```
j3parker: I think you could not use this extension. It probably makes sense for static things too. I don't think you need to worry about implicitly declared members.

j3parker: i.e. I think you could just use `type.GetMembers()`

src/D2L.CodeStyle.Analyzers/DependencyScope/DependencyScopeAnalyzer.cs
```diff
@@ -0,0 +1,142 @@
...
+				if( member is IErrorTypeSymbol ) {
+					continue;
+				}
+
+				foreach (var memberSymbol in GetTypeSymbols( member ) ) {
```
j3parker: I find this kinda hard to read. I think it'd be clearer to make the body be a `switch ( member.TypeKind ) { ... }` rather than creating a `List<ITypeSymbol>`, looping over that and then `switch` ing. Also avoids the casting (better perf)

src/D2L.CodeStyle.Analyzers/DependencyScope/DependencyScopeAnalyzer.cs
```diff
@@ -0,0 +1,142 @@
...
+		}
+
+		private IEnumerable<ITypeSymbol> GetTypeSymbols(ISymbol symbol) {
+			ITypeSymbol result;
+			if( symbol is IFieldSymbol f ) {
+				result = f.Type;
+			} else if (symbol is IPropertySymbol p) {
+				result = p.Type;
+			} else {
+				result = default( ITypeSymbol );
```
j3parker: This will be `null`, which will cause a `NullReferenceException` on line 65, right?

src/D2L.CodeStyle.Analyzers/DependencyScope/DependencyScopeAnalyzer.cs
```diff
@@ -0,0 +1,142 @@
...
+			}
+
+			if (result is INamedTypeSymbol nt) {
+				if (nt.IsGenericType) {
+					return nt.TypeArguments;
```
j3parker: what about `nt` itself?

src/D2L.CodeStyle.Analyzers/DependencyScope/DependencyScopeAnalyzer.cs
```diff
@@ -0,0 +1,142 @@
...
+						location.SourceSpan.Start,
+						decl.TypeParameterList.GetLocation().SourceSpan.End
+					)
+				);
+			}
+
+			return location;
+		}
+
+		private static T GetDeclarationSyntax<T>( ISymbol symbol )
```
j3parker: Unused?

j3parker:
Cool! I'll come by

PR #317, opened 2018-04-11
[CustomerState] analyzer

toddlang:


toddlang:
I really have to figure out how to keep my local branch up to date with master properly. :/

src/D2L.CodeStyle.Analyzers/CustomerState/CustomerStateAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	public sealed class CustomerStateAnalyzer : DiagnosticAnalyzer {
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+			Diagnostics.SingletonDependencyHasCustomerState
+		);
+
+		public override void Initialize(
+			AnalysisContext context
+		) {
+			//context.EnableConcurrentExecution();
```
j3parker: Why commented out?

src/D2L.CodeStyle.Analyzers/CustomerState/CustomerStateAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+				return;
+			}
+
+			// skip classes not marked singleton
+			if( !symbol.IsTypeMarkedSingleton() ) {
+				return;
+			}
+
+			var inspectedTypes = new HashSet<ITypeSymbol>();
+			AnalyzeType( context, root, symbol, customerStateAttribute, inspectedTypes );
```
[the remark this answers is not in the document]
j3parker: The worst case runtime could, for example, be exponential in the depth of the object hierarchy (yes this is a very contrived example):

```cs
[Singleton]
public sealed class Class0 {
  public readonly Class1 M0;
  public readonly Class1 M1;
}

public sealed class Class1 {
  public readonly Class2 M0;
  public readonly Class2 M1;
}

public sealed class Class2 {
  public readonly Class3 M0;
  public readonly Class3 M1;
}

public sealed class Class3 { // n = 3
  public readonly Class4 M0;
  public readonly Class4 M1;
}

// ...
```

Inspecting `Class0` would take `O(2^n)` calls to `AnalyzeType` if `n` was the number of classes this went on for..

src/D2L.CodeStyle.Analyzers/CustomerState/CustomerStateAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+			if( type.IsTypeMarkedWith( customerStateAttribute ) ) {
+				var location = root.Identifier.GetLocation();
+
+				context.ReportDiagnostic( Diagnostic.Create(
+					Diagnostics.SingletonDependencyHasCustomerState,
+					location
+				) );
+			}
+
+			foreach( ISymbol member in type.GetMembers() ) {
```
j3parker: This will pick up `static` members but I think that's fine. It will also pick up "implicit" members, don't remember what they are but I'll chec tomorrow.

Note that `GetMembers()` will return fewer things if the type is defined in another assembly (e.g. pretty sure it doesn't return the `private` stuff?) This could lead to a situation where a public type XYZ that deep down uses a `[CustomerState]` thing might look safe in a different assembly if XYZ is not annotated.

Perhaps we should also enforce that all `public` things that hold a `[CustomerState]` deep down also be `[CustomerState]`. That would ensure we could do cross-assembly analysis I think.

src/D2L.CodeStyle.Analyzers/CustomerState/CustomerStateAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+			ITypeSymbol typeSymbol = default( ITypeSymbol );
+			switch( symbol.Kind ) {
+				case SymbolKind.Field:
+					typeSymbol = ( (IFieldSymbol)symbol ).Type;
+					break;
+				case SymbolKind.Property:
+					typeSymbol = ( (IPropertySymbol)symbol ).Type;
+					break;
+				default:
+					// We're not interested...bail out now.
```
j3parker: What about events?

Haha jk

src/D2L.CodeStyle.Analyzers/CustomerState/CustomerStateAnalyzer.cs
```diff
@@ -0,0 +1,164 @@
...
+					return result;
+			}
+
+			if( ( typeSymbol is INamedTypeSymbol nts ) && nts.IsGenericType ) {
+				// Basically if we can decompose type aggregated generic type
+				// then we'll return the base generic type and all of its
+				// type arguments
```
j3parker: Neat, ok

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -150,6 +150,31 @@ internal static partial class RoslynExtensions {
 			return false;
 		}
 
+		public static bool IsTypeMarkedWith(
```
j3parker: We have so many of these sort of extension methods!!! :'(

j3parker:
Ok overall I like the idea. Sorry about the delay, I can rebase your PR tomorrow morning. I think there was one thing we should maybe talk about in person tomorrow.

src/D2L.CodeStyle.Analyzers/CustomerState/CustomerStateAnalyzer.cs
toddlang: Whoops.  Let get the caching you wanted in, and whatever you wanted to talk about and we'll take another swing at this review.

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
toddlang: To be fair, I wrote this one so that we would stop writing IsTypeMarkedImmutable, IsTypeMarkedSingleton, etc.  You would just say IsTypeMarkedWith( Immutable ), IsTypeMarkedWith( Singleton ), but I also thought that modifying the existing code to use that would be outside the scope of this PR.

j3parker:
Hm, if you want to create a separate PR and get your diagnostic descriptor merged in first that could help with merge conflicts.

src/D2L.CodeStyle.Analyzers/CustomerState/CustomerStateAnalyzer.cs
```diff
@@ -0,0 +1,262 @@
...
+		}
+
+		private void RegisterAnalysis(
+			CompilationStartAnalysisContext context
+		) {
+			var singletonAttribute = context.Compilation.GetTypeByMetadataName(
+				"D2L.LP.Extensibility.Activation.Domain.SingletonAttribute" );
+
+			var customerStateAttribute = context.Compilation.GetTypeByMetadataName(
+				"D2L.LP.Extensibility.Activation.Domain.CustomerStateAttribute" );
```
j3parker: Could we make this live in `D2L.CodeStyle.Annotations` instead?

src/D2L.CodeStyle.Analyzers/CustomerState/CustomerStateAnalyzer.cs
```diff
@@ -0,0 +1,262 @@
...
+
+			if( singletonAttribute == default( INamedTypeSymbol )
```
j3parker: There's an extension method `.IsNullOrErrorType()` if that's the behaviour you're looking for.

src/D2L.CodeStyle.Analyzers/CustomerState/CustomerStateAnalyzer.cs
```diff
@@ -0,0 +1,262 @@
...
+			);
+		}
+
+		private void AnalyzeClass(
+			SyntaxNodeAnalysisContext context,
+			INamedTypeSymbol singletonAttribute,
+			INamedTypeSymbol customerStateAttribute,
+			ConcurrentDictionary<ITypeSymbol, CustomerStateAnalysis> previousAnalysis
+		) {
+			var root = context.Node as ClassDeclarationSyntax;
```
j3parker: You could just make this a hard cast because it won't fail.

omsmith:
Is this taint checking to some degree?

Would this enable say

```csharp
registry.Register<ITryProvider<ID2LPrincipal>, D2LPrincipalProvider>( ObjectScope.Singleton );

[CustomerState]
[Immutable]
public interface ID2LPrincipal { /* ... */ }

[Immutable]
[Singleton]
internal sealed class FooManager {

  private readonly ITryProvider<ID2LPrincipal> m_principalProvider;

  // Error D2L0036
  private readonly ID2LPrincipal m_principal;

  public FooManager(
    ITryProvider<ID2LPrincipal> principalProvider
  ) {
    m_principalProvider = principalProvider;

    // preventing this
   principalProvider.TryGet( out m_principal );
  }

  public DoThing() {
    // this is okay since m_principalProvider does HttpContext.Current internally
    if( m_principalProvider.TryGet( out currentPrincipal ) ) {
      DoThingWithCurrentPrincipal()
    }
  }

}
```

j3parker:
That's the idea yeah.

We've currently pivoted-ish to a thing where we run the constructors for singletons without access to customer state in prod. So catching stuff like this at runtime with some guarantees.

This approach is nice too (static analysis is cool) and we should probably pick it up again at some point.

omsmith:
Although I suppose if singletons must be Immutable then that is moot

PR #318, opened 2018-04-11
Remove duplicated logic in UnsafeStaticsAnalyzer

j3parker:
Next step for this analyzer is to merge it with the ImmutabilityAnalyzer to enable some future simplifications to MutabilityInspector (dropping the flags) and sharing a cache.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -103,6 +103,12 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			return InspectConcreteType( type, typesInCurrentCycle, flags );
 		}
 
+		public MutabilityInspectionResult InspectMember( ISymbol symbol ) {
+			var seen = new HashSet<ITypeSymbol>();
+
+			return InspectMemberRecursive( symbol, seen );
+		}
+
 		private MutabilityInspectionResult InspectType(
 			ITypeSymbol type,
 			MutabilityInspectionFlags flags,
--- a/src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
@@ -11,7 +11,6 @@ using Microsoft.CodeAnalysis.Diagnostics;
 namespace D2L.CodeStyle.Analyzers.Immutability {
 	[DiagnosticAnalyzer( LanguageNames.CSharp )]
 	public sealed class UnsafeStaticsAnalyzer : DiagnosticAnalyzer {
-
 		public const string PROPERTY_FIELDORPROPNAME = "FieldOrProprName";
 		public const string PROPERTY_OFFENDINGTYPE = "OffendingType";
 
@@ -53,7 +52,10 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			);
 		}
 
-		private void AnalyzeField( SyntaxNodeAnalysisContext context, MutabilityInspector inspector ) {
+		private void AnalyzeField(
+			SyntaxNodeAnalysisContext context,
+			MutabilityInspector inspector
+		) {
 			var root = context.Node as FieldDeclarationSyntax;
 
 			if( root == null ) {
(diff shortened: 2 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -103,6 +103,12 @@ IPropertySymbol property
 			return InspectConcreteType( type, typesInCurrentCycle, flags );
 		}
 
+		public MutabilityInspectionResult InspectMember( ISymbol symbol ) {
```
j3parker: Common pattern for `public` methods of this class is to hide the infinite-loop detector.

src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
```diff
@@ -75,20 +79,19 @@ public sealed class UnsafeStaticsAnalyzer : DiagnosticAnalyzer {
...
 					location: variable.GetLocation(),
 					isStatic: isStatic,
-					isReadOnly: isReadOnly,
-					fieldOrProperty: symbol,
 					fieldOrPropertyType: symbol.Type,
-					fieldOrPropertyName: variable.Identifier.ValueText,
-					initializer: variable.Initializer?.Value,
-					isProperty: false,
-					isAutoImplementedProperty: false
+					fieldOrPropertyName: variable.Identifier.ValueText
```
j3parker: A lot of things don't need to be passed down anymore, but I also tried to consistently sort the arguments in the different functions so some of the changes are just things moving around.

src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
```diff
@@ -251,73 +242,18 @@ string fieldOrPropertyName
...
-			ExpressionSyntax initializationExpression,
-			bool isProperty,
-			bool isAutoImplementedProperty
+			string fieldOrPropertyName
 		) {
 			if ( !isStatic ) {
 				yield break;
 			}
 
-			if( isProperty && !isAutoImplementedProperty ) {
```
j3parker: All of this logic is unnecessary because it's already implemented inside the mutability inspector (its here for historical reasons.)

j3parker:
This uncovered a bug in the mutability inspector because this analyzer now uses the mutability inspector slightly more than it did before. I noticed the bug because the (info) diagnostic about unnecessary static annotations fired for things it shouldn't have. I'll fix that bug in another PR. This one could merge before that's fixed but I'm undecided currently on if we should hold off.

src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
```diff
@@ -110,15 +113,11 @@ public sealed class UnsafeStaticsAnalyzer : DiagnosticAnalyzer {
...
+				fieldOrProperty: prop,
 				location: root.GetLocation(),
 				isStatic: isStatic,
-				isReadOnly: prop.IsReadOnly,
-				fieldOrProperty: prop,
 				fieldOrPropertyType: prop.Type,
-				fieldOrPropertyName: prop.Name,
-				initializer: root.Initializer?.Value,
-				isProperty: true,
-				isAutoImplementedProperty: isAutoImplementedProperty
```
mtjandra: Can remove line 111

```
bool isAutoImplementedProperty = root.IsAutoImplemented();
```

mtjandra:
> Tests are failing because I messed up the diagnostic formatting.

One thing I found odd is that the `MutabilityInspectionResult` doesn't contain all the information to generate the diagnostic, specifically the location. Is that intentional? My guess is because inspectors work off symbols, but the location comes from the declaration syntax? I'll have to check later but wondering if you knew off the top of your head

j3parker:
Nah I'm not really familiar with the diagnostic formatting parts.

j3parker:
Another change I'll make (in another PR) before merging this analyzer with the other one is getting rid of the `GatherDiagnostics` thing. It made more sense when `GatherDiagnostics` yielded different diagnostics/kinds of diagnostics but looks really dumb now that all the logic has moved into MI.

src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
```diff
@@ -251,76 +240,20 @@ string fieldOrPropertyName
...
-			// be paranoid about mutable derived classes.
-			if ( initializationExpression is ObjectCreationExpressionSyntax ) {
-				result = inspector.InspectConcreteType( fieldOrPropertyType );
-			} else {
-				result = inspector.InspectType( fieldOrPropertyType );
-			}
+			var result = inspector.InspectMember( fieldOrProperty );
 
 			if ( result.IsMutable ) {
-				result = result.WithPrefixedMember( fieldOrPropertyName );
```
j3parker: Removed this to get the diagnostics to render correctly; makes sense because we pushed these steps into the MI so this is now being done for us.

src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
j3parker: Thanks! Got it.

PR #321, opened 2018-04-11
Test Nested Mutability.Unaudited Reasons

mtjandra:
Adding this test to ensure when I choose to analyze past `Mutability.Unaudited` and `Mutability.Audited` we don't accidentally surface nested unaudited reasons. This would break current behaviour around the "Immutable Except" logic.

### DIFF
--- a/tests/D2L.CodeStyle.Analyzers.Test/Immutability/MutabilityInspectorTests.cs
+++ b/tests/D2L.CodeStyle.Analyzers.Test/Immutability/MutabilityInspectorTests.cs
@@ -478,6 +478,25 @@ sealed class Foo {
 		}
 
 		[Test]
+		public void InspectType_TypeHasUniqueNestedUnauditedMembers_DoesNotReturnNestedUnauditedReasonsAsSeenUnauditedReasons() {
+
+			const string source = AnnotationsPreamble + @"
+sealed class Foo {
+	[Mutability.Unaudited( Because.ItsSketchy )]
+	public Bar XYZ { get; }
+}
+
+sealed class Bar {
+	[Mutability.Unaudited( Because.ItsUgly )]
+	public Func<string> Baz { get; }
+}";
+
+			TestSymbol<ITypeSymbol> ty = CompileAndGetFooType( source );
+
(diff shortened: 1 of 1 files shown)


j3parker:
Neat! I like adding the tests first. That’s a subtle case you found too.

@mthjones is this the direction you’d expect for the test case? The background is that @mtjandra is going to start emitting diagnostics for unnecessary Mutability.* annotations like we do for Statics.* and it will require analyzing “past” the annotations.

j3parker:
Approved but let's wait for @mthjones to weigh in

mthjones:
Huh, this is actually a little surprising, but I think it's expected behaviour. I think there are two things at play here:
1) The immutability exception analysis intentionally doesn't check anything under something marked as `Unaudited`. If it's `Unaudited`, the assumption is that everything in that type tree is being considered unaudited, so I think the guarantee provided by the analyzer is still correct (maybe*).
2) We allow `[Mutability.Unaudited( ... )]` in types that aren't marked `[Immutable]` (directly or via inheritance). That's surprising behaviour to me, and maybe something you're planning on catching with the new diagnostic.

\*There is one interesting scenario that is brought to my attention here, where an `[Immutable]` type with an excepted unaudited field is held in a non-`[Immutable]` type, which is held in an `[Immutable]` type with a different unaudited reason.

i.e.
```csharp
[Immutable( Except = Except.ItsSketchy )]
public class Foo {
    [Mutability.Unaudited( Because.ItsSketchy )]
    private List<string> m_xs;
}

public class Bar {
    private Foo m_foo;
}

[Immutable( Except = Except.ItsUgly )]
public class Baz {
    [Mutability.Unaudited( Because.ItsUgly )]
    private Bar m_bar;
}
```

I don't expect that this scenario is common, but I think it's reasonable to consider this to be a significant flaw in the immutability exception logic. However, I don't know if there's any way to handle this. `Bar` could be defined in another assembly, which means that there's no way of knowing its internal implementation. We kind of just have to assume that the unaudited we can read is enough. This makes me sad...

Anyways, I think this new test looks legit. It's expected behaviour. 👍

PR #322, opened 2018-04-12
Add associated symbol to MutabilityInspectionResult

mtjandra:
Save the symbol that was inspected in the `MutabilityInspectionResult`. 

In next PR I can refactor properties `MemberPath` ~and `TypeName`~, should be able to get this from the symbol. After that we can get the violating location from the symbol. Then more to come...

With these changes when you have a `MutabilityInspectionResult.NotMutable` you will know what is not mutable (same goes for mutable). This approach should allow us to structure the `MutabilityInspectionResult` to be nested (can capture results past first violation). I am planning to get the location of the violation from the symbol itself.

I don't think it makes sense for the Mutability inspector to return a `IEnumerable<MutabilityInspectionResult>` because its recursive nature will make things confusing. For example `InspectClassOrStruct` might seem intuitive to return `IEnumerable<MutabilityInspectionResult>`  but I would not expect that return type for `InspectProperty`.

There should be no change is how current logic works, except that you are adding the appropriate symbol the the `MutabilityInspectionResult`. Hope the explanation conveys what I'm trying to do.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspectionResult.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspectionResult.cs
@@ -23,13 +23,13 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 
 	public sealed class MutabilityInspectionResult {
 
-		private readonly static MutabilityInspectionResult s_notMutableResult = new MutabilityInspectionResult( false, null, null, null, null, ImmutableHashSet<string>.Empty );
-
 		public bool IsMutable { get; }
 
 		public string MemberPath { get; }
 
 		public string TypeName { get; }
+
+		public ISymbol Symbol { get; }
 
 		public MutabilityCause? Cause { get; }
 
@@ -41,6 +41,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			bool isMutable,
 			string memberPath,
 			string typeName,
+			ISymbol symbol,
 			MutabilityTarget? target,
 			MutabilityCause? cause,
 			ImmutableHashSet<string> seenUnauditedReasons
@@ -48,27 +49,41 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			IsMutable = isMutable;
 			MemberPath = memberPath;
 			TypeName = typeName;
+			Symbol = symbol;
 			Target = target;
 			Cause = cause;
 			SeenUnauditedReasons = seenUnauditedReasons;
 		}
 
-		public static MutabilityInspectionResult NotMutable() {
-			return s_notMutableResult;
+		public static MutabilityInspectionResult NotMutable( ISymbol symbol ) {
+			return new MutabilityInspectionResult(
+				isMutable: false,
+				memberPath: null,
+				typeName: null,
+				symbol: symbol,
+				target: null,
+				cause: null,
+				seenUnauditedReasons: ImmutableHashSet<string>.Empty
+			); ;
 		}
 
-		public static MutabilityInspectionResult NotMutable( ImmutableHashSet<string> seenUnauditedReasons ) {
+		public static MutabilityInspectionResult NotMutable(
+			ISymbol symbol,
+			ImmutableHashSet<string> seenUnauditedReasons
+		) {
 			return new MutabilityInspectionResult(
-					false,
-					null,
-					null,
-					null,
-					null,
-					seenUnauditedReasons
+					isMutable: false,
+					memberPath: null,
+					typeName: null,
+					symbol: symbol,
+					target: null,
+					cause: null,
+					seenUnauditedReasons: seenUnauditedReasons
 				);
 		}
 
 		public static MutabilityInspectionResult Mutable(
+			ISymbol symbol,
 			string mutableMemberPath,
 			string membersTypeName,
 			MutabilityTarget kind,
@@ -78,6 +93,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				true,
 				mutableMemberPath,
 				membersTypeName,
+				symbol,
 				kind,
 				cause,
 				ImmutableHashSet<string>.Empty
@@ -89,6 +105,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			MutabilityCause cause
 		) {
 			return Mutable(
+				type,
 				null,
 				type.GetFullTypeName(),
 				MutabilityTarget.Type,
@@ -101,6 +118,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			MutabilityCause cause
 		) {
 			return Mutable(
+				field,
 				field.Name,
 				field.Type.GetFullTypeName(),
 				MutabilityTarget.Member,
@@ -113,6 +131,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			MutabilityCause cause
 		) {
 			return Mutable(
+				property,
 				property.Name,
 				property.Type.GetFullTypeName(),
 				MutabilityTarget.Member,
@@ -124,6 +143,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			ISymbol member
 		) {
 			return Mutable(
+				member,
 				member.Name,
 				null,
 				MutabilityTarget.Member,
@@ -144,6 +164,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				this.IsMutable,
 				newMember,
 				this.TypeName,
+				this.Symbol,
 				this.Target,
 				this.Cause,
 				this.SeenUnauditedReasons
@@ -155,6 +176,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				this.IsMutable,
 				this.MemberPath,
 				this.TypeName,
+				this.Symbol,
 				newTarget,
 				this.Cause,
 				this.SeenUnauditedReasons
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -382,12 +382,6 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			ExpressionSyntax expr,
 			HashSet<ITypeSymbol> typeStack
 		) {
-			if( expr.Kind() == SyntaxKind.NullLiteralExpression ) {
-				// This is perhaps a bit suspicious, because fields and
-				// properties have to be readonly, but it is safe...
-				return MutabilityInspectionResult.NotMutable( null );
-			}
-
 			var model = m_compilation.GetSemanticModel( expr.SyntaxTree );
 
 			var typeInfo = model.GetTypeInfo( expr );
@@ -396,6 +390,12 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			// the expression alone doesn't have a type. For example:
 			//   int[] foo = { 1, 2, 3 };
 			var exprType = typeInfo.Type ?? typeInfo.ConvertedType;
+
+			if( expr.Kind() == SyntaxKind.NullLiteralExpression ) {
+				// This is perhaps a bit suspicious, because fields and
+				// properties have to be readonly, but it is safe...
+				return MutabilityInspectionResult.NotMutable( exprType );
+			}
 
 			if( expr is ObjectCreationExpressionSyntax ) {
 				// If our initializer is "new Foo( ... )" we only need to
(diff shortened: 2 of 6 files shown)


tests/D2L.CodeStyle.Analyzers.Test/Immutability/UnsafeStaticsAnalyzerTests.cs
```diff
@@ -83,13 +83,15 @@ public sealed class Foo {
         }
     }";
             var diag1 = CreateDiagnosticResult( 20, 44, "Default", MutabilityInspectionResult.Mutable(
-                mutableMemberPath: "Default.uhoh",
+				null,
```
mtjandra: Symbol not used by analyzer yet

tests/D2L.CodeStyle.Analyzers.Test/Immutability/MutabilityInspectionResultFormatterTests.cs
```diff
@@ -9,88 +9,88 @@ class MutabilityInspectionResultFormatterTests {
 
 		private static readonly IEnumerable<TestCaseData> m_testCases = new[] {
```
mtjandra: Symbol not used by formatter yet

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -385,7 +385,7 @@ HashSet<ITypeSymbol> typeStack
 			if( expr.Kind() == SyntaxKind.NullLiteralExpression ) {
 				// This is perhaps a bit suspicious, because fields and
 				// properties have to be readonly, but it is safe...
-				return MutabilityInspectionResult.NotMutable();
+				return MutabilityInspectionResult.NotMutable( null );
```
mtjandra: Maybe move this after `var exprType = typeInfo.Type ?? typeInfo.ConvertedType;`, yeah I think I need to change this, and have test around it.

mtjandra: No that doesn't make sense. The typeInfo could just be say `string`, there isn't any location information associated with that.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspectionResult.cs
```diff
@@ -23,14 +23,14 @@ public enum MutabilityCause {
...
 
-		private readonly static MutabilityInspectionResult s_notMutableResult = new MutabilityInspectionResult( false, null, null, null, null, ImmutableHashSet<string>.Empty );
-
 		public bool IsMutable { get; }
 
 		public string MemberPath { get; }
 
 		public string TypeName { get; }
 
+		public ISymbol Symbol { get; }
```
j3parker: What is `Symbol`?

`MutabilityInspectionResult` is hellla confusing because its two cases merged into one (mutable vs. not mutable) but things like `TypeName` are self-explanitory if you notice its only used for `Mutable` (in which case its storing data for a particular reason the class is mutable (sadly we only return an arbitrary reason, not all the reasons.))

`Symbol` also seems to be being used for the "not mutable" case which is surprising - there are a lot of symbols that occur in the process of judging something as not mutable so why do we care about a particular one?

I haven't read the rest of the diff yet but this seems fishy. If it makes sense maybe we need to document it in a comment.

j3parker:
I'm a bit confused, I'll msg you on Slack in a bit.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -397,6 +391,12 @@ HashSet<ITypeSymbol> typeStack
 			//   int[] foo = { 1, 2, 3 };
 			var exprType = typeInfo.Type ?? typeInfo.ConvertedType;
 
+			if( expr.Kind() == SyntaxKind.NullLiteralExpression ) {
+				// This is perhaps a bit suspicious, because fields and
+				// properties have to be readonly, but it is safe...
+				return MutabilityInspectionResult.NotMutable( exprType );
```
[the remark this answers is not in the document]
mtjandra: I believe it you have `public readonly string SomeType = new RealType();"` then you will get a type symbol for `RealType`.
If you cannot get the type as the comment above explains, you should get the implicit type.
`int[] foo = { 1, 2, 3 };` Should give you type symbol for `int[]`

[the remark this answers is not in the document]
j3parker: But in this situation its `public readonly string SomeType = null`. You think it'll be `string`? Could you check? (and if it is, is it from `Type` or `ConvertedType`?)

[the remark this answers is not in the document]
mtjandra: Use a debugger and converted it uses `string` which comes from `ConvertedType`.

This should be covered in test [InspectField_StringInitializedNull_NotMutable](https://github.com/Brightspace/D2L.CodeStyle/pull/322/files#diff-0da4c090fb800cf892372a7b7ad9699dR333)

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspectionResult.cs
mtjandra: > What is Symbol?

I don't have the clearest understanding, I was actually going to ask you. My understanding is that is how the compiler represents things like `namespace, class, method, parametenamespace, class, method, parameter` (got that from the documented in the ISymbol interface). I believe it is anything we are interested in analyzing.

> Symbol also seems to be being used for the "not mutable" case which is surprising

I interpret it as after inspecting symbol `s` we found it to be immutable. I think of an inspector walking down some path and taking notes on every symbol it analyzes, it then marks each symbol as either mutable or immutable.  We want to have the symbol for the "not mutable" case because what if we happen to stumble on a `Mutability.Unaudited` and find out it is Immutable. We need to know that specific symbol was immutable, and there might be more than one, also mixed in with real violations.


> there are a lot of symbols that occur in the process of judging something as not mutable so why do we care about a particular one?

Perhaps there are some optimizations to be made once we start scanning past `Unaudited` and `Audited`, where we will only keep the mutability violators, and annotated violators and throw out the ones that are fine. But we need the care of a particular one because unlike before where we returned what could be thought of as one inspection result, soon we will be able to return an aggregate of inspection results.


We can talk tomorrow if I'm off or there is a better way.

mtjandra: Yes my original description is wrong `TypeName` must stay but `MemberPath` can be derived from the ISymbol

tests/D2L.CodeStyle.Analyzers.Test/Immutability/UnsafeStaticsAnalyzerTests.cs
mtjandra: Do you want tabs? I'm seeing a mix of tabs and spaces throughout the project

j3parker: D2L standard is tabs, feel free to reformat files as you see them.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspectionResult.cs
j3parker: > >   What is Symbol?
>
> I don't have the clearest understanding, I was actually going to ask you. My understanding is that is how the compiler represents things like namespace, class, method, parametenamespace, class, method, parameter (got that from the documented in the ISymbol interface). I believe it is anything we are interested in analyzing.

So `ISymbol` (not "Symbol") identifies a semantic object, that I'm cool with. What I was asking though is what is `MutabilityInspectorResult.Symbol` - how would you describe it? But that was addressed next:

> I interpret it as after inspecting symbol s we found it to be immutable. I think of an inspector walking down some path and taking notes on every symbol it analyzes, it then marks each symbol as either mutable or immutable. We want to have the symbol for the "not mutable" case because what if we happen to stumble on a Mutability.Unaudited and find out it is Immutable. We need to know that specific symbol was immutable, and there might be more than one, also mixed in with real violations.

Ok - so this is were I'm not following. There would be multiple symbols in that case, e.g.

```cs
[Mutability.Unaudited( ... )]
private readonly SafeThing m_safe;

// ...

public sealed class SafeThing {
    public readonly int m_immutableThing1;
    public readonly int m_immutableThing2;
}
```

`MutabilityInspectionResult` can only hold one symbol. Is your goal for it to hold the symbol for `SafeThing`, not `m_immutableThing1` or `m_immutableThing2` when we return from inspecting `SafeThing`? What about when we return from inspecting the `m_safe` field - should it be `m_safe` or `SafeThing`?

mtjandra: My intention is when inspecting `SafeThing` you get symbol for `SafeThing`, when inspecting `m_safe` you get symbol for `m_safe`. Internally it is a little more interesting than that because our current implementation for say `InspectField`, doesn't only verify if the field is mutable, it will also verify if its type is mutable. This means if you inspect `m_safe` you will get a symbol for `m_safe` if it is not readonly. But if when inspect `m_safe` it is `readonly` then you will get a symbol for `SafeThing` because that is the final thing you inspect.  That is how it would work now.

In the future I want to surface all results from the inspection. You could think of the current `MutabilityInspectionResult` to hold only one result, where I'd like to hold multiple results. You can think of it being restructured to as roughly follows (some fields omitted).

After: 
```
{  
  MutabilityInspectionResult:{  
    IsMutable: false,
    UserMarkedIgnored: true,
    Symbol: m_safe,
    Children:[  
      {  
        MutabilityInspectionResult:{  
          IsMutable:false,
          UserMarkedIgnored: false,
          Symbol: SafeThing,
          Children:[  
              {  
                MutabilityInspectionResult:{  
                  IsMutable:false,
                  UserMarkedIgnored: false,
                  Symbol:m_immutableThing1,
                  Children:[  

                  ]
                }
              },
              {  
                MutabilityInspectionResult:{  
                  IsMutable:false,
                  UserMarkedIgnored:false,
                  Symbol:m_immutableThing2,
                  Children:[  

                  ]
                }
              }
          ]
        }
      }
    ]
  }
}
```

Before:
```
{
  MutabilityInspectionResult: {
    IsMutable: false
  }
}
```

j3parker:
> My intention is when inspecting SafeThing you get symbol for SafeThing, when inspecting m_safe you get symbol for m_safe.

Right, so you wouldn't need to return it in that case. The caller already has that information (they passed it in in the first place - e.g. if you do `InspectType( <symbol for SafeThing> )` we don't need to return `<symbol for SafeThing>`.)

> our current implementation for InspectField, doesn't only verify if the field is mutable, it will also verify if its type is mutable.

This doesn't make sense: our/the definition for a field being immutable requires that its type also be immutable, that's not an "also" thing. You can't have an immutable field with a mutable type. Immutable is not the same thing as `readonly` (if it was we wouldn't have needed to write these analyzers :) )

> In the future I want to surface all results from the inspection. You could think of the current MutabilityInspectionResult to hold only one result, where I'd like to hold multiple results. You can think of it being restructured to as roughly follows (some fields omitted).

Think about `MutabilityInspectionResult` like this: you ask `MutabilityInspector` to look at something (e.g. a field, a class, a type, ...) and it gives you back an answer. The answer can be one of two things:

1. The thing you asked me about is immutable
   - We track every `Because` reason we see in any annotation.
   - (Soon, you're working on it) everytime we see an annotation we make a note if the field/property was actually immutable (i.e. if the annotation was unnecessary) but otherwise behave the same as before (ignore the immutability of the field/property)
2. The thing you asked me about is mutable
   - We track enough to format a diagnostic explaining why its mutable. Currently we track just one reason.

For what you're doing I'd expect a new property on `MutabilityInspector` named `UnnecessaryAnnotations` of type, maybe, `ImmutableHashSet<AttributeSyntax>`. It would only apply in the case of `NotMutable` and it would combine with unioning.

mtjandra:
Thanks for the feedback I'll give that a try, I have a couple questions.

Regarding the first case:
* If the thing I asked about has an annotated member, and the annotated member is found to be mutable, what should the value be for `isMutable`?
* If the thing I asked about is truly immutable and it has annotated members, how would I find where those unnecessary annotated member are and then later fix them if I only hold onto the `AttributeSyntax`? - I guess this is to be looked at in the future

Regarding the second case:
* If the thing I asked you about is mutable, since we only track one reason, we stop inspecting the rest of its members. This means the unnecessary mutability annotation diagnostics is dependent on the thing to be immutable. - I guess this doesn't matter since master should never emit error level diagnostics?

j3parker:
> If the thing I asked about has an annotated member, and the annotated member is found to be mutable, what should the value be for isMutable?

`false`: same as it is today

> If the thing I asked about is truly immutable and it has annotated members, how would I find where those unnecessary annotated member are and then later fix them if I only hold onto the AttributeSyntax? - I guess this is to be looked at in the future?

`AttributeSyntax` is a sibling of the field/property in the syntax tree. The thing that needs fixing though is the annotation not the member itself. It probably makes most sense to underline the attribute, and the fixit you'll have to write to delete the attribute is mostly going to care about the attribute syntax (other than whitespace.)

>  If the thing I asked you about is mutable, since we only track one reason, we stop inspecting the rest of its members. This means the unnecessary mutability annotation diagnostics is dependent on the thing to be immutable. - I guess this doesn't matter since master should never emit error level diagnostics?

Yep!

Issue #319, opened 2018-04-11
Immutability: analyzer assumes initializers are the only writes to readonly members


j3parker:
When the mutability analyzer sees an initializer for a field/property it assumes that its safe to inspect the (possibly narrower) type of the initializer rather than the type of field/property.

However there may be writes to the field/property after the initializer if there are constructors. An example:

```cs
public interface IFoo {}

[Immutable]
public sealed class Good : IFoo {}

public class Bad : IFoo { public int x; }

[Immutable]
public sealed Bar {
    public readonly IFoo m_foo = new Good();

    public Bar() {
        m_foo = new Bad(); // happens after the field initializer, not seen by the analyzer
    }
}
```

The analyzer judges `Bar` to be immutable because `Bar` is `sealed`, has an immutable base-class (`System.object`) and it's only field/property `Bar.m_foo` has an initializer of type `Good` which is immutable. However, the constructor overwrites `m_foo` with something of type `Bad` after the initializer...

We need to consider other writes to `m_foo`. Judging by the spec (mostly [here](https://github.com/dotnet/csharplang/blob/master/spec/classes.md#readonly-fields)) we can get away with finding each constructor and checking for these sorts of syntax:

1. `AssignmentExpressionSyntax`: use the same logic as initializers for the RHS expr.
2. `ArgumentSyntax` with `RefKindKeyword` of `SyntaxKind.OutKeyword` (`out` parameters): narrow to the type of the corresponding parameter.
3. `ArgumentSyntax` with `RefKindKeyword` of `SyntaxKind.RefKeyword` (`ref` parameters): narrow to the type of the corresponding parameter.

We could inspect each write independently and combine the results pessimistically (if any write introduces potential immutability, judge the member as mutable.)

We could do something fancier where we order the writes and determine that a particular write has to happen before another write (i.e. the field always gets overwritten) but I don't think that's likely to ever be helpful in practice so I don't think it's worth the trouble.

This was found while doing #318: since we're using the mutability inspector a *tiny* bit more it changed the behaviour enough that the unnecessary annotations diagnostic for static variables triggered for code like this:

```cs
public class Foo {
    [Statics.Unaudited( Because.ItsStickyDataOhNooo )]
    public static readonly IEvilThing m_evil = null;

    public static Foo() {
        m_evil = new ActualEvil();
    }
}
```

[...]

PR #323, opened 2018-04-12
Simplify UnsafeStaticsAnalyzer

j3parker:
There's still some more to go but this is getting pretty close. End goal is still to merge with `ImmutabilityAnalyzer` into one mega analyzer ;)

I tested this by turning the unnecessary annotations up to error level and building the LMS with these changes. The only diagnostics it emitted were due to the #319 bug.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
@@ -127,12 +127,8 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		///    members don't come from a base class/interface so we can't do
 		///    this generically.
 		/// 2. Fields can define multiple variables and we may wish to output
-		///    multiple diagnostics (individual ariables in a field declaration
+		///    multiple diagnostics (individual variables in a field declaration
 		///    may be alright because of their initializers.)
-		///
-		/// The arguments are organized roughly based on the common grammar of
-		/// fields and properties: attributes, modifiers, type, name,
-		/// initializer.
 		/// </summary>
 		private void InspectFieldOrProperty(
 			SyntaxNodeAnalysisContext context,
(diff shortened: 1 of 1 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
```diff
@@ -143,44 +139,36 @@ MutabilityInspector inspector
...
-					location,
-					fieldOrPropertyName
+			bool hasAnnotation = hasUnauditedAnnotation || hasAuditedAnnotation;
+
+			if ( !isStatic && hasAnnotation ) {
+				// non-statics don't need these annotations.
+				// TODO: Consider dropping the static-specific annotations in
+				// favour of Mutability.Audited/Unaudited. Will require some
+				// effort due to the reporting we do but it'll make things
+				// simpler once we merge the analyzers.
```
j3parker: Once you get the unnecessary diagnostics working for `Mutability.*` attributes I'll investigate this. The benefit would be simplifying the code (which is more obvious when considering how to merge the analyzers.)

src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
```diff
@@ -216,50 +216,17 @@ string fieldOrPropertyName
...
-
-		/// <summary>
-		/// All logic relating to either emitting or not emitting diagnostics other
-		/// than the ones about unnecessary annotations belong in this function.
-		/// This allows InspectMember to implement the logic around the unnecessary
-		/// annotations diagnostic. Any time we bail early in AnalyzeField or
-		/// AnalyzeProperty we risk not emitting unnecessary annotation
-		/// diagnostics.
-		/// </summary>
-		private IEnumerable<Diagnostic> GatherDiagnostics(
```
mtjandra: Nice, yeah I was wondering about this portion when I first looked at it.

mtjandra:
> There's still some more to go but this is getting pretty close. End goal is still to merge with ImmutabilityAnalyzer into one mega analyzer 

What the benefit to having a mega analyzer?

> I tested this by turning the unnecessary annotations up to error level and building the LMS with these changes. 

How are you doing this? Creating nuget package and referencing it during build from `references.ps1` ?

src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs
j3parker: It made sense at one point I think.

j3parker:
> What's the benefit to having a mega analyzer?

Shared cache (performance) and de-duplicating some logic. The analyzer then becomes "put the stuff that for whatever reason needs to be immutable through the inspector".

It would be a stepping stone to inferring `[Immutable]` for unsealed types (like interfaces) if they are `private`/`internal` but that's a bigger piece of work that I don't think is likely any time soon (not a lot of benefit for us at this point.)

> How are you doing this? Creating nuget package and referencing it during build from references.ps1 ?

Nope I just copy the DLL into the right subfolder of `checkout\packages\D2L.CodeStyle.Analyzers\` and do a quicky build.

PR #328, opened 2018-04-20
Adding mutability annotations to preamble

mtjandra:
Preparing for testing diagnostic `D2L0030`


j3parker:
I don't object to this but could the tests be spec tests instead?

mtjandra:
> Hm maybe we could just link in the annotations DLL instead. But not really keen on doing that for these tests if we don't have to.

Perhaps, I can look into it when we begin to emit the diagnostics. Previously I wasn't aware the spec tests were preferred over these

j3parker:
Oh and maintaining the line and column numbers for the start of the diagnostic is a real drag. Note these assert helpers don't check the full extents of the diagnostic and this has previously led to accidentally not underlining what we intended to.

j3parker:
Here's an example of a short spec file: https://github.com/Brightspace/D2L.CodeStyle/blob/master/tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutableCollectionsAnalyzer.cs

PR #329, opened 2018-04-23
Adding integration test for JsonParamBinderAnalyzerFixer

mtjandra:
Adding integration test for `JsonParamBinderAnalyzerFixer`. Paves the way for future `CodeFixProvider` tests.


tests/D2L.CodeStyle.Analyzers.Test/Verifiers/CodeFixVerifier.cs
```diff
@@ -0,0 +1,99 @@
...
+using System.Linq;
+using System.Threading;
+using D2L.CodeStyle.Analyzers.Test.Verifiers;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using NUnit.Framework;
+
+namespace D2L.CodeStyle.Analyzers.Verifiers {
+	internal abstract class CodeFixVerifier : DiagnosticVerifier {
```
mtjandra: This hierarchy probably doesn't make sense. Perhaps we should lift some of the plumbing into an abstract class called `Verifier`, which will allow us to retrieve documents, diagnostics, etc. `CodeFixVerifier` and `DiagnosticVerifier` only care about exposing its `Verify*` methods. Could do that in a separate PR.

mtjandra: @j3parker do you have a vision in mind on how that should work? Something like providing two files `Foo.cs` and `Foo.Fixed.cs`, apply our CodeFix on `Foo.cs` then compare the new source equals the source in `Foo.Fixed.cs`? Or something more granular as you do in the diagnostic specs?

PR #330, opened 2018-04-24
Missed a necessary exemption (ContentHttpHandlerMapper)

grant-cleary:
Unfortunately, when I was writing the analyzer I neglected to consider that `ContentHttpHandler` isn't injected, but rather is directly instantiated in `ContentHttpHandlerMapper`... Hence this addition to the whitelisted classes list.

Apologies if you already pushed a new version and now need to do so again, Jacob. :(


j3parker:
> Apologies if you already pushed a new version and now need to do so again, Jacob. :(

That's fine - have to do that pretty often anyway.

PR #331, opened 2018-04-25
Experimental "ID Type" parameter mismatch

omsmith:
This is much better solved with the id type structs and named parameters, but trying to detect cases where orgId and userId parameters were swapped, for instance

Many false positives AFAICT, laregly due to passing orgUnitId into orgId parameters, but orgId means orgUnitId.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/IdTypesAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/IdTypesAnalyzer.cs
@@ -1,0 +1,83 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+using System.Linq;
+using D2L.CodeStyle.Analyzers.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.Analyzers.ApiUsage {
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal sealed class IdTypesAnalyzer : DiagnosticAnalyzer {
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+			Diagnostics.IdTypeParameterMismatch
+		);
+
+		private static readonly ImmutableDictionary<string, ImmutableHashSet<string>> PARAM_TO_VARIABLE_BLACKLIST = new Dictionary<string, string[]> {
+			{ "userId", new[] { "orgId", "orgUnitId" } },
+			{ "orgId", new[] { "orgUnitId", "userId" } },
+			{ "orgUnitId", new[] { "userId" } },
+			
+		}.ToImmutableDictionary( x => x.Key.ToUpperInvariant(), x => x.Value.Select( y => y.ToUpperInvariant() ).ToImmutableHashSet() );
+
+		private static readonly IImmutableSet<string> INTERESTING_PARAM_NAMES = PARAM_TO_VARIABLE_BLACKLIST.Keys.ToImmutableHashSet();
+		private static readonly IImmutableSet<string> INTERESTING_VARAIBLE_NAMES = PARAM_TO_VARIABLE_BLACKLIST.SelectMany( x => x.Value ).ToImmutableHashSet();
+
+		public override void Initialize( AnalysisContext context ) {
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( RegisterIdTypedAnalyzer );
+		}
+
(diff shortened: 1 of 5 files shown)


src/D2L.CodeStyle.Analyzers/ApiUsage/IdTypesAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+
+			SeparatedSyntaxList<ArgumentSyntax> arguments = argumentList.Arguments;
+			foreach( var argument in arguments ) {
+				IdentifierNameSyntax argumentExpression = argument.Expression as IdentifierNameSyntax;
+				if( argumentExpression == null ) {
+					continue;
+				}
+
+				string variableName = argumentExpression.Identifier.Text;
+				string variableNameNormalized = variableName.ToUpperInvariant();
```
j3parker: Minor: could set the comparers on the dict/hashsets instead

j3parker:
Think it'd feasible to clean up the false positives?

omsmith:
I think time would be better spent moving everything to `UserId`/`OrgId`/`OrgUnitId` and then add an analyzer that requires things named `orgUnitId` be an `OrgUnitId` (VariableDeclartionSyntax, MethodDeclarationSyntax (parameters))?

omsmith:
That said, it's only about ~90 things that pop up. Obviously it's limited to things named exactly as I've checked for here. There's other stuff such as `courseOfferingId` or just `orgUnit` and whatnot.

src/D2L.CodeStyle.Analyzers/ApiUsage/IdTypesAnalyzer.cs
```diff
@@ -0,0 +1,83 @@
...
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.Analyzers.ApiUsage {
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal sealed class IdTypesAnalyzer : DiagnosticAnalyzer {
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+			Diagnostics.IdTypeParameterMismatch
+		);
+
+		private static readonly ImmutableDictionary<string, ImmutableHashSet<string>> PARAM_TO_VARIABLE_BLACKLIST = new Dictionary<string, string[]> {
```
j3parker: What if it was still `string -> [String]` but instead of mapping to a blacklist we map to synonyms, like this:

```cs
ImmutableDictionary<string, ImmutableHashSet<string>> idTypes = 
  new Dictionary<string, string[]> {
    { "userId", new [] { "user" } },
    { "orgId", new [] { "org" } },
    { "orgUnitId", new [] { "orgUnit", "course", "courseId", "semester", "department", "dept" } }, // I think some of these are used...
  }.ToImmutableDictionary(
    keySelector: x => x,
    keyComparer: StringComparer.InvariantCultureIgnoreCase,
    elementSelector: syn => syn.ToImmutableHashSet( StringComparer.InvariantCultureIgnoreCase )
  );
```

Then you could do

```cs
ImmutableArray<string> allIdLikeNames = IdTypes.Keys.Union(
  IdTypes.Values.SelectMany( x => x )
).ToImmutableArray();
```

and then finally

```cs
PARAM_TO_VARIABLE_BLACKLIST = idTypes.Keys
  .ToImmutableDictionary(
    keySelector: x => x,
    keyComparer: StringComparer.InvariantCultureIgnoreCase,
    elementSelector: canonicalIdName => allIdLikeNames
      .Where( idLikeName =>
        canonicalIdName != idLikeName
        && !idTypes[canonicalIdName].contains( idLikeName )
      )
  );
```
  
whew maybe I answered my own question

j3parker:
That would be an interesting approach. @toddlang was really keen on doing something there too...

omsmith:
I know when I uncovered the first case of this (that I had seen) I tried to go through and update everything. And then again a while later, and then again. There's a lot of params... and more complicated these false positive cases where `long orgId != OrgId orgId`, but instead should be `OrgUnitId orgUnitId`

j3parker:
You might be interested in https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46317.pdf

This is a general theoretically cool approach. It'd be fun to implement

TL;DR:

* create vertices for the arguments
* create verticies for the parameters
* create edges between the args and params as they exist in the code, score them by [edit distance](https://en.wikipedia.org/wiki/Edit_distance) (e.g. `orgId` and org are closer than `orgId` and `orgUnit`)
* add edges for alternatives when the alternative is feasible (type checks) and cheaper but only when the argument isn't explicitly named
* [find a minimum cost bipartite matching](https://en.wikipedia.org/wiki/Hungarian_algorithm)

If that results in something different, throw diagnostics. 

Implemented here: https://github.com/google/error-prone/tree/master/core/src/main/java/com/google/errorprone/bugpatterns/argumentselectiondefects

j3parker:
Decided not to deal with this specifically?

PR #332, opened 2018-04-25
D2L0033: Interface implementation parameter names should match the interface

omsmith:
… param names

- [x] .view files spitting out _Expected implementation parameter name to be 'rc'_ (Legit case, PR open)
- [x] Definitely cases where interface method isn't matched still
- [x] ~Cache interface method collection?~
- [x] Reduce noise for different casing

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
@@ -3,7 +3,6 @@ using System.Collections.Immutable;
 using System.Linq;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
-using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 
 namespace D2L.CodeStyle.Analyzers.Language {
--- a/src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
@@ -60,6 +60,11 @@ namespace D2L.CodeStyle.Analyzers.Language {
 						continue;
 					}
 
+					// don't know when this happens
+					if( implParameter.DeclaringSyntaxReferences.Length == 0 ) {
+						continue;
+					}
+
 					Location location = implParameter.DeclaringSyntaxReferences.First().GetSyntax().GetLocation();
 					context.ReportDiagnostic( Diagnostic.Create(
 						Diagnostics.InterfaceImplementationParamNameMismatch,
--- a/src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
@@ -1,0 +1,87 @@
+﻿using System;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.Analyzers.Language {
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal sealed class ParamNamesShouldMatchInterfaceAnalyzer : DiagnosticAnalyzer {
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+			Diagnostics.InterfaceImplementationParamNameMismatch
+		);
+
+		public override void Initialize( AnalysisContext context ) {
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( RegisterAnalyzer );
+		}
+
+		private static void RegisterAnalyzer( CompilationStartAnalysisContext context ) {
+			context.RegisterSyntaxNodeAction(
+				ctx => AnalyzeClassDefinition(
+					ctx,
+					( INamedTypeSymbol )ctx.SemanticModel.GetDeclaredSymbol( ctx.Node )
+				),
+				SyntaxKind.ClassDeclaration
+			);
+		}
+
+		private static void AnalyzeClassDefinition(
+			SyntaxNodeAnalysisContext context,
+			INamedTypeSymbol typeSymbol
+		) {
+			if( !typeSymbol.Interfaces.Any() ) {
+				return;
+			}
+
+			ImmutableArray<IMethodSymbol> interfaceMethods = CollectInterfaceMethods( typeSymbol );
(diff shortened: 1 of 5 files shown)


omsmith:
I think Visual Studio is done collecting errors now...

I'll round to 3000, and guesstimate at 50% being the .view file noise

src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
```diff
@@ -0,0 +1,156 @@
...
+					continue;
+				}
+
+				ImmutableArray<IParameterSymbol> classMethodParameters = classMethod.Parameters;
+				ImmutableArray<IParameterSymbol> interfaceMethodParameters = interfaceMethod.Parameters;
+				for( int i = 0; i < classMethodParameters.Length; ++i ) {
+					string classMethodParameterName = classMethodParameters[ i ].Name;
+					string interfaceMethodParameterName = interfaceMethodParameters[ i ].Name;
+
+					if( classMethodParameterName == interfaceMethodParameterName ) {
```
omsmith: Consider case-insensitive match to reduce not-all-that-important reports

src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
```diff
@@ -0,0 +1,156 @@
...
+			context.RegisterCompilationStartAction( RegisterAnalyzer );
+		}
+
+		private static void RegisterAnalyzer( CompilationStartAnalysisContext context ) {
+			context.RegisterSymbolAction(
+				ctx => AnalyzeClassDefinition(
+					ctx,
+					(INamedTypeSymbol)ctx.Symbol
+				),
+				SymbolKind.NamedType
```
j3parker: IRL @omsmith mentioned this came from the sealed analyzer. Maybe looking at `ClassDeclarationSyntax` instead would be good (and seems like it would apply to the sealed analyzer too...)

Downsides to named type:

- non-class things
- external symbols: we shouldn't bother checking these
- duplicate instances of a named type (would be fixed by caching?)

still would want to cache, maybe, due to `partial` classes (wouldn't want to emit diagnostics multiple times.)

src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
```diff
@@ -0,0 +1,156 @@
...
+					s.MethodKind == MethodKind.ExplicitInterfaceImplementation
+					|| (
+						s.MethodKind == MethodKind.Ordinary
+						&& s.DeclaredAccessibility == Accessibility.Public
+					)
+				)
+				.ToImmutableArray();
+		}
+
+		private static IMethodSymbol MatchMethod(
```
j3parker: Maybe the SemanticModel has stuff to help with this, e.g. getting the declaration for that symbol?

j3parker:
Thoughts on how to roll it out?

omsmith:
1. Fix the easy easy cases (PRs open for `Compare`, `Equals`, and `GetHashCode`
2. Greylist existing files
3. Emit warning instead of error for greylist
4. Get a co-op (im-a-jerk-alert) to go through the greylist?

src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
```diff
@@ -0,0 +1,95 @@
...
+		private static ImmutableArray<IMethodSymbol> CollectInterfaceMethods(
+			INamedTypeSymbol typeSymbol
+		) {
+			var builder = ImmutableArray.CreateBuilder<IMethodSymbol>();
+
+			foreach( INamedTypeSymbol @interface in typeSymbol.AllInterfaces ) {
+				ImmutableArray<ISymbol> members = @interface.GetMembers();
+				ImmutableArray<IMethodSymbol> methods = members
+					.Where( s => s.Kind == SymbolKind.Method )
+					.Cast<IMethodSymbol>()
```
j3parker: You could maybe do

```cs
.OfType<IMethodSymbol>()
```

rather than

```cs
.Where( s => s.Kind == SymbolKind.Method )
.Cast<IMethodSymbol>()
```

but maybe what you wrote is faster? No preference just thinking out loud

omsmith:
I've seen one or two obviously wrong cases thus far in my random clicking, but definitely will take time so sort through

src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
```diff
@@ -0,0 +1,95 @@
...
+					.Where( s => s.MethodKind == MethodKind.Ordinary )
+					.ToImmutableArray();
+
+				foreach( var method in methods ) {
+					builder.Add( method );
```
j3parker: Minor: [`ImmutableArray<T>.Builder.AddRange`](https://msdn.microsoft.com/en-us/library/dn628328(v=vs.111).aspx) there's an overload that takes `IEnumerable<T>` so it'd be a tiny bit better than the extra `.ToImmutableArray()` above I think.

j3parker:
Sweet. I'm happy for this to get in ASAP. Nice analyzer.

j3parker:
RE: "Cache interface method collection?" I think maybe it's not a big deal... I think the semantic model caches a lot of the heavy stuff. I have no strong opinion.

j3parker:
This will hit the same class multiple times because of `partial` classes but maybe it's fine...

src/D2L.CodeStyle.Analyzers/Language/ParamNamesShouldMatchInterfaceAnalyzer.cs
```diff
@@ -0,0 +1,87 @@
...
+		) {
+			if( !typeSymbol.Interfaces.Any() ) {
+				return;
+			}
+
+			ImmutableArray<IMethodSymbol> interfaceMethods = CollectInterfaceMethods( typeSymbol );
+			foreach( var interfaceMethod in interfaceMethods ) {
+				IMethodSymbol implMethod = (IMethodSymbol)typeSymbol.FindImplementationForInterfaceMember( interfaceMethod );
+
+				// In the middle of coding and haven't implemented an interface method yet
```
j3parker: Its possible to add something to the spec for this

omsmith: Tru, that's where I saw it

j3parker:
P.S. https://docs.microsoft.com/en-us/visualstudio/code-quality/ca1725-parameter-names-should-match-base-declaration

I haven't really been paying attention to the built-ins...

omsmith:
Yeah, I figured it might exist, but we don't tend to have any built-ins enabled right?

j3parker:
Nope. I don't really know how that works.

omsmith:
So I'm thinking (and in the process of) just `#pragma warning disable` all the current cases individually? Super easy to find that way...?

j3parker:
Maybe you could disable it at the assembly level to start if you were worried about noise

omsmith:
Could also land as warning first?

j3parker:
Yes that'd be ok.

j3parker:
`-werror`? I've landed as `Info` before.

PR #333, opened 2018-04-25
Add first-pass at RequireNamedArgsAnalyzer

j3parker:
Over time I'd like to expand this:
- shrink the limit from 10 to something like 5
- two args of the same type (or convertable?) => name them
- literal args must be named

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -281,5 +281,15 @@ namespace D2L.CodeStyle.Analyzers {
 			isEnabledByDefault: true,
 			description: "ICorsHeaderAppender should not be used, as it can introduce security vulnerabilities."
 		);
+
+		public static readonly DiagnosticDescriptor UseNamedArgsForInvocationWithLotsOfArgs = new DiagnosticDescriptor(
+			id: "D2L0032",
+			title: "This function takes lots of arguments which makes it confusing. Please used named arguments so future readers can figure it out and so that it's clearer that the arguments are in the right order.",
+			messageFormat: "This function takes lots of arguments which makes it confusing. Please used named arguments so future readers can figure it out and so that it's clearer that the arguments are in the right order.",
+			category: "Readability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true,
+			description: "This function takes lots of arguments which makes it confusing. Please used named arguments so future readers can figure it out and so that it's clearer that the arguments are in the right order."
+		);
 	}
 }
--- a/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
@@ -14,7 +14,7 @@ namespace D2L.CodeStyle.Analyzers.Language {
 		);
 
 		// TODO: shrink this number over time. Maybe 5 would be good?
-		public const int TOO_MANY_UNNAMED_ARGS = 10;
+		public const int TOO_MANY_UNNAMED_ARGS = 64;
 
 		public override void Initialize( AnalysisContext context ) {
 			context.EnableConcurrentExecution();
--- a/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
@@ -91,13 +91,14 @@ namespace D2L.CodeStyle.Analyzers.Language {
 					allowParams: false
 				);
 
-				// We presumably can't name this param anyway
+				// Not sure if this can happen but it'd be hard to name this
+				// param so ignore it.
 				if ( param == null ) {
 					continue;
 				}
 
 				// IParameterSymbol.Name is documented to be possibly empty in
-				// which case it is unnamed, so ignore it.
+				// which case it is "unnamed", so ignore it.
 				if ( param.Name == "" ) {
 					continue;
 				}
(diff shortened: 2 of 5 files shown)


j3parker:
I haven't tested that 10 is a safe start in the LMS but I'm hoping that's big enough. Will report back before any merging.

j3parker:
No fixer yet

j3parker:
@omsmith ok if you think the plan makes sense I can bump the limit to 63 and fix the tests so they know about the new limit.

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -0,0 +1,106 @@
...
+		private static IEnumerable<ArgumentSyntax> GetUnnamedArgs(
+			SemanticModel model,
+			ArgumentListSyntax args
+		) {
+			foreach( var arg in args.Arguments ) {
+				// Ignore args that are already named. This will mean that args
+				// can be partially named which is sometimes helpful (the named
+				// args have to be at the end of the list though.)
+				if ( arg.NameColon != null ) {
+					continue;
```
omsmith: nit: could this theoretically just be a `return;` instead of of a `continue;` since named args need to be trailing?

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -281,5 +281,15 @@ public static class Diagnostics {
 			isEnabledByDefault: true,
 			description: "ICorsHeaderAppender should not be used, as it can introduce security vulnerabilities."
 		);
+
+		public static readonly DiagnosticDescriptor UseNamedArgsForInvocationWithLotsOfArgs = new DiagnosticDescriptor(
+			id: "D2L0032",
```
omsmith: Not a big deal since can find-and-replace easy, but I'm currently using 32 in my pragma stuff. Perhaps best practice to just use `D2LXXXX` or something until landing?

tests/D2L.CodeStyle.Analyzers.Test/Specs/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+				a5: 5,
+				a6: 6,
+				a7: 7,
+				a8: 8,
+				a9: 9,
+				a10: 10
+			);
+			#endregion
+
+			#region named args don't count against the unnamed args budget
```
omsmith: I find this questionable, but cool to start, and maybe reasonable as the limit gets dropped

tests/D2L.CodeStyle.Analyzers.Test/Specs/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -0,0 +1,79 @@
...
+			_arg10( 1, 2, 3, 4, 5, 6, 7, 8, 9, a10: 10 );
+			_arg11( 1, 2, 3, 4, 5, 6, 7, 8, 9, a10: 10, a11: 11 );
+			#endregion
+
+			/* UseNamedArgsForInvocationWithLotsOfArgs */ _arg11( 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, a11: 11 ) /**/;
+
+			#region params don't count against the unnamed args budget
```
omsmith: Future: require all non-params args be named?

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
j3parker: Oh true. I'll do that and leave a comment.

tests/D2L.CodeStyle.Analyzers.Test/Specs/RequireNamedArgumentsAnalyzer.cs
j3parker: Yeah, I am of two minds. I sometimes do it like [here](https://github.com/j3parker/D2L.CodeStyle/blob/a486ce8d7162a23f21fed6ad09d50e0eea2981c0/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs#L79) when the first args are obvious but then there are defaulted arguments, etc.

Not super attached to it.

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
j3parker: That makes sense. I think the test that checks for unique ids will still pass because they are opaque strings.

tests/D2L.CodeStyle.Analyzers.Test/Specs/RequireNamedArgumentsAnalyzer.cs
j3parker: Could you explain this more?

P.S. with named args and params I think you end up doing `foo( arg0: blah, arg1: blah2, paramArgs: new [] { 1, 2, 3} );` which is pretty ugly (the `new [] { }` part)

omsmith: `foo( a, b, c, x, y, z )` where did the params start?
alt being `foo( c, x, y, z, arg0: a, arg1: b )`, dunno if that actually works, or if it's better, but just a thought

omsmith: Or if it matters

j3parker: That's a compiler error because the first argument gets implicitly named `arg0`... error is "Named argument 'arg0' specifies a parameter for which a positional argument has already been given"

j3parker:
I stole your diagnostic id ;)

j3parker:
Oh, one thing I've done/seen done is to PR the diagnostic independently from the impl. That might be an ok practice. Easy to trace it down later if it never gets implemented.

PR #334, opened 2018-04-26
Show Analyzer Bug in Test

mtjandra:
Since the statics analyzer uses the mutability inspector to determine if it is safe, `Mutability` annotations bypasses the static analyzer check. This test represents the bug. Maybe we should only the the `Mutability` annotation and remove the `Statics` one? Or just not consider this a bug?


j3parker:
The mutability attribute isn't meant to be used on static variables. It only gets used when inspecting a type and looking at its *non-static* members. We skip the static members because they don't belong to instances of those types which is what we're analyzing. We have a separate analyzer that grabs every static variable and inspects it individually. We have a separate set of annotations for them (`[Statics.Unaudited(...)]` and `[Statics.Audited(...)]`.)

So this isn't a bug it's by design.

PR #335, opened 2018-04-26
D2LXXXX: Super brute "Detecting Argument Selection Defects"

omsmith:
**Current status**:
- [ ] Probably slow still for the whole LMS, but haven't tried yet since clamping it down to 6 args. Building eP was more than reasonable time-wise
- [ ] Needs a better cost algorithm, see [Section 5 of the paper](https://static.googleusercontent.com/media/research.google.com/en//pubs/archive/46317.pdf) where they go into details. Levenschtein on its own is full of false positives

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
@@ -139,17 +139,22 @@ namespace D2L.CodeStyle.Analyzers.Language {
 				var remainingArgs = args.RemoveAt( i );
 
 				for( int j = 0; j < @params.Length; ++j ) {
-					ImmutableArray<Edge> perm = current.Add( new Edge(
+					var assignment = new Edge(
 						model,
 						args[ i ],
 						@params[ j ]
-					) );
+					);
+
+					// Don't continue with impossible branches
+					if( assignment.Cost == int.MaxValue ) {
+						continue;
+					}
 
 					Permutate(
 						model,
 						remainingArgs,
 						@params.RemoveAt( j ),
-						perm,
+						current.Add( assignment ),
 						result
 					);
 				}
--- a/src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
@@ -407,6 +407,7 @@ namespace D2L.CodeStyle.Analyzers.Language {
 			}
 		}
 
+		#region Levenschtein Distance
 		// Adapted from https://en.wikipedia.org/wiki/Levenshtein_distance
 		private static int LevenshteinDistance( string x, string y ) {
 			int m = x.Length;
@@ -444,6 +445,7 @@ namespace D2L.CodeStyle.Analyzers.Language {
 
 			return v0[ n ];
 		}
+		#endregion
 
 
 		#region Hungarian Algorithm
--- a/src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
@@ -16,7 +16,7 @@ namespace D2L.CodeStyle.Analyzers.Language {
 		);
 
 		public override void Initialize( AnalysisContext context ) {
-			// context.EnableConcurrentExecution();
+			context.EnableConcurrentExecution();
 			context.RegisterCompilationStartAction( RegisterAnalyzer );
 		}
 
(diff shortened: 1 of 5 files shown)


omsmith:
It's a little slow - who'da thunk.

Surely can do a better job than my generate all permutations approach, but I wonder if it will be feasible as a real-time/build-time analyzer at all?

src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
```diff
@@ -0,0 +1,211 @@
...
+
+namespace D2L.CodeStyle.Analyzers.Language {
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal sealed class LikelyArgumentMismatchAnalyzer : DiagnosticAnalyzer {
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+			Diagnostics.LikelyArgumentMismatch
+		);
+
+		public override void Initialize( AnalysisContext context ) {
+			// context.EnableConcurrentExecution();
```
j3parker: Why?

j3parker:
Idea: bail if there are too many arguments, because soon we'll require named arguments for those anyway

src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
omsmith: Debugging

omsmith:
As per the paper, Levenschtein on its own is insufficient

j3parker:
What about implementing the Hungarian algorithm? @travisMiehm was noodling on that a bit but I think he didn't want to commit to doing it.

travisMiehm:
I haven't had time to read the math behind the algorithm to see if there's a way to augment it so that we can keep our "do not increase edge cost" criteria.

Also, my hastily written implementation will infinitely loop sometimes, and I haven't fixed it yet :(

j3parker:
If edit distance is bigger, make cost be "infinity"?

omsmith:
"Do not increase edge cost" isn't necessarily a rule? It's probably likely an indicator of correctness though.

One method, which as far as I can think, kind of gets you the same thing without being an explicit rule, is building in a preference to non-changes into your costs. As an example, the exact value needing to be relative to other costs, but you can add a base cost of 10 for making a change, thus the other differences need to outweigh making the change:
```
int GetCost( arg, param ) {
  int cost = 0;

  if (!isSourceMatching) {
    cost += 10;
  }

  cost += EditDistance(...);

  cost += TypeDifference(...);
}
```

travisMiehm:
Oh, https://www.nuget.org/packages/HungarianAlgorithm/ looks like it should just work.  We can build the cost matrix your cost function, and 💥

j3parker:
We can't (easily) bring in a nuget package here for boring administrative reasons but it looks like the license should be compatible with copy+paste (attribution would still be good in a comment.)

I'd give the code a once-over but... neat!

j3parker:
The fact that it is all array based is very promising.

src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
```diff
@@ -0,0 +1,773 @@
...
+			if( bestMatching == null ) {
+				return;
+			}
+
+			ImmutableDictionary<ArgumentDetails, Edge> sourceEdges = sourceMatching
+				.ToImmutableDictionary( x => x.Arg, x => x );
+			foreach( var edge in bestMatching ) {
+				Edge sourceEdge = sourceEdges[ edge.Arg ];
+				if( edge.Param != sourceEdge.Param ) {
+					context.ReportDiagnostic( Diagnostic.Create(
```
zommerfelds: Will this generate two warnings? Should we have a `break` once if finds one diagnostic?

src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
```diff
@@ -0,0 +1,773 @@
...
+				v0 = v1;
+				v1 = tmp;
+			}
+
+			return v0[ n ];
+		}
+
+
+		#region Hungarian Algorithm
+		// https://github.com/vivet/HungarianAlgorithm/blob/master/HungarianAlgorithm/HungarianAlgorithm.cs
```
zommerfelds: Would it be easier to split this into a separate file?

j3parker:
If this works it could be really good for interface (or base) method parameter to impl parameter matching. Treat the impl the same as you are treating call sites, but none of the "args" are named and all are effectively identifiers.

At some point it'd be nice to split this PR up a bit. E.g. one for levenstien and one for the hungarian algorithm (with tests :P.) After the fancy bits get in then the core algorithm could be outside the analyzer and the interface takes in `IEnumerable<(ISymbol, IParameterSymbol)>` (but a better type) in an abstract way and returns a suggested permutation.

src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
j3parker: Yeah I think they're just fooling around right now :)

travisMiehm: Almost certainly.  I just wanted this in to see how effective it would be

omsmith:
Rebased and force-pushed to fix the diff

src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
```diff
@@ -0,0 +1,669 @@
...
+			double cost = 10;
+			double costMultiplier = 1;
+
+			Conversion conversion = model.ClassifyConversion( argument.Syntax, parameter.Type );
+			if( !conversion.Exists ) {
+				// if the types don't match, make the cost large so we don't choose it
+				return int.MaxValue;
+			}
+
+			if( !conversion.IsIdentity && !conversion.IsUserDefined ) {
```
travisMiehm: What does it meant for `conversion.IsIdentity` to be `true`?

src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
```diff
@@ -0,0 +1,669 @@
...
+
+			return matching;
+		}
+
+		private static int[,] GetCostMatrix(
+			SemanticModel model,
+			ImmutableArray<ArgumentDetails> arguments,
+			ImmutableArray<ParameterDetails> parameters
+		) {
+			int[,] matrix = new int[ arguments.Length, parameters.Length ];
```
travisMiehm: This needs to be padded with `int.MaxValue` to be a square matrix, otherwise the algo won't work

src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
omsmith: Means no conversion. `long -> long`

src/D2L.CodeStyle.Analyzers/Language/LikelyArgumentMismatchAnalyzer.cs
travisMiehm: I'm not sure why you'd believe me either!

Hm, rethinking this, I'm not sure we actually want to augment the matrix to be square, given what the cols/rows represent.  I think `params` and optional args are the cases we have to worry about?  Basically, how can we have a method invocation with more arguments than the function has parameters (`params`) or vice versa (  default values ).

For `params`, I think it would be unnecessary to suggest different orders.  Not really sure about optionals.

omsmith: Hmm, so I changed it back to only include args/params that were matched by non-excluded arguments. For instance, "trusting the dev" with a named parameter - should trust them for the arg _and_ the parameter.

Wasn't thinking about what this would look like for `params` or default values though... feels like defaulted parameters should definitely make their way in... perhaps similarly parameterss that had a literal passed.

So, somewhere in between what the two of us had would be good I think.

omsmith:
> Nice work guys! I read the paper earlier this week and it seems useful.
> I'm not super familiar with the D2L analyzer framework and I'm curious about the following:

> 1. Will this only issue warnings or will it force you to fix it?
> 2. Is there a way to ignore false positives?
> 3. Did you guys run this on the whole codebase and discover issues?
> 4. Did you guys consider using the false positive detection heuristics in the paper?
> 5. What's the running time of this? Does it perform well?

1. Currently set to an error, meaning it would force you to fix it
2. Currently doesn't inspect named arguments, which means naming your arguments is an out. There's also always `#pragma ignore` statements
3. I've not yet run an accurate version of this on the codebase (since it doesn't really exist yet), but yes, problems this might find do exist.
4. I've not read that part of the paper, but I believe it's referencing another layer of infrastructure surrounding code analysis that we don't have.
5. My initial lazy implementation was somewhere around **O(n!)**, I believe it's currently closer to **O(n<sup>4</sup>)**

zommerfelds:
@omsmith Thanks!
For the running time I asked more about a practical "it takes 1h to run on the code base" vs "1 min". But I guess we'll know when we eventually run this on the codebase!
It would take some time for me to ramp up to our tools but let me know if I can help.

zommerfelds:
For question 4, I think that some heuristics might be not as hard to implement. The obviously easy one is to check for regex's in the function name (e.g. `backward(s)?`, `swap(ped)?`, `transpose(d)?`, etc). The other ones involve looking at the AST and checking if there are other function calls nearby or if we are in a recursive call.
Either we, we should probably test out the algorithm on our code before implementing any heuristic. We can probably come up with a few of our own to remove false positives.

Unrelated question: if it finds positives, does that mean that we need to fix all of them before enabling the analyzer? Or is there a way to gradually enforce it for new code?

j3parker:
> Unrelated question: if it finds positives, does that mean that we need to fix all of them before enabling the analyzer? Or is there a way to gradually enforce it for new code?

So, fixing would be adding named arguments for the most part. This matches what Google does with comments (because Java doesn't have named arguments proper.) I don't think `#pragma` would be appropriate. But yeah, either the code base would need to be clean for this to merge in. We don't have a way to stage analyzers currently.

Other thoughts:

## perf
This should be fast because it runs on essentially every key press in VS. It is running a non-linear time algorithm but its non-linear in the number of arguments. We now set a maximum limit of 19 unnamed arguments in [this analyzer](https://github.com/Brightspace/D2L.CodeStyle/blob/f0e21e18207e66167d6760f582deaafaace289d9/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs#L33-L34) https://github.com/Brightspace/D2L.CodeStyle/blob/f0e21e18207e66167d6760f582deaafaace289d9/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs#L33-L34

and could/will shrink that. One thing I'm waiting on is upgrading to C# 7.2+ where named arguments can go in the middle of unnamed arguments (that will allow the suggested fix to be more targeted for things like literals or swappable arguments.

If we were to use this to detect bad override/interface-impl parameter names (e.g. `IFoo.Bar( long orgId, long userId` being implemented by `Bar( long userId, long orgId )`) we wouldn't have an equivalent for named parameters to bound our search... but there are also far fewer function decls to check vs. all function/constructor calls.

## possible redundancy for function/constructor calls?
We're also considering forcing naming for all swappable arguments. The impl looks essentially like:

```cs
bool CanSwapArgs( ArgBinding b1, ArgBinding b2 ) {
  bool canMoveLeft = model.ClassifyConversion(
    b2.Arg.Expression,
    b1.Param.Type
  ).IsImplicit;

  bool canMoveRight = model.ClassifyConversion(
    b1.Arg.Expression,
    b2.Param.Type
  ).IsImplicit;

  return canMoveLeft && canMoveRight;
}
```

(untested but I've got that locally. Also I wrote it a bit different to short-circuit.)

If we did succeed at enforcing this then there would be nothing worth running this algorithm on I think*? I can get numbers tomorrow on how big enforcing just that would be in terms of manual reviewer labour.

* if you account for your algorithm not considering unswappable thingstypes. I don't know if you implemented that, I haven't looked at the code.

## heuristics/the paper

Yeah, reading the paper is a good idea. I don't think we should implement heuristics without checking if they'd be useful on our code base first though (best way to check that is to implement it but don't tell anyone ;) ).

zommerfelds:
Yeah it sounds like @j3parker's suggested rule makes this one useless. It is essentially a much stricter rule. This PR tries to enforce a more loose rule (if you are doing what the paper is doing, haven't analyzed all the code).

PR #336, opened 2018-04-26
Add a code fix that uses named arguments

j3parker:


src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -9,12 +9,29 @@
...
+			public ArgumentSyntax Syntax { get; }
+		}
+
 		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
 			Diagnostics.UseNamedArgsForInvocationWithLotsOfArgs
 		);
 
 		// TODO: shrink this number over time. Maybe 5 would be good?
-		public const int TOO_MANY_UNNAMED_ARGS = 64;
+		public const int TOO_MANY_UNNAMED_ARGS = 63;
```
j3parker: Accidentally made this bigger than necessary in original commit.

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -46,10 +63,18 @@ SyntaxNodeAnalysisContext ctx
...
+					keySelector: binding => binding.Position.ToString(),
+					elementSelector: binding => binding.ParamName
+				);
+
 				ctx.ReportDiagnostic(
 					Diagnostic.Create(
 						descriptor: Diagnostics.UseNamedArgsForInvocationWithLotsOfArgs,
-						location: expr.GetLocation()
+						location: expr.GetLocation(),
+						properties: fixerContext
```
omsmith: TIL

j3parker:
So VS had a built-in code fix for naming arguments but I couldn't find a way to use it.
Having our own code fix allows us to "do all in solution" and register it against the diagnostic.
Long term (after all the arguments we want to be named are name) it might be worth discarding this and letting people use the default one (which shows up under alt+enter)

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
j3parker: Not fixing in this PR because I'd have to update the tests and that's really unrelated.

j3parker:
image

First one is right, second one the arguments are missing their leading trivia. Need to figure that out.

src/D2L.CodeStyle.Analyzers/Language/UseNamedArgumentsCodeFix.cs
```diff
@@ -0,0 +1,112 @@
...
+								argNames,
+								ct
+							)
+					),
+					diagnostic
+				);
+			}
+		}
+
+		private static SyntaxToken CommaNewline( object _ ) {
```
omsmith: Unused?

src/D2L.CodeStyle.Analyzers/Language/UseNamedArgumentsCodeFix.cs
```diff
@@ -0,0 +1,112 @@
...
+				var arg = args.Arguments[idx];
+
+				// Some args might already be named
+				if ( !argNames.ContainsKey( idx ) ) {
+					yield return arg;
+					continue;
+				}
+
+				SyntaxTriviaList leadingTrivia =
+					arg.RefOrOutKeyword.Kind() == SyntaxKind.None
```
j3parker: This is what was getting me... I was checking for `null` not a none-token. Sigh.

j3parker:
@omsmith ok done

PR #337, opened 2018-04-27
Reduce limit on unnamed arguments to 31

j3parker:
This should be safe to merge soon.


j3parker:
Confirmed this limit is high enough after I get that last change in.

PR #340, opened 2018-04-30
Consider constructors in addition to function calls for named arguments

j3parker:
Forgot to handle this. Hopefully the fallout won't be too bad.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
@@ -1,5 +1,4 @@
-﻿using System;
-using System.Collections.Generic;
+﻿using System.Collections.Generic;
 using System.Collections.Immutable;
 using D2L.CodeStyle.Analyzers.Extensions;
 using Microsoft.CodeAnalysis;
--- a/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using D2L.CodeStyle.Analyzers.Extensions;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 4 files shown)


j3parker:
Fallout from missing this isn't too bad.

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -137,5 +144,20 @@ ArgumentListSyntax args
 				);
 			}
 		}
+
+		// Not an extension method because there may be more cases (e.g. in the
+		// future) and if more than this fix + its analyzer used this logic
+		// there could be undesirable coupling if we handled more cases.
+		internal static ArgumentListSyntax GetArgs( SyntaxNode syntax ) {
+			if ( syntax is InvocationExpressionSyntax ) {
```
omsmith: nit: consider switch with pattern matching?

src/D2L.CodeStyle.Analyzers/Language/UseNamedArgumentsCodeFix.cs
```diff
@@ -120,5 +114,21 @@ CancellationToken cancellationToken
...
 			}
 		}
+
+		public static ArgumentListSyntax GetArgs(
+			SyntaxNode root,
+			TextSpan span
+		) {
+			// FindNode() may identify a parent of the ExpressionSyntax
+			// we'd like to fix that has an equal span. We're grabbing the
+			// "first" child InvocationExpressionSyntax. I'm guessing that will
```
omsmith: nit: comment references invocationexpression

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -137,5 +144,20 @@ ArgumentListSyntax args
 				);
 			}
 		}
+
+		// Not an extension method because there may be more cases (e.g. in the
+		// future) and if more than this fix + its analyzer used this logic
+		// there could be undesirable coupling if we handled more cases.
+		internal static ArgumentListSyntax GetArgs( SyntaxNode syntax ) {
+			if ( syntax is InvocationExpressionSyntax invocation ) {
```
omsmith: That's cool, but I meant

```csharp
switch( syntax ) {
  case InvocationExpressionSyntax invocation:
    return invocation.ArgumentList;
  case ObjectCreationExpressionSyntax objectCreation:
    return objectCreation.ArgumentList;
  default:
    return null;
}
```

j3parker: Whoa cool tech

(I wish you could have sealed enum types so `default` wasn't necessary.)

j3parker:
That last commit is the right way to use `FindNode`. The thing I was doing with `.DescendantNodes` was wrong.

j3parker:
Passes in the LMS once some stuff gets merged. Will wait a couple of days after merging to deploy to maybe avoid aggro due to folks checkouts.

j3parker:
I think I've waited long enough for this. I don't think waiting really would help much but wanted to space out the inevitable split-checkout breakages that people have a bit.

PR #342, opened 2018-05-01
Require named args when the arg expression is a literal

j3parker:


src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -63,29 +64,52 @@ SyntaxNodeAnalysisContext ctx
...
+						Diagnostic.Create(
+							descriptor: Diagnostics.LiteralArgShouldBeNamed,
+							location: expr.GetLocation(),
+							properties: GetDiagnosticContext( unnamedArgs ),
+							messageArgs: binding.ParamName
+						)
+					);
+
+					// Don't need any more diagnostics for this argument
+					continue;
```
omsmith: Similar to the previous comment, since the diagnostic and codefix is for the entirety of the call, this should be a return?

j3parker: I guess that's true ya

j3parker:
This needs some work. We upgraded to C# 7.2+ so we can now name specific arguments without naming all of them (or doing the complicated thing.)

I'll have to revisit this when I get back.

PR #351, opened 2018-05-15
Assert.IsTrue/IsFalse analyzer

cbortos-d2l:


src/D2L.CodeStyle.TestAnalyzers/NUnit/AssertIsBool/AssertIsBoolBinaryExpressions.cs
```diff
@@ -0,0 +1,270 @@
...
+				expression.Kind() == SyntaxKind.NumericLiteralExpression && expression.ToString() == "0";
+
+			bool IsTrueLiteral( ExpressionSyntax expression ) =>
+				expression.Kind() == SyntaxKind.TrueLiteralExpression;
+
+			bool IsFalseLiteral( ExpressionSyntax expression ) =>
+				expression.Kind() == SyntaxKind.FalseLiteralExpression;
+
+			ExpressionSyntax otherOperand;
+			if( binaryExpression.TryGetSingleOperandForEqualityReplacement( IsNullLiteral, out otherOperand ) ) {
```
omsmith: It would be really great if each of these had a comment of what they're looking for, and what they would transform to, especially the final `else`

e.g. `// Assert.IsTrue( foo == null ) -> Assert.IsNull( foo )`

cbortos-d2l: Done

mthjones:
Seems Github added me to all the older PRs as a code owner, so since I'm here...

NUnit has a suite of analyzers now that will warn for this scenario (and [many more](https://github.com/nunit/nunit.analyzers/blob/master/documentation/index.md)!). Bringing that into the projects and configuring it as applicable may be worth considering as an alternative.

j3parker:
Whoa, cool!
I forgot about that PR :( 
Those analyzers look great.

j3parker:
Opps, I didn't mean to close this.

cbortos-d2l:
Given the NUnit analyzers that @mthjones mentioned and because this change is really old, I'm considering deleting the branch. Let me know if you have anything against me doing so.

PR #352, opened 2018-05-16
Framework and language upgrade

j3parker:


j3parker:
Note: D2L.CodeStyle.Annotations is .NET 2.0 for compatability, but I upgraded it to C# 7.3. I think that's OK.

PR #353, opened 2018-05-17
MapPath dangerous methods

cbortos-d2l:


src/D2L.CodeStyle.Analyzers/D2L.CodeStyle.Analyzers.csproj
```diff
@@ -150,6 +150,7 @@
       <HintPath>..\..\packages\System.Reflection.Metadata.1.2.0\lib\portable-net45+win8\System.Reflection.Metadata.dll</HintPath>
       <Private>True</Private>
     </Reference>
+    <Reference Include="System.Web" />
```
j3parker: Hm, could we go with strings rather than referencing this?

j3parker:
Sweet, cool stuff!

j3parker:
(One of the reasons to not reference `System.Web`: it would prevent us from shipping a netstandard/.NET core build... we don't currently but it's definitely something we'll want to do in the future..)

tests/D2L.CodeStyle.Analyzers.Test/SpecRunner.cs
```diff
@@ -350,6 +351,12 @@ ImmutableArray<string> arguments
 					MetadataReference
 						.CreateFromFile( typeof( Enumerable ).Assembly.Location ) )
 
+				// system.web
```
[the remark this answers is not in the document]
cbortos-d2l: The `spec` test code references classes from this assembly (i.e. `System.Web.Hosting.HostingEnvironment`). Without it, the symbols were `null`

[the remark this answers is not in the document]
cbortos-d2l: Oh, I see. I can do that.

src/D2L.CodeStyle.Analyzers/D2L.CodeStyle.Analyzers.csproj
j3parker: Yeah :(

tests/D2L.CodeStyle.Analyzers.Test/SpecRunner.cs
```diff
@@ -100,7 +100,7 @@ private sealed class DiagnosticComparer : IEqualityComparer<Diagnostic> {
 				.Intersect(
 					m_expectedDiagnostics,
 					DiagnosticComparer.Instance
-				).ToImmutableHashSet();
```
j3parker: D'oh, thanks!

j3parker:
@cbortos-d2l isn't sure if he'll merge this or when so this will sit open for a bit.

PR #354, opened 2018-05-24
Bump NUnit version

cbortos-d2l:
use the same nunit version as the LMS (at the very least, although newer versions exist)


cbortos-d2l:
Actual motivation for this is that I want to use new Assert methods in one of my other branches.

PR #356, opened 2018-06-20
Adding support for immutable type parameters

toddlang:
-A generic type parameter can now me marked [Immutable] so it can be held inside an immutable generic
-Usages of the generic will ensure the type passed in to the immutable generic parameter is itself immutable

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -441,48 +441,6 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			);
 		}
 
-		// Walks all ancestors to this type parameter to determine if it was
-		// intended that the type be immutable.  If *any* ancestor says the
-		// type was supposed to be immutable, then the type will be 
-		// assumed to mandatory immutable.
-		private bool InspectInterfaceTypeParameter(
-			int parameterIndex,
-			ITypeSymbol symbol
-		) {
-			if( symbol == default ) {
-				return false;
-			}
-
-			if( IsTypeParameterToBeImmutable( parameterIndex, symbol as INamedTypeSymbol ) ) {
-				return true;
-			}
-
-			foreach( var intf in symbol.Interfaces ) {
-				if( InspectInterfaceTypeParameter( parameterIndex, intf ) ) {
-					return true;
-				}
-			}
-
-			return false;
-		}
-
-		// Indicates if the type parameter at the specified slot is marked
-		// for immutability.
-		private bool IsTypeParameterToBeImmutable(
-			int parameterIndex,
-			INamedTypeSymbol symbol
-		) {
-			if( symbol.TypeParameters.Length <= parameterIndex ) {
-				// This type ends the chain of inspection since the generic
-				// type can't line up here
-				return false;
-			}
-			var parameter = symbol.TypeParameters[parameterIndex];
-			var scope = parameter.GetImmutabilityScope();
-
-			return scope != ImmutabilityScope.None;
-		}
-
 		private MutabilityInspectionResult InspectInitializer(
 			ExpressionSyntax expr,
 			HashSet<ITypeSymbol> typeStack
@@ -586,6 +544,61 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				MutabilityInspectionFlags.Default,
 				typeStack
 			).WithPrefixedMember( field.Name );
+		}
+
+		/// <summary>
+		/// Walks all ancestors to this type parameter to determine if it was
+		/// intended that the type be immutable.  If *any* ancestor says the
+		/// type was supposed to be immutable, then the type will be 
+		/// assumed to mandatory immutable.
+		/// </summary>
+		/// <param name="parameterIndex">The position of the parameter in the symbols TypeParameter array.</param>
+		/// <param name="symbol">The symbol containing the set of type parameters to be inspected.</param>
+		/// <returns>Returns <code>True</code> if the type parameter or any ancestor
+		/// type parameter in the specified <paramref name="parameterIndex"/> position is
+		/// marked for immutability.</returns>
+		private static bool InspectInterfaceTypeParameter(
+			int parameterIndex,
+			ITypeSymbol symbol
+		) {
+			if( symbol == default ) {
+				return false;
+			}
+
+			if( IsTypeParameterToBeImmutable( parameterIndex, symbol as INamedTypeSymbol ) ) {
+				return true;
+			}
+
+			foreach( var intf in symbol.Interfaces ) {
+				if( InspectInterfaceTypeParameter( parameterIndex, intf ) ) {
+					return true;
+				}
+			}
+
+			return false;
+		}
+
+		/// <summary>
+		/// Determines if the <code>TypeParameter</code> at the 
+		/// <code>parameterIndex</code> position is marked for immutability.
+		/// </summary>
+		/// <param name="parameterIndex">The index of the type parameter to be examined.</param>
+		/// <param name="symbol">The symbol containing the type parameters to be examined.</param>
+		/// <returns>Return <code>True</code> if the type parameter is marked for immutability,
+		/// otherwise returns <code>False</code>.</returns>
+		private static bool IsTypeParameterToBeImmutable(
+			int parameterIndex,
+			INamedTypeSymbol symbol
+		) {
+			if( symbol.TypeParameters.Length <= parameterIndex ) {
+				// This type ends the chain of inspection since the generic
+				// type can't line up here
+				return false;
+			}
+			var parameter = symbol.TypeParameters[parameterIndex];
+			var scope = parameter.GetImmutabilityScope();
+
+			return scope != ImmutabilityScope.None;
 		}
 
 		/// <summary>
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -31,6 +31,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		private static readonly ImmutableDictionary<string, string[]> ImmutableContainerTypes = new Dictionary<string, string[]> {
 			["D2L.LP.Utilities.DeferredInitializer"] = new[] { "Value" },
 			["D2L.LP.Extensibility.Activation.Domain.IPlugins"] = new[] { "[]" },
+			["System.Collections.Immutable.IImmutableSet"] = new[] { "[]" },
 			["System.Collections.Immutable.ImmutableArray"] = new[] { "[]" },
 			["System.Collections.Immutable.ImmutableDictionary"] = new[] { "[].Key", "[].Value" },
 			["System.Collections.Immutable.ImmutableHashSet"] = new[] { "[]" },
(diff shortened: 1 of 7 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -372,12 +414,75 @@ HashSet<ITypeSymbol> typeStack
...
+					return true;
+				}
+			}
+
+			return false;
+		}
+
+		// Indicates if the type parameter at the specified slot is marked
+		// for immutability.
+		private bool IsTypeParameterToBeImmutable(
```
JeffAshton: static?

tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutabilityAnalyzer.cs
```diff
@@ -315,4 +326,86 @@ sealed class ClassWithExceptedImmutableMemberSucceeds {
...
+		class ConcreteClassWithImmutableGenericType {
+			internal readonly GenericClassWithImmutableState<ImmutableClass> m_ConcreteClassWithImmutableGenericType;
+		}
+	}
+
+	class GenericTypeParameterTests {
+		// Ensures generic type parameters from an interface are examined
+
+		[Objects.Immutable]
+		interface ImmutableGenericInterface<[Objects.Immutable] T> {
```
ChrisAtD2L: We should check that we complain when the generic interface has an implementation that holds onto type `T`, but we should also check that if there is no implementation holding onto type `T` then everything still passes.  Fortunately, it looks like you are covered by some existing tests for that:
`GenericClassWithoutStateIsSafe<T>`

tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutabilityAnalyzer.cs
```diff
@@ -315,4 +326,86 @@ sealed class ClassWithExceptedImmutableMemberSucceeds {
...
+		[Objects.Immutable]
+		sealed class ConcreteUsingMultiGenericFromInterface {
+			internal readonly MultiGenericFromInterface<ImmutableClass, NonReadOnlyClass> m_ConcreteUsingMultiGenericFromInterface;
+		}
+
+		[Objects.Immutable]
+		sealed class /* ImmutableClassIsnt('m_ConcreteUsingMultiGenericFromInterfaceWrongOrder.m_MultiGenericFromInterface.m_MutableClass' is not read-only) */ ConcreteUsingMultiGenericFromInterfaceWrongOrder /**/ {
+			internal readonly MultiGenericFromInterface<MutableClass, ImmutableClass> m_ConcreteUsingMultiGenericFromInterfaceWrongOrder;
+		}
+	}
```
ChrisAtD2L: Other test ideas:
* What happens if you add `Objects.Immutable` to type parameter, when it is superfluous?  Ie, nothing actually holds on to that type.  Does it do nothing (since the implementations are always immutable), or complain if a mutable type is used? (Someone will make a refactoring mistake someday)
* Could we add a test where more than one type parameter is marked `Immutable`?  Probably works as intended, but I'd like the safety net.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -483,6 +556,61 @@ HashSet<ITypeSymbol> typeStack
...
+			if( symbol == default ) {
+				return false;
+			}
+
+			if( IsTypeParameterToBeImmutable( parameterIndex, symbol as INamedTypeSymbol ) ) {
+				return true;
+			}
+
+			foreach( var intf in symbol.Interfaces ) {
+				if( InspectInterfaceTypeParameter( parameterIndex, intf ) ) {
```
ChrisAtD2L: This looks like it is assuming that if an interface has 2 type params, then parent interfaces must have the same type params, in the same order.  That's not true.  This compiles:
```
public interface BaseFoo<T, U> {}
public interface SubFoo<U, T> : BaseFoo<T, U> {}
public interface SubsetFoo<T> : BaseFoo<T, T> {}
public interface SupersetFoo<S, T, U, V> : BaseFoo<U, V> {}
```
The type parameters on `symbol` might be passed to parent interfaces multiple times, in a different order, or not at all.  This code does not seem to account for that.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -483,6 +556,61 @@ HashSet<ITypeSymbol> typeStack
...
+		) {
+			if( symbol.TypeParameters.Length <= parameterIndex ) {
+				// This type ends the chain of inspection since the generic
+				// type can't line up here
+				return false;
+			}
+			var parameter = symbol.TypeParameters[parameterIndex];
+			var scope = parameter.GetImmutabilityScope();
+
+			return scope != ImmutabilityScope.None;
```
ChrisAtD2L: Should this be `== ImmutabilityScope.SelfAndChildren`?
Right now, it looks like we allow either `Immutable` or `ImmutableBaseClass`, but I don't think that the base class one makes sense to apply to type parameters.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -349,6 +372,35 @@ HashSet<ITypeSymbol> typeStack
 			return MutabilityInspectionResult.NotMutable( unauditedReasonsBuilder.ToImmutable() );
 		}
 
+		private MutabilityInspectionResult InspectInterface(
```
ChrisAtD2L: I'm a little bit confused.  I can see that this new code applies to type parameters on interfaces (it is called from the `TypeKind.Interface` switch block).  But I don't see how we end up examining the type parameters for inheritance / abstract classes.  Is it possible to get an error of `MutabilityCause.IsMutableTypeParameter` when inheriting from a base class rather than implementing an interface?

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -427,17 +431,34 @@ HashSet<ITypeSymbol> typeStack
...
+
+					// If it doesn't have a matching type argument then the
+					// type we're inspecting isn't defined in this interface.
+					if ( ordinal < 0) {
+						continue;
+					}
+
+					// Grab the implemented interfaces version of the type 
+					// argument and then pass it down for inspection
+					var subTypeParameter = intf.TypeArguments[ordinal] as ITypeParameterSymbol;
```
ChrisAtD2L: nit: It feels wrong to find the typeParameter (or default), then throw it away to get the ordinal, then use the ordinal to get the original typeArgument back (`intf.TypeArguments[ordinal]`).  It makes more intuitive sense to me to do something like:
```
var matchingArgument = intf.TypeArguments.FirstOrDefault( ... );
if ( matchingArgument == default(TypeArgument) ) { continue; }
var ordinal = intf.TypeArguments.IndexOf( matchingArgument );
```
but that is a really tiny nitpick and it probably won't make any difference in the long run

tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutabilityAnalyzer.cs
```diff
@@ -466,4 +466,28 @@ sealed class /* ImmutableClassIsnt('m_ImmutableBase.m_MutableClass' is not read-
...
+		}
+
+		interface IMiddle<T> : IImmutableRoot<T> {
+		}
+
+		interface IMutableRoot<T> {
+		}
+
+		[Objects.ImmutableBaseClass]
+		class MixinBase<U, V> : IMutableRoot<U>, IMiddle<V> {
```
ChrisAtD2L: I really like this test case

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -372,6 +428,44 @@ HashSet<ITypeSymbol> typeStack
...
+			var currentType = typeParameter.ContainingType;
+			while( currentType != null ) {
+				// Check all interfaces
+				foreach( var intf in currentType.Interfaces ) {
+					// Find the matching type argument in the base interface
+					// that corresponds to this type parameter.  That is,
+					// if we have Foo<S, T>: IFoo<S>, IBar<T>, this will
+					// match up the Foo S, to the IFoo S, but will get -1
+					// from IBar since it doesn't have S.
+					var ordinal = intf.TypeArguments
```
JeffAshton: Found this confusing.  Just break into 2 lines.  

```
var typeArgument = intf.TypeArguments.FirstOrDefault( tp => tp.Name == typeParameter.Name );
var ordinal = intf.TypeArguments.IndexOf( typeArgument );
```

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -142,9 +143,19 @@ HashSet<ITypeSymbol> typeStack
 					+ "are referenced, including transitive dependencies." );
 			}
 
+			// We can't short-circuit below if this is a generic type because
+			// while the type can be marked immutable we can't be certain
+			// all type parameters have been marked immutable at some point
+			// in their heirarchy.
+			bool isGenericType = false;
```
JeffAshton: Feels like a good use for a helper method, since this code is duplicated below.

JeffAshton: I wouldn't be surprised if Jacob has an area for extension methods like this

JeffAshton: I feel like then you can just merge your new comment with the exist comment.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -349,6 +373,38 @@ HashSet<ITypeSymbol> typeStack
...
+			var scope = typeSymbol.GetImmutabilityScope();
+
+			// There is a 1:1 correlation between TypeParameters and TypeArguments.
+			// TypeParameters is the "S", or "T" definition.
+			// TypeArguments are the actual *types* passed to S or T.
+			for (int ordinal = 0; ordinal < typeSymbol.TypeParameters.Length; ordinal++) {
+				var parameterType = typeSymbol.TypeArguments[ordinal];
+				var isToBeImmutable = IsTypeArgumentImmutable( typeSymbol.TypeParameters[ordinal], ordinal, typeSymbol );
+				var result = InspectType( parameterType );
+				if( result.IsMutable && isToBeImmutable ) {
```
JeffAshton: micro optimization if you break up this line, but it might make your branches more clear and easier to verify you've tested

```
if( !result.IsMutable ) {
   continue;
}

if( isToBeImmutable ) {
   return MutabilityInspectionResult.MutableType( parameterType, MutabilityCause.IsMutableTypeParameter );
}
```

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -483,6 +577,44 @@ HashSet<ITypeSymbol> typeStack
...
+			if( symbol == default ) {
+				return false;
+			}
+
+			var isMarkedImmutable = symbol.TypeParameters[parameterOrdinal].GetImmutabilityScope() != ImmutabilityScope.None;
+			if( isMarkedImmutable ) {
+				return true;
+			}
+
+			foreach( var intf in symbol.Interfaces ) {
```
JeffAshton: Looks to be a duplicate of the code above.

JeffAshton: Pull this into a helper method, and leverage in both places

PR #358, opened 2018-06-25
Adding support for immutable type parameters

toddlang:
-A generic type parameter can now me marked [Immutable] so it can be held inside an immutable generic
-Usages of the generic will ensure the type passed in to the immutable generic parameter is itself immutable
-Adding tests for multiple immutable generic parameters.
-Adding tests for superfluous parameter annotation.
-Adding tests for mixin parameters and ordering.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -1,11 +1,11 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
 using System.Collections.Immutable;
+using D2L.CodeStyle.Analyzers.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
-using System;
-using System.Collections.Concurrent;
-using D2L.CodeStyle.Analyzers.Extensions;
 
 namespace D2L.CodeStyle.Analyzers.Immutability {
 	[Flags]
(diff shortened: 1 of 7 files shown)


toddlang:
Sorry guys, couldn't see how to rebase or just re-open, so squashed and repushed.  :/

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -350,18 +366,60 @@ HashSet<ITypeSymbol> typeStack
 			return MutabilityInspectionResult.NotMutable( unauditedReasonsBuilder.ToImmutable() );
 		}
 
+		private MutabilityInspectionResult InspectInterface(
+			ITypeSymbol symbol,
+			HashSet<ITypeSymbol> typeStack
+		) {
+			var typeSymbol = symbol as INamedTypeSymbol;
```
JeffAshton: Throw meaningful message when this isn't the case, instead of null reference exception

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -373,6 +431,37 @@ HashSet<ITypeSymbol> typeStack
...
+			// the immutability attribute is present.
+			INamedTypeSymbol currentType = typeParameter.ContainingType;
+			while( currentType != null ) {
+				// Check all interfaces
+				foreach( INamedTypeSymbol intf in currentType.Interfaces ) {
+
+					// Find out if this interface exposes a type argument
+					// with the specified type name.  If it doesn't, this 
+					// interface doesn't contribute to the immutability chain.
+					var ordinal = intf.IndexOfArgument( typeParameter.Name );
```
JeffAshton: int?

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -484,6 +573,56 @@ HashSet<ITypeSymbol> typeStack
...
+
+			if (typeParameter == default) {
+				return false;
+			}
+
+			// We needt the TypeParameter here, otherwise we're inspecting
+			// the type and not the declaration.  We need to inspect the 
+			// declaration because that's the symbol that will have the
+			// [Immutable] attached to it.
+			var isMarkedImmutable = symbol.TypeParameters[parameterOrdinal]
```
JeffAshton: bool?

toddlang:
For the sake of those who read this in the future:

The IsPotentiallyMutable reason is used so that if a type is analyzed but the analysis is unable to be completed due to being unable to resolve the symbol (if an AS coercion resolved to null) then we say the type is mutable with IsPotentiallyMutable as the cause.

This then ends up in MutabilityInspectionResultFormatter, and when the diagnostic is printed, it will say "{typename} is not deterministically immutable."   Basically, we're trying to say: If the type was supposed to be immutable, but we couldn't finish the analysis to be sure it's immutable, we'll return that it IS mutable with a reason of "We couldn't be sure it's immutable", and the compilation will fail.

PR #361, opened 2018-06-29
Refactoring immutable generic parameters

toddlang:
-Adding tests that reflect how the code is structured in the actual product

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -52,8 +52,8 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		private readonly KnownImmutableTypes m_knownImmutableTypes;
 		private readonly Compilation m_compilation;
 
-		private readonly ConcurrentDictionary<Tuple<ITypeSymbol, MutabilityInspectionFlags>, MutabilityInspectionResult> m_cache
-			= new ConcurrentDictionary<Tuple<ITypeSymbol, MutabilityInspectionFlags>, MutabilityInspectionResult>();
+		private readonly ConcurrentDictionary<(ITypeSymbol Symbol, MutabilityInspectionFlags Flags), MutabilityInspectionResult> m_cache
+			= new ConcurrentDictionary<(ITypeSymbol Symbol, MutabilityInspectionFlags Flags), MutabilityInspectionResult>();
 
 		internal MutabilityInspector(
 			Compilation compilation,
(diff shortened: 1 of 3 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -52,8 +52,8 @@ internal sealed class MutabilityInspector {
 		private readonly KnownImmutableTypes m_knownImmutableTypes;
 		private readonly Compilation m_compilation;
 
-		private readonly ConcurrentDictionary<Tuple<ITypeSymbol, MutabilityInspectionFlags>, MutabilityInspectionResult> m_cache
-			= new ConcurrentDictionary<Tuple<ITypeSymbol, MutabilityInspectionFlags>, MutabilityInspectionResult>();
+		private readonly ConcurrentDictionary<(ITypeSymbol Symbol, MutabilityInspectionFlags Flags), MutabilityInspectionResult> m_cache
```
toddlang: I got type if "query.Item1, query.Item2" and replaced it with a named tuple.

toddlang:
Overall, the biggest change is to start passing the "host" symbol with a type parameter.

ie: 

class Foo<[Immutable] T> {
   private readonly T m_field;
}

When we inspected m_field, we would see "ITypeParameterSymbol" named "T".  But that's all we would have.  We we will also pass Foo<[Immutable] T> with the type parameter so we can get container information about the type parameter.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
ChrisAtD2L: This is so much better than the old way of working with tuples.  Thanks!

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -77,22 +77,23 @@ KnownImmutableTypes knownImmutableTypes
...
+				typesInCurrentCycle,
+				type
 			);
 		}
 
 		public MutabilityInspectionResult InspectField( IFieldSymbol field ) {
 			var typesInCurrentCycle = new HashSet<ITypeSymbol>();
 
-			return InspectField( field, typesInCurrentCycle );
+			return InspectField( field, typesInCurrentCycle, default );
```
JeffAshton: what's ```default```?  As in, without looking at the method signature, I have no idea what you're passing ```null```/```false```/0 for

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
JeffAshton: This type of thing is great as it's own small commit

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
JeffAshton: Actually, instead of passing ```default``` everywhere, maybe just an an overload that doesn't take it.  Better then a default value, is it makes it easier to find cases where we pass it vs we don't pass it

PR #364, opened 2018-07-03
Moving a couple methods to be static extension methods

toddlang:


src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -18,6 +18,45 @@ internal static partial class RoslynExtensions {
...
+		}.ToImmutableDictionary();
+
+		public static bool IsAnImmutableContainerType( this ITypeSymbol type ) {
+			return ImmutableContainerTypes.ContainsKey( type.GetFullTypeName() );
+		}
+
+		public static string[] GetImmutableContainerTypePrefixes( this ITypeSymbol type ) {
+			string fullTypeName = type.GetFullTypeName();
+
+			if( ImmutableContainerTypes.ContainsKey( fullTypeName ) ) {
```
JeffAshton: TryGetValue

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -18,6 +18,45 @@ internal static partial class RoslynExtensions {
 			"System.IO.Abstractions.IFileSystem"
 		);
 
+		/// <summary>
```
JeffAshton: These might be better in their own extension methods class.  Everything to do with immutable containers.

PR #373, opened 2018-07-04
Adding missing hostSymbol parameter

toddlang:
-Updating interface inspection to consider the host symbol


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -224,7 +225,12 @@ HashSet<ITypeSymbol> typeStack
 				}
 
 				ITypeSymbol parameterType = typeSymbol.TypeArguments[ordinal];
-				MutabilityInspectionResult result = InspectType( parameterType );
+				MutabilityInspectionResult result = DoInspectType(
```
JeffAshton: FYI,  Old implementation:

MutabilityInspectionFlags flags = MutabilityInspectionFlags.Default

JeffAshton: Why are we now cool with AllowUnsealed ?

PR #374, opened 2018-07-04
Extending type parameter inspection

toddlang:
-Now properly considers type constraints on type parameters

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -255,6 +255,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			HashSet<ITypeSymbol> typeStack
 		) {
 			var typeParameter = symbol as ITypeParameterSymbol;
+			var hostType = hostSymbol as INamedTypeSymbol;
 
 			// If we can't determine what the symbol is then we'll bail with
 			// a general mutability response, otherwise we're going to have a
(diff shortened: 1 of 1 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -283,10 +284,46 @@ HashSet<ITypeSymbol> typeStack
...
+				if( parameterScope == ImmutabilityScope.SelfAndChildren ) {
+					return MutabilityInspectionResult.NotMutable();
+				}
+			}
+
 			// We have to walk all base types and interfaces to find the
 			// type param with the same name and examine those to see if
 			// the immutability attribute is present.
-			INamedTypeSymbol currentType = typeParameter.ContainingType;
+			INamedTypeSymbol currentType = hostType;
```
toddlang: This was the wrong line.  The containing type is the ultimate ancestor.  Now that we have hostSymbol we can properly only look at the class that contains the type parameter we're inspecting.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -283,10 +284,46 @@ HashSet<ITypeSymbol> typeStack
 				}
 			}
 
+			// Inspect the type parameter as part of its host symbol
```
JeffAshton: It feels like these 2 blocks are methods on their own.  They are asking whether the generic type parameter is immutable.  

var hostType = hostSymbol as INamedTypeSymbol;
if( hostType != default ) {
  
     if( CheckHostTypeArguments => NotImmutable ) {
        return NotImmutable
     }

    if( CheckHostTypeInterfaceArguments => NotImmutable ) {
        return NotImmutable
     }
}

PR #377, opened 2018-07-09
Adding immutability checking for immutable type arguments

toddlang:
-Foo<[Immutable> T> is now enforced for immutable classes

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -223,6 +223,33 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		) {
 			var typeParameter = symbol as ITypeParameterSymbol;
 
+			var hostType = hostSymbol as INamedTypeSymbol;
+
+			if (hostType == default) {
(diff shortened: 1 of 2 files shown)


toddlang:
This is the first pass where you can now do:

```
[Immutable]
Foo<[Immutable] T>{
   private readonly T m_T;
}
```

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -223,6 +223,33 @@ HashSet<ITypeSymbol> typeStack
 		) {
 			var typeParameter = symbol as ITypeParameterSymbol;
 
+			var hostType = hostSymbol as INamedTypeSymbol;
```
JeffAshton: This new code is a complete thought, and unrelated to the code below it in the same method.  Consider breaking it into it's own helper method.

toddlang: Ah, I was viewing it that the method didn't need the hostSymbol coerced in to a INamedTypeSymbol until I had added the logic in the routine below.  Otherwise it was kinda pointless.  Like, up until the logic is in, it's okay if the hostSymbol *isn't* a named type symbol.

PR #380, opened 2018-07-11
Adding inspection of generic arguments on concrete types

toddlang:


src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -549,6 +555,75 @@ HashSet<ITypeSymbol> typeStack
...
+					return MutabilityInspectionResult.MutableType( type, MutabilityCause.IsPotentiallyMutable );
+				}
+
+				int index = 0;
+				foreach( ITypeSymbol argument in symbolType.TypeArguments ) {
+					ImmutabilityScope argumentScope =
+						symbolType.TypeParameters[index].GetImmutabilityScope();
+
+					if( argumentScope == ImmutabilityScope.SelfAndChildren ) {
+						MutabilityInspectionResult result = DoInspectType(
```
toddlang: I would have moved these all out in to the ImmutableGenericArgumentClass, but this needs to be able to call DoInspectType and have the type stack available.

tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutabilityAnalyzer.cs
```diff
@@ -363,4 +363,79 @@ public class MutableClassImmutableState<[Objects.Immutable] T> {
 			private readonly T m_field;
 		}
 	}
+
+	class ConcreteGenericTests {
```
JeffAshton: So I was expecting to see some tests like:

```
		[Objects.Immutable]
		interface ImmutableInterface<[Objects.Immutable] T> {
		}

                class SomeClass {
                    static void Method() {
                          ImmutableInterface<MutableClass> blah = default;
                    } 
                }
```

Which shouldn't compile.  We need to every ```GenericNameSyntax```

JeffAshton: image

JeffAshton: image

toddlang: It hasn't historically been the case that the analyzer concerns itself with things used in a mutable context.  So, if something said Foo<[Immutable]T>, but passed a mutable type to it, as long as Foo wasn't being held in an immutable context, we couldn't actually stop it from happening.

In order to be able to catch that, we'll have to open up the scanner to whole of additional scan entry points.  I'll see what i can do.

toddlang: Given that being able to detect mutable classes passed to immutable type constraints in a mutable context is new functionality, does this commit suffice and I can start a new PR that builds on this to add that functionality?

PR #385, opened 2018-07-23
Upgrading analyzers to be able to analyze C# 7.3 code

toddlang:


toddlang:
Watch for changes in SpecRunner.cs

PR #387, opened 2018-07-23
Fixing the spec test code for DependencyRegistrationsAnalyzer.

toddlang:


tests/D2L.CodeStyle.Analyzers.Test/Specs/DependencyRegistrationsAnalyzer.cs
```diff
@@ -25,6 +25,9 @@ public enum ObjectScope {
 	public abstract class ExtensionPointDescriptor {
 		public abstract string Name { get; }
 	}
+
+	[Singleton]
+	public interface IMarkedSingleton { }
```
ChrisAtD2L: Can you delete this instead?  It looks like Saquib replaced all the references to `IMarkedSingleton` with `ISingleton` in this commit:
https://github.com/Brightspace/D2L.CodeStyle/commit/46d0599443752a888c0e1b9ac379bc3111c19634#diff-28d8752f46afac5feb152b15935f80b9
That change made sense.  If there were any references left behind, updating to ISingleton makes more sense than bringing the Marked/Unmarked stuff back

tests/D2L.CodeStyle.Analyzers.Test/Specs/DependencyRegistrationsAnalyzer.cs
```diff
@@ -134,7 +153,7 @@ public sealed class SomeTestCases {
 			reg.ConfigureInstancePlugins<ImmutableThing>( ObjectScope.Singleton );
 			reg.ConfigureInstancePlugins<ImmutableThing, DefaultExtensionPoint<ImmutableThing>>( ObjectScope.Singleton );
 			reg.RegisterPluginExtensionPoint<DefaultExtensionPoint<ImmutableThing>, ImmutableThing>( ObjectScope.Singleton );
-			reg.RegisterPlugin<DefaultExtensionPoint<ImmutableThing>, IMarkedSingleton, ImmutableThing>( ObjectScope.Singleton );
+			reg.RegisterPlugin<DefaultExtensionPoint<IMarkedSingleton>, IMarkedSingleton, ImmutableThing>( ObjectScope.Singleton );
```
ChrisAtD2L: IMarkedSingleton -> ISingleton

tests/D2L.CodeStyle.Analyzers.Test/Specs/DependencyRegistrationsAnalyzer.cs
```diff
@@ -229,10 +241,16 @@ public interface IImmutableSingleton : ISingleton { }
 	public interface IImmutableSubSingleton : IImmutableSingleton { }
 
 	[Objects.Immutable]
-	public sealed class ImmutableThing : ISingleton, INotSingleton { }
+	public sealed class ImmutableThing : IMarkedSingleton, ISingleton, INotSingleton { }
```
ChrisAtD2L: Can you undo this change?  `ISingleton` and `IMarkedSingleton` are both empty interfaces with the `[Singleton]` attribute.  Seems like we only need one of them

tests/D2L.CodeStyle.Analyzers.Test/Specs/DependencyRegistrationsAnalyzer.cs
```diff
@@ -200,22 +219,15 @@ public struct RegistrationCallsInThisStructAreIgnored {
 			}
 		}
 
-		// Registrations in extension methods with enough information should be handled. 
-		public static class SomeExtensionMethods {
```
ChrisAtD2L: Why is this deleted?  It looks like a requirement that is not duplicated elsewhere.  Can this come back?

toddlang: Discussed offline, we'll put in an extension method test in a separate PR.

ChrisAtD2L: From an offline conversation, it sounds like this was broken already, so it wasn't testing what it claimed to test.  Removing for now, as it might take a lot of work to figure out how to fix it (especially if it may never have worked at all)

PR #396, opened 2018-07-31
Adding attributes for dangerous property usages

JeffAshton:


src/D2L.CodeStyle.Annotations/DangerousMethodUsage/AuditedAttribute.cs
```diff
@@ -8,7 +8,7 @@ public static partial class DangerousMethodUsage {
 		/// Indicates usages of a dangerous method have been declared as safe
 		/// </summary>
 		[AttributeUsage(
-			validOn: AttributeTargets.Method | AttributeTargets.Constructor,
```
JeffAshton: Was missing property definitions, which are sugar methods

cbortos-d2l:
Can you add some tests?

PR #402, opened 2018-08-15
Adding support for enums as generic type arguments

toddlang:


src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs
```diff
@@ -89,6 +89,10 @@ internal KnownImmutableTypes( IAssemblySymbol a )
 				return true;
 			}
 
+			if (type.TypeKind == TypeKind.Enum) {
```
ChrisAtD2L: It looks like someone already tried to cover enums and many types of primitives using the `ImmutableSpecialTypes` array.  In that array there is the following entry: `SpecialType.System_Enum,` (line 56)

Why does enum need a second check (the new one added on this line), but the other special types do not?  I'm confused, because it seems like enums are already allowable in regular static variables, so I am surprised that a change is needed to get them to work with `ImmutableGeneric`

ChrisAtD2L: Alternatively, if the special types thing is broken, should we remove the `enum` check from that code, and rely on this new check instead?  It feels strange to check twice

toddlang: So, the "SpecialTypes" array that includes enum and such is something that checks the 'type.SpecialType' property.  In the case of an enum as the type parameter on a generic, it has no special type.  In fact, the 'UnderlyingType is an int32'.

So, this covers the situation of a type parameter that is an enum being examined.

ChrisAtD2L: It *looks like* the SpecialTypes array is supposed to catch enums and primitives (long, int) that we know to be safe.  But it doesn't actually catch enums -- there is a separate check for `type.TypeKind == TypeKind.Enum` in the MutabilityInspector, so it likely never worked there either.

Todd is going to do a little bit of testing to see what the SpecialTypes block _does_ catch, and probably remove the Enum check from MutabilityInspector in favour of putting it here instead.

saquibmian: Just saw this while browsing around...

`type.SpecialType` is only valid if the type is a _special_ type; it's not the type's _kind_. There are a handful of special types in the framework, including all primitives, `System.Object`, `System.Delegate`, and `System.Enum` (of which all enums are instances, which is what I suspect this check was handling). 

To illustrate:

1. `typeof(BindingFlags)` has a Type of `BindingFlags`, a TypeKind of `Enum`, but no SpecialType.
2. `typeof(BindingFlags).BaseType` has a Type of `System.Enum`, a TypeKind of `Enum`, _and a SpecialType of `Enum`_.

So just checking against SpecialType handles the case where we're going up the class hierarchy to make sure base classes are immutable (`BindingFlags` -> `System.Enum` -> `System.ValueType` -> `System.Object`); but as you've discovered, it doesn't handle the original type itself (hence a check in the mutability inspector for TypeKind). Because both of them have a same TypeKind, just checking for that instead is okay. I agree it's kind of messy, so it could be cleaned up. 

:)

PR #404, opened 2018-08-16
Adding support for extracting the type from Nullable<T>

toddlang:
-Restoring System_Enum as known immutable type because it still is.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs
@@ -53,6 +53,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		/// A list of known immutable special types.
 		/// </summary
 		private readonly static ImmutableArray<SpecialType> ImmutableSpecialTypes = ImmutableArray.Create(
+			SpecialType.System_Enum,
 			SpecialType.System_Boolean,
 			SpecialType.System_Char,
 			SpecialType.System_SByte,
@@ -70,7 +71,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			SpecialType.System_IntPtr,
 			SpecialType.System_UIntPtr
 		);
-		
+
 		/// <summary>
 		/// A list of known immutable types defined for the assembly.
 		/// </summary>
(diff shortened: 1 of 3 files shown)


tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutableGenericDeclarationAnalyzer.cs
```diff
@@ -238,36 +238,40 @@ class GenericMethodClass {
 			}
 		}
 
-		public class ImmutableGenericEnum {
```
toddlang: The public modifier wasn't necessary and was inconsistent with the rest of the file, so I removed it while I was here.

src/D2L.CodeStyle.Analyzers/Language/ImmutableGenericDeclarationAnalyzer.cs
```diff
@@ -64,6 +67,15 @@ KnownImmutableTypes knownImmutableTypes
 					continue;
 				}
 
+				var typeName = typeSymbol.GetFullTypeName();
+				if( typeName == "System.Nullable" ) {
```
toddlang: We do it this way in the KnownImmutableTypes, and it's the only way I could find that would consistently identify when System.Nullable was in use.  (Strangely, GetTypeFromMetadataName was *not* working)

src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs
```diff
@@ -53,6 +53,7 @@ internal sealed class KnownImmutableTypes {
 		/// A list of known immutable special types.
 		/// </summary
 		private readonly static ImmutableArray<SpecialType> ImmutableSpecialTypes = ImmutableArray.Create(
+			SpecialType.System_Enum,
```
toddlang: Honestly, we shouldn't have removed that in the last change.  It's still a valid SpecialType that we can rule out, it just doesn't catch regular enums.  So, I'm putting it back in.

src/D2L.CodeStyle.Analyzers/Language/ImmutableGenericDeclarationAnalyzer.cs
toddlang: (To be clear, I mean matching types by name.)

PR #406, opened 2018-08-17
Adding support for container types as part of immutable generics.

toddlang:
-Removing some superfluous spec tests

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs
@@ -111,8 +111,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 
 			foreach( var attribute in typesAttributes ) {
 				var typeofArgument = attribute.ConstructorArguments[ 0 ];
-				var value = typeofArgument.Value as INamedTypeSymbol;
-				if( value == null ) {
+				if( !( typeofArgument.Value is INamedTypeSymbol value ) ) {
 					// unable to extract the type, continue safely
 					continue;
 				}
--- a/src/D2L.CodeStyle.Analyzers/Language/ImmutableGenericDeclarationAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/ImmutableGenericDeclarationAnalyzer.cs
@@ -27,8 +27,8 @@ namespace D2L.CodeStyle.Analyzers.Language {
 			var knownImmutableTypes = new KnownImmutableTypes( context.Compilation.Assembly );
 
 			context.RegisterSyntaxNodeAction(
-				ctx => AnalyzeNode( 
-					ctx, 
+				ctx => AnalyzeNode(
+					ctx,
 					knownImmutableTypes ),
 				SyntaxKind.GenericName );
 		}
@@ -42,7 +42,7 @@ namespace D2L.CodeStyle.Analyzers.Language {
 			// Attributes are not allowed on local function parameters so we
 			// have to ignore this node, otherwise we'll tell people to
 			// annotate a declaration that is forbidden.
-			if( syntaxNode.Ancestors().Any( a => a is LocalFunctionStatementSyntax )) {
+			if( syntaxNode.Ancestors().Any( a => a is LocalFunctionStatementSyntax ) ) {
 				return;
 			}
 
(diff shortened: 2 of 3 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs
```diff
@@ -111,8 +111,7 @@ internal KnownImmutableTypes( IAssemblySymbol a )
 
 			foreach( var attribute in typesAttributes ) {
 				var typeofArgument = attribute.ConstructorArguments[ 0 ];
-				var value = typeofArgument.Value as INamedTypeSymbol;
-				if( value == null ) {
+				if( !( typeofArgument.Value is INamedTypeSymbol value ) ) {
```
toddlang: Some minor optimization suggested by Visual Studio.

ChrisAtD2L: That is a more natural way to read the same idea.  Cool.

src/D2L.CodeStyle.Analyzers/Language/ImmutableGenericDeclarationAnalyzer.cs
```diff
@@ -67,21 +67,35 @@ KnownImmutableTypes knownImmutableTypes
...
+		private static void ValidateImmutability(
+			SyntaxNodeAnalysisContext context,
+			ITypeSymbol typeSymbol,
+			KnownImmutableTypes knownImmutableTypes
+		) {
+			if( ImmutableContainerMethods.IsAnImmutableContainerType( typeSymbol ) ) {
+
+				var namedTypeSymbol = typeSymbol as INamedTypeSymbol;
+				if( namedTypeSymbol == default( INamedTypeSymbol ) ) {
+					context.ReportDiagnostic( Diagnostic.Create(
```
ChrisAtD2L: I'm trying to understand this part.  At this point, we are passed a type (which can only be a generic type argument today, but this method shouldn't rely on that).  The type is something like `IEnumerable` that we know to be safe, but we also can't generate an `INamedTypeSymbol`, so we can't analyze type arguments recursively.  Is this possible?  Is it only possible for code that does not compile, yet specifies a fully qualified type name for a System object?

Hard to know whether raising a "Not Immutable" diagnostic makes sense, because I don't know how you could hit this line.

toddlang: It'll force the compile to fail on a type that *should* have been possible to convert to a INamedTypeSymbol, but couldn't for whatever reason.  This is intended to be a fail-safe condition - that is, we will fail the compile rather than letting something go in that we couldn't reason about.

Normally I would throw an exception here, but that would abort all future processing of the analysis, and I still want all the other diagnostics too.

src/D2L.CodeStyle.Analyzers/Language/ImmutableGenericDeclarationAnalyzer.cs
```diff
@@ -67,21 +67,35 @@ KnownImmutableTypes knownImmutableTypes
...
 
-				ImmutabilityScope argumentScope = typeSymbol.GetImmutabilityScope();
-				if( argumentScope != ImmutabilityScope.SelfAndChildren ) {
+				foreach( var typeArgument in namedTypeSymbol.TypeArguments ) {
+
+					ValidateImmutability( context, typeArgument, knownImmutableTypes );
+				}
+			} else {
+				if( ( !knownImmutableTypes.IsTypeKnownImmutable( typeSymbol ) ) &&
+					( typeSymbol.GetImmutabilityScope() != ImmutabilityScope.SelfAndChildren ) ) {
```
ChrisAtD2L: This looks like it will work, but it sucks that the standard way to say "is this random type immutable" is:
`typeSymbol.GetImmutabilityScope() != ImmutabilityScope.SelfAndChildren`
Maybe we can make that clearer later on.  It's not super clear as it is

src/D2L.CodeStyle.Analyzers/Language/ImmutableGenericDeclarationAnalyzer.cs
ChrisAtD2L: Ok, makes sense that if we ever hit this condition, we're forced to stop and look at what went wrong.

PR #411, opened 2018-08-31
Require that [Immutable] be applied transitively

j3parker:
Right now I'm trying to fix a bug in analysis but over time some logic has gotten moved around to places it shouldn't be. The result is that the fix I made for the other bug is hard to get in now. Untangling things is getting harder because the analysis has gotten really complex in my opinion.

This very simple analyzer will allow me to remove some recursion from the analyzer and help me kill some helper methods I think have grown out of control (and lost their purpose.)

From what I saw a while ago we won't need too many additional `[Immutable]` and I talked to multiple people yesterday who agreed it was probably better to have those annotations anyway.

Note that the mutability analysis is still going to be recursive. E.g. the intent is for this to always be analyzed as cool:

```cs
[Immutable]
public class BaseClassThatPeopleCanTrustToBeImmutable {
   // ...

   public SealedClassThatHappensToBeImmutable Foo { get; }
}

public sealed class SealedClassThatHappensToBeImmutable {
  // ...
}
```

Before I merge this I'll push a code-fix for adding `[Immutable]` with Alt+Enter. I've written it I just have to try it out to make sure the whitespace is good. I'll make sure the LMS passes this analyzer before merging as well. Getting this up "early" in case there are any strong feelings.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ExplicitImmutableAttributeTransitivityAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ExplicitImmutableAttributeTransitivityAnalyzer.cs
@@ -73,38 +73,49 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				return;
 			}
 
-			// So we don't have [Immutable]: check if we should for some reason
+			// The BaseList is the syntax for the ": A, B, C" part which may
+			// include a base class and any number of interfaces. We aren't
+			// looking at type.Interfaces and type.BaseType (from the semantic
+			// model) because we don't want to produce multiple diagnostics for
+			// partial classes. Instead, we attach any diagnostic to whatever
+			// partial declaration mentions a base class or interface that is
+			// [Immutable].
 
-			foreach( var iface in type.Interfaces ) {
-				if( hasTheImmutableAttribute( iface ) ) {
-					context.ReportDiagnostic(
-						CreateDiagnostic(
-							declaredThing: type,
-							declSyntax: decl,
-							otherThing: iface,
-							otherThingKind: "interface"
-						)
-					);
-					return;
-				}
-			}
-
-			// Only thing left to check is the base type
-
-			if( type.BaseType == null ) {
+			if ( decl.BaseList == null ) {
+				// No base class or interfaces on this decl so nothing to check.
 				return;
 			}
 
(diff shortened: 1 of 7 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ExplicitImmutableAttributeTransitivityAnalyzer.cs
```diff
@@ -0,0 +1,126 @@
...
+
+		public override void Initialize( AnalysisContext context ) {
+			//context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( CompilationStart );
+		}
+
+		public void CompilationStart(
+			CompilationStartAnalysisContext context
+		) {
+			var immutableAttribute = context.Compilation.GetTypeByMetadataName(
```
j3parker: FYI: I get really annoyed seeing analyzers compare symbols via strings. This is the proper way to do it. The immutability stuff does string-based stuff a lot.

tests/D2L.CodeStyle.Analyzers.Test/SpecRunner.cs
```diff
@@ -272,7 +272,7 @@ ImmutableArray<string> arguments
 
 				return new DiagnosticExpectation(
 					name: name,
-					arguments: arguments.Split( ',' ).ToImmutableArray()
+					arguments: arguments.Split( ',' ).Select( arg => arg.Trim() ).ToImmutableArray()
```
j3parker: This is the thing that runs the spec tests. Instead of having to say `/* SomeDiag(arg1,arg2,arg3) */ blah /**/` I wanted to be able to say `/* SomeDiag(arg1, arg2, arg3) */ blah /**/`

j3parker:
Tested the code fix in the LMS by hand and it was good.

j3parker:
Caveats found in practice:

(1) It will create diagnostics for every partial declaration and thus apply `[Immutable]` to each of them. This will cause another error diagnostic because `[Immutable]` is not `AllowMultiple`. Not sure yet how I want to fix this (or if I want to in this PR.)

(2) It doesn't handle the `Except` stuff. I'm adding a code fix for that in #412 and am reasonably happy to just keep it separate even if the UX is a little lame (gotta run two fixes)

j3parker:
The main code base now passes these changes. I intend to merge this shortly.

PR #414, opened 2018-09-11
Remove the Types.Audited feature

j3parker:


src/D2L.CodeStyle.Analyzers/Immutability/KnownImmutableTypes.cs
```diff
@@ -74,19 +72,7 @@ internal sealed class KnownImmutableTypes {
...
-
-		internal KnownImmutableTypes( ImmutableHashSet<string> declaredKnownImmutableTypes ) {
-			DeclaredKnownImmutableTypes = declaredKnownImmutableTypes;
-		}
-
-		internal KnownImmutableTypes( IAssemblySymbol a )
-			: this( LoadFromAssembly( a ).ToImmutableHashSet() ) { }
-
-		internal bool IsTypeKnownImmutable( ITypeSymbol type ) {
+		internal static bool IsTypeKnownImmutable( ITypeSymbol type ) {
```
j3parker: This may not stay `static` because I want to remove the stringly-based comparison of symbols in here but for now this should do.

PR #416, opened 2018-09-13
Add fix that seals classes

j3parker:
We'd like this to be rendered into PR comments but don't have the
infrastructure for that yet. This can still be run in bulk though. The
diagnostic remains at INFO level.


j3parker:
Hm, found a flaw:

image

It needs to delete `virtual` from other decls too.

j3parker:
Urg, don't want to merge this without fixing that but it will require a significant change to do cross-document fixes. We can still run this by hand but I'm closing this for now.

PR #422, opened 2018-09-24
IPlugins support in the SingletonLocator analyzer

cbortos-d2l:
for `IPlugins`, check the generic type for the Singleton attribute


cbortos-d2l:
There's another IPlugins interface: 

```
public interface IPlugins<TExtensionPoint, out T> : IEnumerable<T> where TExtensionPoint : IExtensionPoint<T> { }
```

I'll add code to handle that one, too.

PR #426, opened 2018-10-09
Custom service locator analyzer

grant-cleary:
This is a preliminary start at a test analyzer to prevent future usage of the `TestServiceLocator.Create` methods. Custom service locators are expensive to instantiate and slow down tests significantly, which is less than ideal, so we would far prefer that devs use the default locators instead.

I tested this out in two ways:
1. Building `Lms.sln` and iterating on the whitelist until no errors remained (dropping the built DLL right into the package directory).
2. Running full_all_noupdate with the built DLL in the package directory to catch a few instances that were missed in the first pass.

I realize this isn't foolproof, so if anyone else has a better idea on how to test this sort of thing, please let me know and I'll be happy to take care of it.

Thoughts/suggestions on improvements are, of course, always welcome. I'm worried about the scope of this analyzer, but I'll add those comments in the PR directly.


src/D2L.CodeStyle.TestAnalyzers/ServiceLocator/CustomTestServiceLocatorAnalyzer.cs
```diff
@@ -0,0 +1,152 @@
...
+			bool isLocatorFactory = actualType.Equals( disallowedType );
+			return isLocatorFactory;
+		}
+
+		private static bool IsCreateMethod(
+			IMethodSymbol method
+		) {
+			return method.Name.Equals( "Create" )
+				&& method.Parameters.Length > 0;
+		}
```
grant-cleary: So I worry about this guy a bit. Specifically, one use case of these custom locators is to mock out interfaces like `IEventNotifier`, `IConfigManager`, etc., that have undesirable side effects. Sometimes these are easy to fix (i.e. the offending interface is a direct dependency of the class being tested), but other times those dependencies run substantially deeper.

If we feel like this is setting too high a bar, one idea I had was to differentiate between usages of these methods that set a non-null (non-default) `dependencyLoader` parameter (warning in this case), and throwing an error in other cases. i.e. There's always a better solution to `Create( orgId, orgUnitId, userId )`, but that's not necessarily true if we're overloading a dependency.

j3parker:
Whoops I missed this. I'll try to review it today.

grant-cleary:
No worries; I don't think there's any real rush on this. I planned to do a final sweep to catch any exceptions I might have missed prior to merging anyways. :)

src/D2L.CodeStyle.TestAnalyzers/ServiceLocator/CustomServiceLocatorAnalyzerWhitelist.cs
```diff
@@ -0,0 +1,797 @@
...
+
+namespace D2L.CodeStyle.TestAnalyzers.ServiceLocator {
+
+	internal static class CustomServiceLocatorAnalyzerWhitelist {
+
+		/// <summary>
+		/// A list of test fixtures or helper classes that already use custom
+		/// service locators.
+		/// </summary>
+		internal static readonly ImmutableHashSet<string> WhitelistedClasses = ImmutableHashSet.Create(
```
j3parker: Wow, really?

src/D2L.CodeStyle.TestAnalyzers/ServiceLocator/CustomTestServiceLocatorAnalyzer.cs
```diff
@@ -0,0 +1,152 @@
...
+					return;
+				}
+
+				if( symbolInfo.CandidateSymbols.Length != 1 ) {
+					return;
+				}
+
+				// This happens on method groups, such as
+				// Func<IFoo> fooFunc = OldAndBrokenServiceLocator.Get<IFoo>;
+				method = symbolInfo.CandidateSymbols.First() as IMethodSymbol;
```
j3parker: Huh, didn't realize this was how that shook out. So `.Symbol` is actually `null` here?

j3parker: I guess this came from the `OldAndBrokenServiceLocator` analyzer so yup

src/D2L.CodeStyle.TestAnalyzers/ServiceLocator/CustomTestServiceLocatorAnalyzer.cs
j3parker: Yeah. So the current impl of custom object activators is just so horrible that we need to rid of all users. It's true that there are legitimateish usages like you describe. On the other hand... unit tests are better than big integration tests. Definitely some stuff to think about...

j3parker:
Reaallllllly sorry for dropping this. I'm bad at keeping on top of GItHub. Looks good to me. Let's talk next/steps rollout tomorrow?

src/D2L.CodeStyle.TestAnalyzers/ServiceLocator/CustomServiceLocatorAnalyzerWhitelist.cs
grant-cleary: Yeah :(. I've seen a few PRs from Patrick that remove a couple of these, but there are still a lot.

src/D2L.CodeStyle.TestAnalyzers/ServiceLocator/CustomTestServiceLocatorAnalyzer.cs
grant-cleary: Yeah, I think maybe it's not the worst thing in the world if we ruffle some feathers by banning them outright - adding elements to the whitelist isn't hard, and if we can make people think about whether there's a better way to do what they want to do, that's probably better?

src/D2L.CodeStyle.TestAnalyzers/ServiceLocator/CustomTestServiceLocatorAnalyzer.cs
grant-cleary: My testing suggested that virtually everything falls under the case where `.Symbol` isn't `null`, but I figured I would play it safe in case there were use cases I had missed. I should edit the comment though...

grant-cleary:
Again, no worries. This wasn't super urgent. We can certainly chat about next steps. :)

src/D2L.CodeStyle.TestAnalyzers/Common/Diagnostics.cs
```diff
@@ -41,5 +41,14 @@ public static class Diagnostics {
...
 		);
+		
+		public static readonly DiagnosticDescriptor CustomServiceLocator = new DiagnosticDescriptor(
+			id: "D2LTESTS005",
+			title: "Use the default test service locator.",
+			messageFormat: "Custom service locators should not be used. Use static TestServiceLocator.Get<T>() or TestServiceLocatorFactory.Default instead.",
+			category: "Cleanliness",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true,
+			description: "Custom service locators are expensive to instantiate and slow down tests significantly. Use the default locator instead."
```
grant-cleary: Boo, broken merge commit.

tests/D2L.CodeStyle.TestAnalyzers.Tests/D2L.CodeStyle.TestAnalyzers.Tests.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="NUnit" Version="3.12.0" />
     <PackageReference Include="NUnit3TestAdapter" Version="3.13.0" />
     <PackageReference Include="System.Collections.Immutable" Version="1.3.1" />
+    <Compile Include="ServiceLocator\CustomTestServiceLocatorAnalyzerTests.cs" />
```
grant-cleary: This also seems weird. I probably shouldn't have tried to resolve merge conflicts normally on something this old...

src/D2L.CodeStyle.TestAnalyzers/ServiceLocator/CustomTestServiceLocatorAnalyzer.cs
```diff
@@ -140,13 +154,37 @@ IMethodSymbol method
...
 			bool isWhiteListed = _excludeKnownProblems
-				&& CustomServiceLocatorAnalyzerWhitelist
-					.WhitelistedClasses
-					.Contains( classSymbol.ToString() );
+				&& whitelistedClasses.Contains( classSymbol.ToString() );
 
 			return isWhiteListed;
 		}
+
+		private HashSet<string> GetWhitelist(
```
j3parker: nit: you could return `ImmutableHashSet<string>` from here if you use `ImmutableHashSet.CreateBuilder<string>()` instead of `new HashSet<string>()` in the function (and return with `.ToImmutable()`

grant-cleary: Good thought. I expect I'm going to make some more changes in here anyways, so it seems worth doing.

src/D2L.CodeStyle.TestAnalyzers/ServiceLocator/CustomTestServiceLocatorAnalyzer.cs
```diff
@@ -0,0 +1,189 @@
...
+			);
+
+			if( whitelistFile == null ) {
+				return whitelistedClasses.ToImmutableHashSet();
+			}
+
+			SourceText whitelistText = whitelistFile.GetText();
+
+			foreach( TextLine line in whitelistText.Lines ) {
+				whitelistedClasses.Add( line.ToString().Trim() );
```
grant-cleary: I figure the `Trim()` here can't hurt. Feedback on adding bad whitespace in here would be almost instantaneous anyways, but it can still be a pretty irritating problem to track down if you're not looking for it.

grant-cleary:
Whitelist has been merged, so I believe this should be good to go whenever you're comfortable with it being merged.

j3parker:
Thanks! I'll take care of the version bump stuff for you.

grant-cleary:
Thanks Jacob!

PR #428, opened 2018-10-10
Force users to reference D2L.CodeStyle.Annotations

j3parker:
This is required so that GetTypeByMetadataName resolves for our
annotations. Without explicitly referencing annotations the semantic
model does not have complete information for those attributes.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Build/MandatoryReferencesAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Build/MandatoryReferencesAnalyzer.cs
@@ -1,0 +1,38 @@
+﻿using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.Analyzers.Build {
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	public sealed class MandatoryReferencesAnalyzer : DiagnosticAnalyzer {
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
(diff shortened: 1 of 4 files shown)


src/D2L.CodeStyle.Analyzers/Build/MandatoryReferencesAnalyzer.cs
```diff
@@ -0,0 +1,38 @@
...
+					Diagnostic.Create(
+						Diagnostics.MustReferenceAnnotations,
+						Location.None
+					)
+				);
+			}
+		}
+
+		public static bool IsTheAnnotationsAssembly( MetadataReference mr ) {
+			return mr.Display.EndsWith( "\\D2L.CodeStyle.Annotations.dll" );
```
j3parker: This isn't fantastic. We could maybe investigate strong-naming the assembly.

j3parker:
After looking in to how Roslyn works more I think this is definitely what we should be doing. Kinda lame though.

j3parker:
I'm not sure how to add a code fix for this because we'd have to locate D2L.CodeStyle.Annotations correctly.

I should update our VS templates to have these by default too.

src/D2L.CodeStyle.Analyzers/Build/MandatoryReferencesAnalyzer.cs
JeffAshton: If this is a path, Consider using Path.GetFileName then.

j3parker: Hm, so the docs for `Display` say its a "path or name". I'm not really sure what that means but I've obviously assumed it's a path. I should probably figure out what's actually happening.

j3parker:
Talked IRL

j3parker:
The changes to the LMS necessary to turn this on are in PR.

PR #431, opened 2018-10-18
Blacklisting generic implementations

JeffAshton:
Blacklisting generic implementations of IEventHandler<UserInteractionEvent> and IOrgEventHandler<UserInteractionEvent>

image


j3parker:
Cool. I like what this is doing.

I'm not sure this really makes sense as a general purpose thing though:

* There's not a big difference between blocking a generic type and a non-generic type. I'd expect a general analyzer to support both (it's just a matter of accepting generic types as input as well regular types)
* Why only block base classes/implementing? The error message says `"The generic type foo is blacklisted"` which sounds a lot more general to me. I can understand why this is desired for this specific case (event handlers) but I'm doubtful that it's a common case.

I think this would look better as an analyzer specifically for event handlers. If something comes up again that has a lot of overlap we could generalize but while `n = 1` I think specific would be clearer.

j3parker:
And if we simplified it, I'd expect to just have a `switch` instead of the `GenericTypeArgumentsBlacklist`, etc.
Like just keep it as simple as possible right now.

JeffAshton:
@j3parker Renamed

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlersBlacklist.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.Collections.Generic;
+using System.Collections.Immutable;
+
+namespace D2L.CodeStyle.Analyzers.ApiUsage.Events {
+
+	internal static class EventHandlersBlacklist {
+
+		private static readonly ImmutableArray<ImmutableArray<string>> BlacklistedEventHandlerTypes = ImmutableArray.Create(
```
j3parker: Why `ImmutableArray<ImmutableArray<string>>`?

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlersBlacklistAnalyzer.cs
```diff
@@ -0,0 +1,125 @@
...
+			}
+
+			foreach( ImmutableArray<string> genericTypeArguments in genericTypeArgumentSets ) {
+				if( TryGetGenericType( compilation, genericTypeDefinition, genericTypeArguments, out INamedTypeSymbol genericType ) ) {
+					yield return genericType;
+				}
+			}
+		}
+
+		private static bool TryGetGenericType(
```
j3parker: Ok I see why you had `ImmutableArray<ImmutableArray<string>>` up there, multiple generic type arguments (but we only ever have one for the two event handlers.)

Don't you think it'd be simpler to just have

```cs
ImmutableHashSet<INamedTypeSymbol> badThings = ImmutableArray.Create(
  "D2L.LP.Distributed.Events.Handlers.IEventHandler<D2L.Core.JobManagement.Events.JobAbortRequested>",
  "D2L.LP.Distributed.Events.Handlers.IOrgEventHandler<D2L.Core.JobManagement.Events.JobAbortRequested>",

  "D2L.LP.Distributed.Events.Handlers.IEventHandler<D2L.LP.Distributed.WorkQueues.Events.WorkQueueEvent>",
  "D2L.LP.Distributed.Events.Handlers.IOrgEventHandler<D2L.LP.Distributed.WorkQueues.Events.WorkQueueEvent>",

  // ...
  )
  .Select( t => compilation.GetTypeByMetadataName )
  .ToImmutableHashSet();
```

with none of the (sorta complicated) helper methods. Obviously there's some duplication in there (the generic type name) and you have to be "careful" to add both but it's just so much less code. That's what I was trying to get at in the previous comment.

If you're attached to the way it's working now that's fine, and you can go back to what you had before if you want. The difference between this and what I remember in the initial PR is pretty small and doesn't really hit what I was talking about, so either way IMO.

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlersBlacklist.cs
j3parker: Nvm, see below

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlersBlacklistAnalyzer.cs
JeffAshton: I had originally started with the duplication, but decided it was better to have a little more code and remove the human error prone aspect.

PR #433, opened 2018-10-23
Enforcing [Event] attribute on publish calls.

JeffAshton:


src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventPublisherEVentTypesAnalyzer.cs
```diff
@@ -0,0 +1,131 @@
...
+
+			bool hasEventAttr = eventTypeSymbol
+				.GetAttributes()
+				.Any( attr => attr.AttributeClass.Equals( eventAttributeType ) );
+
+			if( hasEventAttr ) {
+				return;
+			}
+
+			if( eventTypeSymbol.TypeKind == TypeKind.TypeParameter ) {
```
JeffAshton: I'm going to take a second pass at these.  A few happened with wrappers:

http://search.dev.d2l/source/xref/Lms/lp/framework/core/D2L.LP/LP/Distributed/Events/Domain/Default/CompositeEventNotifier.cs

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventPublisherEVentTypesAnalyzer.cs
```diff
@@ -0,0 +1,131 @@
...
+			IImmutableSet<ISymbol> genericPublishMethods = Enumerable
+				.Concat(
+					GetGenericPublishMethods( compilation, IEventNotifierFullName ),
+					GetGenericPublishMethods( compilation, IEventPublisherFullName )
+				)
+				.ToImmutableHashSet();
+
+			context.RegisterSyntaxNodeAction(
+					ctxt => {
+						if( ctxt.Node is InvocationExpressionSyntax invocation ) {
```
j3parker: I don't think this guard is necessary. Pretty sure I checked this in Roslyn's source. I just assume the cast passes when I write analyzers.

j3parker: i.e. if you want to pull it out of the context for your analyze method this works:

```cs
context.RegisterSyntaxNode(
  ctx => AnalyzeMethodInvocation(
    ctx,
    ctx.Node,
    eventAttributeType,
    genericPublishMethods
  )
);
```

j3parker: - [ ] remove unnecessary guard

JeffAshton: Switched to a cast.  I'd rather an invalid cast exception then a null reference exception.

j3parker: Don't worry, its just that they couldn't get the API they really wanted in C#. The cast will never fail.

j3parker: But either way is good, ya hard casting it is probably what I'd do too on second thought

PR #436, opened 2018-10-23
Generalize AddImmutableAttributeCodeFix into AddAttributeCodeFix

j3parker:


j3parker:
git doesn't render a good diff. You can diff the two fixer files to see what actually changed (they moved locations)

j3parker:
Only blocker is confirming this works. We don't have a good testing set-up for fixes yet so I'm going to manually validate it.

j3parker:
@JeffAshton validated this with his analyzer.

PR #438, opened 2018-11-05
Event handler loader types analyzer

JeffAshton:


src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlerLoaderTypesAnalyzer.cs
```diff
@@ -0,0 +1,181 @@
...
+					eventHandlerSymbol.ToDisplayString()
+				);
+
+			context.ReportDiagnostic( diagnostic );
+		}
+
+		private static IEnumerable<ISymbol> GetGenericRegisterMethods( Compilation compilation ) {
+
+			INamedTypeSymbol registryType = compilation.GetTypeByMetadataName( IEventHandlerRegistryFullName );
+			if( registryType.IsNullOrErrorType() ) {
```
j3parker: Doesn't make sense for `GetTypeByMetadataName` (here and elsewhere) to return an error type and it definitely won't because of this line: http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/Symbols/AssemblySymbol.cs,584 Not a big deal I just like to thing through these more precisely.

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlerLoaderTypesAnalyzer.cs
```diff
@@ -0,0 +1,181 @@
...
+
+			bool hasAttr = eventTypeSymbol
+				.GetAttributes()
+				.Any( attr => attr.AttributeClass.Equals( eventAttributeType ) );
+
+			if( hasAttr ) {
+				return;
+			}
+
+			if( eventTypeSymbol.TypeKind == TypeKind.TypeParameter ) {
```
j3parker: Hmmm... any concern about things getting through via this?

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlerLoaderTypesAnalyzer.cs
```diff
@@ -0,0 +1,181 @@
...
+			Diagnostic diagnostic = Diagnostic.Create(
+					Diagnostics.EventTypeMissingEventAttribute,
+					invocation.GetLocation(),
+					eventTypeSymbol.ToDisplayString()
+				);
+
+			context.ReportDiagnostic( diagnostic );
+		}
+
+		private static void InspectEventHandlerType(
```
j3parker: Would it make sense to also check that the Guid parses? Could be a follow up PR.

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlerLoaderTypesAnalyzer.cs
JeffAshton: Don't think this is need for now.  Removed

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlerLoaderTypesAnalyzer.cs
JeffAshton: I feel that's a different analyzer which focuses on the other class definition.

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventHandlerLoaderTypesAnalyzer.cs
JeffAshton: I'll do a pass and replace this type of thing broader with an TryGet extension method

PR #439, opened 2018-11-06
Making event types immutable

JeffAshton:


src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventTypesAnalyzer.cs
```diff
@@ -0,0 +1,104 @@
...
+				SyntaxNodeAnalysisContext context,
+				ClassDeclarationSyntax declaration,
+				INamedTypeSymbol attributeType
+			) {
+
+			if( attributeType == null ) {
+				return false;
+			}
+
+			foreach( AttributeListSyntax attrList in declaration.AttributeLists ) {
```
JeffAshton: @j3parker ,  Do you know of a better way to get all attributes for given ITypeSymbol?

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -270,5 +270,15 @@ string name
 
 			return -1;
 		}
+
+		public static bool TryGetTypeByMetadataName(
```
JeffAshton: Created because of your comment in the last PR.  Will switch other places to use this, instead of remember null checks

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventTypesAnalyzer.cs
j3parker: Yep, going through semantic objects (vs. syntactic) like `ITypeSymbol` will work better. Here's an example:

https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Immutability/ExplicitImmutableAttributeTransitivityAnalyzer.cs#L47-L49

j3parker: Sorry, linked to the wrong lines. Fixed.

j3parker:
Cool. I'll finish checking #436 and we can add your diagnostic to the list.

src/D2L.CodeStyle.Analyzers/ApiUsage/Events/EventTypesAnalyzer.cs
JeffAshton: Thanks

PR #443, opened 2018-11-08
Refact extract try get constructor

omsmith:
On top of https://github.com/Brightspace/D2L.CodeStyle/pull/442

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/DependencyRegistrationsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/DependencyRegistrationsAnalyzer.cs
@@ -208,13 +208,22 @@ namespace D2L.CodeStyle.Analyzers.ApiUsage.DependencyInjection {
 			return ImmutableArray.Create( registration.DependencyType );
 		}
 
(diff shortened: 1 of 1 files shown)


j3parker:
Your commit is actually in the middle of #442's commits. Rebase?

j3parker:
(Probably doesn't actually matter because of squash)

omsmith:
Hmm, yeah. Rebase, but I think GitHub orders by date or something?

src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/DependencyRegistrationsAnalyzer.cs
```diff
@@ -208,13 +208,22 @@ ITypeSymbol type
 			return ImmutableArray.Create( registration.DependencyType );
 		}
 
-		private bool TryGetDependenciesFromConstructor( ITypeSymbol type, out ImmutableArray<ITypeSymbol> dependencies  ) {
-			var ctor = type.GetMembers()
+		private bool TryGetInjectableConstructor( ITypeSymbol type, out IMethodSymbol injectableConstructor ) {
```
omsmith: Dunno if there's a reason everything can't be `INamedTypeSymbol` (don't expect so), but it has a `.Constructors` if we did

j3parker:
Oh, ya, I only looked at GitHub. Grumble grumble.

PR #444, opened 2018-11-08
D2L0046: DI registration requires a public constructor

omsmith:
~On top of https://github.com/Brightspace/D2L.CodeStyle/pull/443~

While working on "useless factories" codefix, I realized I may have been rewriting dependency registrations to concrete types without a public constructor. Adding this analyzer so I can better verify the changes I was making during compile time instead of waiting for activation diagnostics

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/DependencyRegistrationsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/DependencyRegistrationsAnalyzer.cs
@@ -27,7 +27,8 @@ namespace D2L.CodeStyle.Analyzers.ApiUsage.DependencyInjection {
 		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create( 
 			Diagnostics.UnsafeSingletonRegistration,
 			Diagnostics.RegistrationKindUnknown,
-			Diagnostics.AttributeRegistrationMismatch
+			Diagnostics.AttributeRegistrationMismatch,
+			Diagnostics.DependencyRegistraionMissingPublicConstructor
 		);
 
 		public override void Initialize( AnalysisContext context ) {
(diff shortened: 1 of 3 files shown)


src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/DependencyRegistrationsAnalyzer.cs
```diff
@@ -208,13 +219,56 @@ ITypeSymbol type
...
+			ITypeSymbol type = registration.FactoryType ?? registration.ConcreteType;
+
+			if( type.IsNullOrErrorType() ) {
+				instantiatedType = null;
+				return false;
+			}
+
+			// handle RegisterSubInterface<IFoo, IFooBar>():
+			// IFooBar is in the registration as "concrete type" and don't want to deal with that refactoring
+			if( type.TypeKind == TypeKind.Interface ) {
```
j3parker: The disconnect between the code that is looking for `RegisterSubInterface` and this code is pretty high.
Maybe we should revisit `DependencyRegistration`? Sounds like it isn't modelling our DI as good as it could be...

src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/DependencyRegistrationsAnalyzer.cs
```diff
@@ -208,6 +219,40 @@ ITypeSymbol type
 			return ImmutableArray.Create( registration.DependencyType );
 		}
 
+		private bool TryGetInstantiatedTypeForRegistration( DependencyRegistration registration, out ITypeSymbol instantiatedType ) {
```
j3parker: Name is a bit akward. Maybe a doc comment explaining what it does (gets the factory for `.RegisterFactory` stuff, or the concrete type when that makes sense.) I can't think of a great alternative name...

j3parker:
Neat. Same issue with the commit ordering, but this one is an ancestor of #443 so as long as that one is fine we're good.

src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/DependencyRegistrationsAnalyzer.cs
omsmith: For sure, it's probably enough to do one or the other with `Marker` (only the dependency type) and ignore the other. Or ignore the call all together? Don't know what analysis is done on it right now.

PR #445, opened 2018-11-09
Add DefaultValueConsistencyForOverriddenMethodsAnalyzer

j3parker:
Wanted to get your thoughts about this before I put 100% effort into it.

Remaining work:
[] Tests
[] Fixes (for each of the diagnostics, doing what's requested in their diagnostic messages)

image

At InFusion we talked about explicit interface implementation to prevent some of these confusing problems. I'm on board with that too, generally. This is a little more conservative and easier to roll out ASAP. Maybe we could start with implicit impls as a warning?

This also obviously handles base classes which is nice.


ppaskaris-d2l:
Default value aren't valid for explicitly implemented interfaces, so you'll at least need an exception for that, but this sure is neat.

src/D2L.CodeStyle.Analyzers/Language/DefaultValueConsistencyForOverriddenMethodsAnalyzer.cs
```diff
@@ -0,0 +1,135 @@
...
+				SyntaxKind.MethodDeclaration
+			);
+		}
+
+		public static void AnalyzeMethodDeclaration(
+			SyntaxNodeAnalysisContext ctx
+		) {
+			var methodDecl = (MethodDeclarationSyntax)ctx.Node;
+
+			var hostDecl = methodDecl.Parent as TypeDeclarationSyntax;
```
j3parker: `.Parent` comes from the syntax base class. I'm assuming here that its a `TypeDeclarationSyntax` (particular subset of all syntax). I.e. every method decl has a parent of either a [class, struct or interface decl syntax](http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/Syntax/TypeDeclarationSyntax.cs,726b9d80aa76a768,references). I think that's true but haven't checked.

I should have used a stronger cast though. Will fix.

j3parker:
@ppaskaris-d2l oh ya, of course. Thanks.

j3parker:
This would make changing things across multiple repos in the LMS result in either (1) multi-PR-merge-racing or (2) really tedious "blue green" type transitions. I'm not sure I care about reducing (1) by weakening stuff like analyzers though..

src/D2L.CodeStyle.Analyzers/Language/DefaultValueConsistencyForOverriddenMethodsAnalyzer.cs
j3parker: Traced up through the parser and they will actually parse methods in some weird places just so they can continue on and find more diagnostics. For example:

image

So I'll keep the weak cast and bail if `null`.

j3parker: [Link to Roslyn code](http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/Parser/LanguageParser.cs,c55189f366e3bc94,references)

j3parker:
Ok, from chats it sounds like people are cool with the idea.

I implemented interfaces and added tests.

@ppaskaris-d2l would you mind checking this to see if I missed weird scenarios? Looking at the test cases can help (the `/* DiagnosticName */ syntax /**/` block comments are assertions about where the red squiggles should show up, the two comments are delimiters.)

ppaskaris-d2l:
Sure yeah. Sorry, I don't get email notifications when you replace the history via force pushing, apparently.

ppaskaris-d2l:
Maybe a test for junk like this?

```csharp
interface IBaseInterface {
	void GetStuff( int count = 99 );
}

interface IMiddleInterface : IBaseInterface {
	new void GetStuff( int count = 55 );
}

class SomeClass : IMiddleInterface {
	public void GetStuff( int count = 99 ) {
		//  ...
	}

	void IMiddleInterface.GetStuff( int count ) {
		// ..
	}
}
```

j3parker:
Urg I forgot about `new`. Hissssss.

j3parker:
Added that and a bit as extra test cases, thanks.
I agree it should probably be an error because it's evil. Would also be open to considering banning the `new` modifier more because we don't use it much and it's a confusion factory.

j3parker:
Cleaning up the code base with this before merging. Found 3 problems with the analyzer and one quality-of-life issue.

src/D2L.CodeStyle.Analyzers/Language/DefaultValueConsistencyAnalyzer.cs
```diff
@@ -181,7 +181,16 @@ IMethodSymbol implMethod
 				var implDefault = implParameter.ExplicitDefaultValue;
 				var baseDefault = baseParameter.ExplicitDefaultValue;
 
-				if( !implDefault.Equals( baseDefault ) ) {
+				if ( implDefault == null && baseDefault == null ) {
```
j3parker: Opps this was unnecessary stuff beacuse of the `Equals` below, sorry. I first wrote this and realized it should already exist and then found the static method... Will remove

PR #446, opened 2018-11-13
Remove accidentally committed redundant check

j3parker:
The Equals below covers this. I wrote this first and then realized
object.Equals probably had a static version for times when things could
be `null`...


j3parker:
It's fine. I also added this after you approved, and merged before hearing back. I'm guilty of thinking our internal bitbucket practices are better than what I practice on GitHub.

PR #449, opened 2018-11-16
Don't NullReferenceException when we can't resolve the base method

j3parker:


src/D2L.CodeStyle.Analyzers/Language/DefaultValueConsistencyAnalyzer.cs
```diff
@@ -64,13 +64,23 @@ SyntaxNodeAnalysisContext ctx
 				return;
 			}
 
-			if( method.IsOverride ) {
-				AnalyzeMethod(
-					ctx.ReportDiagnostic,
-					baseMethod: method.OverriddenMethod,
-					implMethod: method
-				);
+			if( !method.IsOverride ) {
```
j3parker: Could probably omit this given the line below. Meh.

tests/D2L.CodeStyle.Analyzers.Test/Specs/DefaultValueConsistencyAnalyzer.cs
```diff
@@ -1,14 +1,20 @@
...
 	public void FunctionOutOfPlace( int arg = 3 );
 
 	// This has a BaseListSyntax but should be ignored
 	enum Foo : int {
 		A, B, C
 	}
+
+	public class EmptyBase { }
+
+	public class ThingWithMysteriousOverride : EmptyBase {
```
j3parker: Could only repro with a base class other than `System.Object`, shrug.

PR #451, opened 2018-11-18
tests: restore spec test cases that were using ObjectScope.Thread

omsmith:


j3parker:
Haha thanks that was last of me.

PR #453, opened 2018-11-29
Adding D2L.Lms.Dropbox to old and broken service locator Whitelist

mtjandra:
* This is after discussion on retrieving impersonation context for publishing events

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/ServiceLocator/OldAndBrokenServiceLocatorAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/ServiceLocator/OldAndBrokenServiceLocatorAnalyzer.cs
@@ -136,6 +136,7 @@ namespace D2L.CodeStyle.Analyzers.ApiUsage.ServiceLocator {
 			"D2L.Lms.Discussions.Implementation",
 			"D2L.Lms.Domain",
 			"D2L.Lms.Domain.DataAccess",
+			"D2L.Lms.Dropbox",
 			"D2L.Lms.Dropbox.Integration",
 			"D2L.Lms.Dropbox.Web",
 			"D2L.Lms.Question",


src/D2L.CodeStyle.Analyzers/ApiUsage/ServiceLocator/OldAndBrokenServiceLocatorAnalyzer.cs
```diff
@@ -136,6 +136,7 @@ List<INamedTypeSymbol> disallowedTypes
 			"D2L.Lms.Discussions.Implementation",
 			"D2L.Lms.Domain",
 			"D2L.Lms.Domain.DataAccess",
+			"D2L.Lms.Dropbox",
```
JeffAshton: What's the use case?  I feel like this is not the direction we want to go

mtjandra: I've had a lot of back and forth on the correct path forward. I've been consulting mainly with @j3parker and @toddlang . Best reach out to them as they have a much stronger understanding and opinion on the correct way forward at this time.

The problem in general is that I need to get ImpersonationContext in a singleton that publishes events to the BDP. The impersonation context is part of the information needed within the BDP event.

1. Tried fanning it out by passing impersonation ID through arguments. But a little too much so was looking for a happy medium.
2. Thinking of using a `ImpersonationContextInfoProvider`, then deems it does not make sense to have a class like this
3. Suggested using old and broken service locator

mtjandra: Yes I don't either. At this point in time we can't go with solution number 1, although when suggested by members of little foot I was open to it and experimented with it. At the time we though solution number 2 would be much better, since we are creating provider for userId, httpContext, orgUnitContext, etc. But after Jacob has some discussions it was decided this was not great. This is the latest option we've come to.

I'm not clear on what you mean by refactoring out. The methods we need to publish are all within one singleton called `AssignmentEventPublisher`

j3parker: No, (2) makes sense but there is some complexity with implementing it correctly. OBSL will have the same bugs as what you were going to do with the context provider but without enabling the bugs to spread.

j3parker: ((2) isn't the only option either, another approach is a publisher that auto-fills impersonation context when necessary. That's a more involved change but has lots of benefits.)

PR #459, opened 2018-12-11
Don't complain about an arg being unnamed if its a literal that matches the param name

j3parker:
This is a little more polite.

As a reader of code, it does mean that you need to 

1. Know that we treat this case specially
2. Trust the analyzer/know its current limits (e.g. the very high 20 arg threshold)

but I think when we make it complain about literals and inter-changeable arguments (2) will be less of a problem. (1) is only significant for people who are already careful on reviews and they are more likely to become aware of (1) anyway.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
@@ -1,4 +1,5 @@
-﻿using System.Collections.Generic;
+﻿using System;
+using System.Collections.Generic;
 using System.Collections.Immutable;
 using D2L.CodeStyle.Analyzers.Extensions;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


j3parker:
(fixed typo)

j3parker:
(Fixed typo)

ppaskaris-d2l:
This is to make piping stuff like `orgId, orgUnitId, userId` to called methods less tedious?

j3parker:
Yep!

j3parker:
Might want to strip `m_` prefixes too:

```cs
void DoThing( OrgUnitId srcOrgUnitId, OrgUnitId dstOrgUnitId );

// ...

// This is sufficiently readable IMO but if we always error'd on swappable unnamed args
// we'd require this to be named
Copy( m_srcOrgUnitId, m_dstOrgUnitId ); 
```

That could be a follow-up...

j3parker:
In addition to above, looking for `SimpleMemberAccessSyntax`s so that the following would also not error would be consistent:

```cs
Copy( dto.SrcOrgUnitId, dto.DstOrgUnitId );
```

PR #460, opened 2018-12-11
Do a better job at implicit argument names

j3parker:


src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -154,6 +156,37 @@ ArgumentListSyntax args
...
+					ident = access.Name.Identifier.ValueText;
+					break;
+			}
+
+			if( ident == null ) {
+				return null;
+			}
+
+			// Strip uninteresting prefixes off the identifier
+			if ( ident.StartsWith( "m_" ) || ident.StartsWith( "s_" ) ) {
```
ppaskaris-d2l: We do `s_something` in the LMS..? I've never seen that..

j3parker: It is uncommon but does happen. It happens more in some of the libraries (which also use these analyzers.)

j3parker: I personally don't do it. I see a bit of an argument for it but *shrug*.

PR #461, opened 2018-12-12
Disallow more than 8 unnamed arguments

j3parker:


src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -32,7 +32,7 @@ ArgumentSyntax syntax
 		);
 
 		// TODO: shrink this number over time. Maybe 5 would be good?
-		public const int TOO_MANY_UNNAMED_ARGS = 20;
+		public const int TOO_MANY_UNNAMED_ARGS = 9;
```
j3parker: Getting a lot less shameful!

j3parker: This is the only real change. The rest is adjusting tests.

PR #470, opened 2019-02-26
Storageable > Prevent new usages of PhysicalPath property

cbortos-d2l:
added analyzers to prevent new usages of 
- `ILpContentFile.PhysicalPath`
- `ILegacyLpContentFile.PhysicalPath`
- `ILegacyLpContentDirectory.FullName`

**Q:** Why didn't I use the `DangerousMemberUsagesAnalyzer` instead?
**A:** There are several things I did differently in the new analyzer:
1. it supports a white list
1. it detects usages in explicit implementations accessed via concrete types
1. it reports the diagnostic on the symbol access (as opposed to reporting it on the containing method)

The second point above is the **main** reason for having the separate analyzer.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/ContentPhysicalPaths/ILpContentFilePhysicalPathAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/ContentPhysicalPaths/ILpContentFilePhysicalPathAnalyzer.cs
@@ -2,21 +2,23 @@
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.Diagnostics;
 
-namespace D2L.CodeStyle.Analyzers.ApiUsage.PhysicalPaths {
+namespace D2L.CodeStyle.Analyzers.ApiUsage.ContentPhysicalPaths {
 
 	[DiagnosticAnalyzer( LanguageNames.CSharp )]
 	internal sealed class ILpContentFilePhysicalPathAnalyzer : DiagnosticAnalyzer {
 
 		private const string TypeName = "D2L.LP.Files.Domain.ILpContentFile";
+		private const string PropertyName = "PhysicalPath";
 
 		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
-				ContentFilePhysicalPathPropertyAnalysis.DiagnosticDescriptor
+				PhysicalPathPropertyAnalysis.DiagnosticDescriptor
 			);
 
 		public override void Initialize( AnalysisContext context ) {
 
-			ContentFilePhysicalPathPropertyAnalysis analysis = new ContentFilePhysicalPathPropertyAnalysis(
+			PhysicalPathPropertyAnalysis analysis = new PhysicalPathPropertyAnalysis(
 					TypeName,
+					PropertyName,
 					WhitelistedTypes
 				);
 
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/ContentPhysicalPaths/PhysicalPathPropertyAnalysis.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/ContentPhysicalPaths/PhysicalPathPropertyAnalysis.cs
@@ -6,22 +6,24 @@ using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 
-namespace D2L.CodeStyle.Analyzers.ApiUsage.PhysicalPaths {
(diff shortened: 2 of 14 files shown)


src/D2L.CodeStyle.Analyzers/ApiUsage/ContentPhysicalPaths/PhysicalPathPropertyAnalysis.cs
```diff
@@ -0,0 +1,164 @@
...
+
+			IImmutableSet<ISymbol> dangerousProperties = GetDangerousPropertySymbols( context.Compilation );
+
+			if( dangerousProperties.Count == 0 ) {
+				return;
+			}
+
+			context.RegisterSyntaxNodeAction(
+				ctxt => {
+					if( ctxt.Node is MemberAccessExpressionSyntax propertyAccess ) {
```
j3parker: So this will only get called when `ctxt.Node is SimpleMemberAccessExpression` is true... isn't `MemberAccessExpressionSyntax` more general? So I think this `if` is unnecessary because the case should always succeed.

src/D2L.CodeStyle.Analyzers/ApiUsage/ContentPhysicalPaths/PhysicalPathPropertyAnalysis.cs
```diff
@@ -0,0 +1,164 @@
...
+				IImmutableSet<ISymbol> dangerousMembers
+			) {
+		
+			ISymbol originalDefinition = memberSymbol.OriginalDefinition;
+		
+			if( dangerousMembers.Contains( originalDefinition ) ) {
+				return true;
+			}
+		
+			if( !ReferenceEquals( memberSymbol, originalDefinition ) ) {
```
j3parker: nit: usually we just do `memberSymbol != originalDefinition` but this is OK.

src/D2L.CodeStyle.Analyzers/ApiUsage/ContentPhysicalPaths/PhysicalPathPropertyAnalysis.cs
```diff
@@ -0,0 +1,164 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.Analyzers.ApiUsage.ContentPhysicalPaths {
+
+	internal sealed class PhysicalPathPropertyAnalysis {
+
+		internal static readonly DiagnosticDescriptor DiagnosticDescriptor = Diagnostics.ContentPhysicalPathUsages;
```
j3parker: nit: I wouldn't be offended if you just inlined this (I was reading `ReportDiagnostic`, saw that the descriptor was passed down, looked up each caller and saw that it was coming from this property, and then had to find this property to figure out which diagnostic it was... not sure it helps to hide that?)

src/D2L.CodeStyle.Analyzers/ApiUsage/ContentPhysicalPaths/ILpContentFilePhysicalPathAnalyzer.cs
```diff
@@ -0,0 +1,42 @@
...
+namespace D2L.CodeStyle.Analyzers.ApiUsage.ContentPhysicalPaths {
+
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal sealed class ILpContentFilePhysicalPathAnalyzer : DiagnosticAnalyzer {
+
+		private const string TypeName = "D2L.LP.Files.Domain.ILpContentFile";
+		private const string PropertyName = "PhysicalPath";
+
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+				PhysicalPathPropertyAnalysis.DiagnosticDescriptor
```
j3parker: Ah, you had it in a property to share it here.. ok. Shrug.

src/D2L.CodeStyle.Analyzers/ApiUsage/ContentPhysicalPaths/PhysicalPathPropertyAnalysis.cs
j3parker: This was the only non-nit comment

cbortos-d2l: Makes sense. I removed the if-statement. 
To my defense, I copied that code from the `DangerousMemberUsagesAnalyzer` ... so maybe it is not needed there either.

j3parker: Haha probably

Issue #35, opened 2017-01-12
Analyzer should be aware of implicit casts


j3parker:
See #34 for repro

PR #480, opened 2019-04-16
Consider lambda initializers for fields safe

travisMiehm:
Member fields initalized with lambdas (for Func<T>, Action<T>, etc) can
be considered safe, because they will either not contain state, or the
state contained will be analyzed by other analyzers. Initializers that
do no include state include those which reference static members of
other classes, which are also checked with another analyzer

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
@@ -283,24 +283,22 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			ExpressionSyntax expr,
 			HashSet<ITypeSymbol> typeStack
 		) {
-			if( expr.Kind() == SyntaxKind.NullLiteralExpression ) {
+
+			switch( expr.Kind() ) {
 				// This is perhaps a bit suspicious, because fields and
 				// properties have to be readonly, but it is safe...
-				return MutabilityInspectionResult.NotMutable();
-			}
-
-			SemanticModel model = m_compilation.GetSemanticModel( expr.SyntaxTree );
-
-			if( expr.Kind() == SyntaxKind.SimpleLambdaExpression ||
-				expr.Kind() == SyntaxKind.ParenthesizedLambdaExpression
-			) {
+				case SyntaxKind.NullLiteralExpression:
 
 				// Lambda initializers for readonly members are safe
 				// because they can only close over other members, which
 				// will be checked independently, or static members of
 				// another class, which are also analyzed
-				return MutabilityInspectionResult.NotMutable();
+				case SyntaxKind.SimpleLambdaExpression:
+				case SyntaxKind.ParenthesizedLambdaExpression:
+					return MutabilityInspectionResult.NotMutable();
 			}
+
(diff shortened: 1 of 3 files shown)


tests/D2L.CodeStyle.Analyzers.Test/Immutability/MutabilityInspectorTests.cs
```diff
@@ -551,6 +552,67 @@ sealed class Baz { }
 			AssertUnauditedReasonsResult( ty, "ItsUgly", "ItHasntBeenLookedAt" );
 		}
 
+		[Test]
```
j3parker: Could you add these to the spec tests?

travisMiehm: Along side, or instead of?

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -291,6 +291,17 @@ HashSet<ITypeSymbol> typeStack
 
 			SemanticModel model = m_compilation.GetSemanticModel( expr.SyntaxTree );
 
+			if( expr.Kind() == SyntaxKind.SimpleLambdaExpression ||
```
j3parker: nit: you could move this above the `GetSemanticModel` (for performance) and then merge with the other `if` and create a `switch`

travisMiehm: Yep. It was left down there from when I was playing around with the `DataFlowAnalysis` stuff. I'll fix that

j3parker: Oh yea, cool

tests/D2L.CodeStyle.Analyzers.Test/Immutability/MutabilityInspectorTests.cs
j3parker: Either works for me. I'm surprised we didn't have a test case covering this already! (but asserting there was a diagnostic.)

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityInspector.cs
```diff
@@ -283,10 +283,19 @@ HashSet<ITypeSymbol> typeStack
...
-				return MutabilityInspectionResult.NotMutable();
+				case SyntaxKind.NullLiteralExpression:
+
+				// Lambda initializers for readonly members are safe
+				// because they can only close over other members, which
+				// will be checked independently, or static members of
+				// another class, which are also analyzed
+				case SyntaxKind.SimpleLambdaExpression:
+				case SyntaxKind.ParenthesizedLambdaExpression:
+					return MutabilityInspectionResult.NotMutable();
```
omsmith: Did we decide this was safe regardless of the type it was assigned to, and implicit conversions to another type would get us into a potentially-mutable state?

j3parker: We've got another issue open to deal with implicit conversions (#35). It's definitely a problem, but a pretty general one.

PR #482, opened 2019-04-25
Add an analyzer for D2L.StatelessFunc<T>

travisMiehm:
By ensuring that every instantiation of StatelessFunc<T> and friends is
safe, we can convert other classes from Func<T> members to
StatelessFunc<T> and then mark that class [Immutable] safely. This will
allow us to remove a large number of [Mutability.Audited] attributes.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -465,5 +465,15 @@ namespace D2L.CodeStyle.Analyzers {
 			description: "Use the ContentPath property instead."
 		);
 
+		public static readonly DiagnosticDescriptor StatelessFuncIsnt = new DiagnosticDescriptor(
+			id: "D2L0053",
+			title: "StatelessFunc cannot hold state",
+			messageFormat: "StatelessFunc cannot hold state",
+			category: "Safety",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true,
+			description: "StatelessFunc is to be used to hold Func private members, and need to be undiff safe."
+		);
+
 	}
 }
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -468,7 +468,7 @@ namespace D2L.CodeStyle.Analyzers {
 		public static readonly DiagnosticDescriptor StatelessFuncIsnt = new DiagnosticDescriptor(
 			id: "D2L0053",
 			title: "StatelessFunc cannot hold state",
-			messageFormat: "StatelessFunc cannot hold state",
+			messageFormat: "StatelessFunc cannot hold state: {0}",
 			category: "Safety",
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true,
--- a/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
@@ -71,7 +71,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				// lambda has parens
 				// eg () => 1, (x, y) => x + y
 				case SyntaxKind.ParenthesizedLambdaExpression:
-					if( !CheckForClosures( context, argument ) ) {
+					if( !HasCaptures( context, argument ) ) {
 						return;
 					}
 					break;
@@ -80,7 +80,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				// lambda does not have parens
 				// eg x => x + 1
 				case SyntaxKind.SimpleLambdaExpression:
-					if( !CheckForClosures( context, argument ) ) {
+					if( !HasCaptures( context, argument ) ) {
 						return;
 					}
 					break;
@@ -124,7 +124,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			return symbol.IsStatic;
 		}
 
-		private static bool CheckForClosures(
+		private static bool HasCaptures(
 			SyntaxNodeAnalysisContext context,
 			ExpressionSyntax expression
 		) {
--- a/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
@@ -25,9 +25,10 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 
 			context.RegisterSyntaxNodeAction(
 				ctx => {
-					if( ctx.Node is ObjectCreationExpressionSyntax expr ) {
-						AnalyzeObjectCreationExpression( ctx, expr, statelessFuncs );
-					}
+						AnalyzeObjectCreationExpression(
+							ctx,
+							statelessFuncs
+						);
 				},
 				SyntaxKind.ObjectCreationExpression
 			);
@@ -35,9 +36,10 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 
 		private static void AnalyzeObjectCreationExpression(
 			SyntaxNodeAnalysisContext context,
-			ObjectCreationExpressionSyntax syntax,
 			ImmutableHashSet<ISymbol> statelessFuncs
 		) {
+
+			ObjectCreationExpressionSyntax syntax = context.Node as ObjectCreationExpressionSyntax;
 
 			ISymbol symbol = context
 				.SemanticModel
--- a/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
@@ -4,6 +4,7 @@ using Microsoft.CodeAnalysis.CSharp.Syntax;
 using Microsoft.CodeAnalysis.Diagnostics;
 
 using System.Collections.Immutable;
+using System.Linq;
 
 namespace D2L.CodeStyle.Analyzers.Immutability {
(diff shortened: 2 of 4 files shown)


src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -465,5 +465,15 @@ public static class Diagnostics {
 			description: "Use the ContentPath property instead."
 		);
 
+		public static readonly DiagnosticDescriptor StatelessFuncIsnt = new DiagnosticDescriptor(
+			id: "D2L0053",
+			title: "StatelessFunc cannot hold state",
+			messageFormat: "StatelessFunc cannot hold state",
```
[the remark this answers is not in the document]
j3parker: That'd be good. It could be one or more follow-ups though.

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -0,0 +1,153 @@
...
+
+			if( symbol == null ) {
+				return;
+			}
+
+			if( !IsStatelessFunc( symbol ) ) {
+				return;
+			}
+
+			ExpressionSyntax argument = syntax.ArgumentList.Arguments[ 0 ].Expression;
```
j3parker: If `ArgumentList` is nullable I'd check that, and I'd check that there is at least one element of `Arguments` before assuming it. Otherwise the analyzer will crash as you type. I don't think it's possible for an arguments `.Expression` to be `null` though.

j3parker: (`ArgumentList` I could buy being `null` depending on how (the invalid code) `new Foo;` gets parsed.)

j3parker: You can add a spec test for something like `new StatelessFunc()` and it should fail because of an exception right now.

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -0,0 +1,153 @@
...
+					if( !CheckForClosures( context, argument ) ) {
+						return;
+					}
+					break;
+
+				default:
+					// should we do something else here because constructors
+					// of D2L.StatelessFunc<T> and friends should always take
+					// a func?
+					return;
```
j3parker: I think this should probably be a diagnostic

```cs
Func<int> EvilFactory() {
  int x = 0;
  return () => {
    x = x + 1;
    return x;
  };
}

new StatelessFunc<T>( EvilFactory() )
```

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -0,0 +1,153 @@
...
+				.GetSymbolInfo( expression ).Symbol;
+
+			if( symbol == null ) {
+				return false;
+			}
+
+			return symbol.IsStatic;
+		}
+
+		private static bool CheckForClosures(
```
j3parker: nit: it might be more obvious what the return value is if this were named `HasCaptures` or `CapturesData` or something (vs. "Check" - it wasn't obvious to me which direction it was checking for)

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -0,0 +1,153 @@
...
+
+		private static bool IsStatelessFunc( ISymbol symbol ) {
+			ImmutableHashSet<string> statelessFuncs = GetStatelessFuncTypes();
+
+			string fullName = $"{ symbol.ContainingNamespace.Name }.{ symbol.MetadataName }";
+
+			return statelessFuncs.Contains( fullName );
+		}
+
+		private static ImmutableHashSet<string> GetStatelessFuncTypes() {
```
j3parker: Could this just become a `private static readonly ImmutableHashSet<string>` field/property? Rather than the builder etc.

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -0,0 +1,153 @@
...
+				.AnalyzeDataFlow( expression );
+
+			ImmutableArray<ISymbol> captures = dataFlow.Captured;
+			return captures.Length > 0;
+		}
+
+		private static bool IsStatelessFunc( ISymbol symbol ) {
+			ImmutableHashSet<string> statelessFuncs = GetStatelessFuncTypes();
+
+			string fullName = $"{ symbol.ContainingNamespace.Name }.{ symbol.MetadataName }";
```
j3parker: We can avoid strings if at compilation time we look up the `ISymbol`s for the constructors we care about like this: https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs#L21

j3parker:
Generally looks good! I left a few comments but they're mostly minor

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
travisMiehm: This case is an `InvocationExpressionSyntax`. Sounds like you're proposing we reject those to start?

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
j3parker: Ok, cool!

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
j3parker: I'm proposing we reject everything by default :)

travisMiehm: Done

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
travisMiehm: Fixed

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
travisMiehm: Done

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -0,0 +1,189 @@
...
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+			Diagnostics.StatelessFuncIsnt
+		);
+
+		public override void Initialize( AnalysisContext context ) {
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( RegisterAnalysis );
+		}
+
+		private void RegisterAnalysis( CompilationStartAnalysisContext context ) {
```
j3parker: nit: maybe later make these `static`

It's not important. I tend to look for it when there is `EnableConcurrentExecution` because then I know I only need to worry about other `static` members and captures (ahem) but like you don't have any here so it's not actually important.

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -0,0 +1,189 @@
...
+
+			Compilation compilation = context.Compilation;
+
+			context.RegisterSyntaxNodeAction(
+				ctx => {
+					if( ctx.Node is ObjectCreationExpressionSyntax expr ) {
```
j3parker: more nits:

This case will always succeed, feel free to change it to `AnalyzerInvocation( ctx, ctx.Node as ObjectCreationExpression )` (or you could do the cast in `AnalyzeInvocation` and just pass `AnalyzeInvocation` directly here.

Consider renaming it to `AnalyzeObjectCreationExpression` since they're different things.

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -0,0 +1,189 @@
...
+				"D2L.StatelessFunc`1",
+				"D2L.StatelessFunc`2",
+				"D2L.StatelessFunc`3",
+				"D2L.StatelessFunc`4",
+				"D2L.StatelessFunc`5",
+				"D2L.StatelessFunc`6",
+			};
+
+			foreach( string typeName in types ) {
+				INamedTypeSymbol typeSymbol = compilation.GetTypeByMetadataName( typeName );
```
j3parker: It might be nice to cache this `ImmutableHashSet<ISymbol>`. I don't 100% remember what the perf of `GetTypeByMetadataName` looked like (probably pretty fast once caches are primed, at least.)

How I've done this before is something like call `GetStatelessFunctTypes` in `RegisterAnalysis` and then pass it to `AnalyzeInvocation` (that way its still different per compilation, but it gets totally cached.

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -0,0 +1,189 @@
...
+
+				if( typeSymbol == null ) {
+					// These types are usually defined in another assembly,
+					// ( usually ) and thus we have a bit of a circular
+					// dependency. Allowing this lookup to return null
+					// allow us to update the analyzer before the other
+					// assembly, which should be safer
```
j3parker: Nice comment

PR #485, opened 2019-05-07
add an analyzer for a [StatelessFunc] attribute

omsmith:
(Future?) TODOs:
- Remove the ObjectCreation analysis, and just do `public StatelessFunc( [StatelessFunc] Func<> func )`
- Expand to trust `StatelessFunc<>` as an argument
- Expand to trust references to a local argument that's marked `[StatelessFunc]`

Why I think this could be beneficial:

- Introduce analysis into an API without requiring `new StatelessFunc`age
- `StatelessFunc<>` is only needed for storage
- Reasons?

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
@@ -1,4 +1,4 @@
-﻿using D2L.CodeStyle.Analyzers.Extensions;
+using D2L.CodeStyle.Analyzers.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
--- a/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
@@ -125,14 +125,36 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 					);
 
 					break;
-
+				
+				/**
+				 * This is the case where a variable is passed in
+				 * eg Foo( f )
+				 * Where f might be a local variable, a parameter, or field
+				 * 
+				 * class C<T> {
+				 *   StatelessFunc<T> m_f;
+				 *   
+				 *   void P( StatelessFunc<T></T> f ) { Foo( f ); }
+				 *   void Q( [StatelessFunc] Func<T> f ) { Foo( f ); }
+				 *   void R() { StatelessFunc<T> f; Foo( f ); }
+				 *   void S() { Foo( m_f ); }
+				 * }
+				 */
 				case SyntaxKind.IdentifierName: {
 
+					/**
+					 * Doesn't matter where it came from, if it's a StatelessFunc<T> we're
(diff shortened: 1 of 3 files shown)


omsmith:
Pushed some more things to play with the noted TODOs, but still looking for thoughts in general

tests/D2L.CodeStyle.Analyzers.Test/Specs/StatelessFuncAnalyzerSpec.cs
```diff
@@ -85,6 +107,41 @@ internal sealed class Usages {
...
+			var func = new StatelessFunc<int>( f );
+			AttributeFuncReceiver.Accept( f );
+		}
+
+		public void StatelessFuncFromParam( StatelessFunc<int> f) {
+			var func = new StatelessFunc<int>( f );
+			AttributeFuncReceiver.Accept( f );
+		}
+
+		public void StatelessFuncAttrParam( [StatelessFunc] Func<int> f ) {
```
omsmith: I guess this is a little broken since I could do `f = () => this.things`...?

travisMiehm: Maybe? But isn't this also one of the primary ways you wanted to be able to do this?

travisMiehm:
I like the idea. Even smaller changes to convert something, which was a big motivator for the current design.

tests/D2L.CodeStyle.Analyzers.Test/Specs/StatelessFuncAnalyzerSpec.cs
omsmith: Why the frig doesn't C# have readonly vars/parameters yet

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -22,72 +23,85 @@ public sealed class StatelessFuncAnalyzer : DiagnosticAnalyzer {
 		private static void RegisterAnalysis( CompilationStartAnalysisContext context ) {
 
 			Compilation compilation = context.Compilation;
+			ISymbol statelessFuncAttr = compilation.GetTypeByMetadataName( "D2L.StatelessFuncAttribute" );
```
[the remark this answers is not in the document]
j3parker: Ah yeah. If it lives in annotations you can assume that.

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -165,19 +225,45 @@ ExpressionSyntax expression
 			ISymbol symbol,
 			ImmutableHashSet<ISymbol> statelessFuncs
 		) {
-
-			if( statelessFuncs.Contains( symbol ) ) {
```
travisMiehm: Was this just an unnecessary check?

travisMiehm: Ah, we're calling this on arguments now, not constructors. So the type info is different, and this became superfluous.

omsmith: I think it was also always unnecessary yes. The OriginalDefinition of the original defintion, is still itself, so it's always okay to just do OriginalDefinition.

PR #486, opened 2019-05-08
D2LXXXX: [ReadOnly] parameters

omsmith:
Provides an attribute and analyzer for marking parameters as read only (can't be assigned to or passed by non-readonly reference).

The attribute can be subclassed, so other attributes can get the same behaviour (such as `[StatelessFunc]`).

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -475,5 +475,15 @@ namespace D2L.CodeStyle.Analyzers {
 			description: "StatelessFunc is to be used to hold Func private members, and need to be undiff safe."
 		);
 
+		public static readonly DiagnosticDescriptor ReadOnlyParameterIsnt = new DiagnosticDescriptor(
+			id: "D2L0054",
+			title: "Parameter is not readonly",
+			messageFormat: "Parameter is not readonly: {0}",
+			category: "Safety",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true,
+			description: "[ReadOnly] paramaters must not be assigned to or passed by non-readonly reference."
+		);
+
 	}
 }
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ReadOnlyParameterAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ReadOnlyParameterAnalyzer.cs
@@ -1,0 +1,117 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+using System.Collections.Immutable;
+using System.Linq;
+
+namespace D2L.CodeStyle.Analyzers.Immutability {
+
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal sealed class ReadOnlyParameterAnalyzer : DiagnosticAnalyzer {
+
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+			Diagnostics.ReadOnlyParameterIsnt
+		);
+
+		public override void Initialize( AnalysisContext context ) {
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( RegisterAnalysis );
+		}
+
+		private static void RegisterAnalysis( CompilationStartAnalysisContext context ) {
+
+			Compilation compilation = context.Compilation;
+
+			INamedTypeSymbol readOnlyAttribute = compilation.GetTypeByMetadataName( "D2L.CodeStyle.Annotations.ReadOnlyAttribute" );
+
+			if( readOnlyAttribute == null ) {
+				return;
+			}
+
+			context.RegisterSyntaxNodeAction(
+				ctx => AnalyzeParameter(
+					ctx: ctx,
+					readOnlyAttribute: readOnlyAttribute,
+					syntax: ctx.Node as ParameterSyntax
+				),
+				SyntaxKind.Parameter
+			);
+		}
+
+		private static void AnalyzeParameter(
+			SyntaxNodeAnalysisContext ctx,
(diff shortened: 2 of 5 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ReadOnlyParameterAnalyzer.cs
```diff
@@ -0,0 +1,124 @@
...
+						continue;
+					}
+
+					if( returnedThing.OriginalDefinition != parameter ) {
+						continue;
+					}
+
+					ctx.ReportDiagnostic( Diagnostic.Create(
+						Diagnostics.ReadOnlyParameterIsnt,
+						syntax.GetLocation(),
```
omsmith: hmm, this one would be reasonable to switch to `returnStatement.GetLocation()` (though perhaps the wording should change then?). For the dataflow, I didn't want to try to find where the write came from.

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -475,5 +475,15 @@ public static class Diagnostics {
...
 		);
 
+		public static readonly DiagnosticDescriptor ReadOnlyParameterIsnt = new DiagnosticDescriptor(
+			id: "D2LXXXX",
+			title: "Parameter is not readonly",
+			messageFormat: "Parameter is not readonly: {0}",
+			category: "Safety",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true,
+			description: "[ReadOnly] paramaters must not be assigned to (or possibly assigned to)."
```
omsmith: `s/(or possible assigned to)/or passed by non-readonly reference/`

omsmith:
Removed "ref return" checks, since testing it out, you need to explicitly `return ref foo`, and you can't do that to parameters unless they're ref/out themselves, which we already restrict.

src/D2L.CodeStyle.Analyzers/Immutability/ReadOnlyParameterAnalyzer.cs
```diff
@@ -0,0 +1,117 @@
...
+				) );
+			}
+
+			IMethodSymbol method = parameter.ContainingSymbol as IMethodSymbol;
+			BlockSyntax methodBody = ( method.DeclaringSyntaxReferences[0].GetSyntax( ctx.CancellationToken ) as MethodDeclarationSyntax ).Body;
+
+			DataFlowAnalysis dataflow = model.AnalyzeDataFlow( methodBody );
+			if( dataflow.WrittenInside.Contains( parameter ) ) {
+				/**
+				 * public void Foo( [ReadOnly] int foo ) {
```
j3parker: So why are we worried about this? Attributes on parameters are kind of "for the caller", and the caller shouldn't need to care about this... there is no way they can observe this happening and its equivalent to assigning foo to bar and then manipulating bar. I'm not sure why we want to prevent harmless code like this?

j3parker: Harmless, and the alternative is just more verbose

j3parker: Maybe the hang-up for me is just that we're not filtering out value types? (`delegate`/`Func`/etc. are reference types)
But in the reply about `in` you mentioned that it not working for value types was bad. I guess I'm not sure why.

j3parker:
In your example `in` isn't necessary, you should just pass by value to get the desired behaviour...?

omsmith:
`StatelessFunc<T, ..., TResult>` and it's corresponding analyzer was introduced to allow Func fields to be part of otherwise `[Immutable]` objects. The one downside of it is that in order to use it, your API needs to change from `Func<>` to `StatelessFunc<>`, and every caller thus needs to add `new StatelessFunc( ... )` for your new API to roll out.

In order to make the transition easier / less invasive, I thought of adding a `[StatelessFunc]` attribute as well, that could be added to `Func<>` parameters in order to get the analysis, without requiring changes to every caller (modulo callers being stateful).

That is to say, I wanted to be able to do something like below, where `StatelessFunc<>` is buried well within my API, only used for storage, and the callers (`Spigot*`) can use `Func<>` or `StatelessFunc<>`, whatever.

```charp
[Immutable]
class SpigotA {
  private static readonly Sink m_sink = Pipe.Do( () => "hello' );
}
```
```
[Immutable]
class SpigotB {
  private static readonly StatelessFunc m_func = new StatelessFunc( () => "hello" );
  private static readonly Sink m_sink = Pipe.Do( m_func );
}
```
```csharp
static class Pipe {
  public static Sink Do( [StatelessFunc] Func<string> func ) {
    return new Sink( func );
  }
}
```
```csharp
[Immutable]
class Sink {
  private readonly StatelessFunc<string> m_func;

  public Sink( [StatelessFunc] Func<string> func ) {
    m_func = new StatelessFunc( func );
  }
}
```

Pipe and Sink being the somewhat interesting parts here. Even though we can't see the function definition, I want to trust that `[StatelessFunc]` means that we analyzed what went in.

And so I updated the analyzer to go from "creations of `StatelessFunc<>`" to "arguments passed to `[StatelessFunc]` parameters" with the following ru

omsmith:
Oops, fat fingers, still working on typing the above ^

omsmith:
Another approach for non-invasive introduction of stateless func checking would be implicit operators for `Func<> -> StatelessFunc<>`, but I don't know how to write an analyzer that would catch any arbitrary conversion.

j3parker:
Yeah, analyzing for implicit conversions is a pain. I think it's better in the operator-based analyzers (which we haven't used yet, but I think they're available without a feature toggle now.)

The `[StatelessFunc]` attribute itself sounds awesome. I'm still a bit skeptical of uses for `[ReadOnly]` outside of this case, but OK. The idea is for `StatelessFuncAttribute` to subclass `ReadOnlyAttribute` to avoid two attributes?

I'd be fine with the logical all rolled into `[StatelessFunc]` but I won't object to this on it's own.

j3parker:
Thanks for the detailed response btw - it really helped to understand the context.

omsmith:
I'll roll with it separate for now. Can always change later.

PR #487, opened 2019-05-09
ReadOnlyParameterAnalyzer: handle constructors

omsmith:


omsmith:
I had a bunch of constructors in the StatelessFunc, but it wasn't actually bringing in this analyzer. Oops

PR #490, opened 2019-05-10
Add analyzer for nameof usage in ConfigTestSetupAttribute

ppaskaris-d2l:
This PR adds the suggested analyzer to prevent `[ConfigTestSetup( "SymbolName" )]` usage, recommending `nameof(SymbolName)` usage instead. This was Owen's suggestion on my PR in core.

### DIFF
--- a/src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
@@ -1,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.CodeAnalysis;
+
+namespace D2L.CodeStyle.TestAnalyzers.Extensions {
+	public static class RoslynExtensions {
+
+		// Copied from the non-test assembly because we do not reference it.
+
+		public static bool IsNullOrErrorType( this ITypeSymbol symbol ) {
+
+			if( symbol == null ) {
+				return true;
+			}
+
+			if( symbol.Kind == SymbolKind.ErrorType ) {
+				return true;
+			}
+
+			if( symbol.TypeKind == TypeKind.Error ) {
+				return true;
+			}
+
+			return false;
+		}
+
+		public static bool IsNullOrErrorType( this ISymbol symbol ) {
+
+			if( symbol == null ) {
+				return true;
+			}
+
+			if( symbol.Kind == SymbolKind.ErrorType ) {
+				return true;
+			}
+
+			return false;
+		}
+
+	}
+}
--- a/src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
@@ -1,0 +1,107 @@
+using System.Collections.Immutable;
+using System.Linq;
+using D2L.CodeStyle.TestAnalyzers.Common;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.TestAnalyzers.NUnit {
+
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	public sealed class ConfigTestSetupStringsAnalyzer : DiagnosticAnalyzer {
+
+		private const string AttributeName = "ConfigTestSetupAttribute";
+
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
+			=> ImmutableArray.Create( Diagnostics.ConfigTestSetupStrings );
+
+		public override void Initialize( AnalysisContext context ) {
+
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( Register );
+		}
+
+		private void Register( CompilationStartAnalysisContext compilation ) {
+
+			compilation.RegisterSyntaxNodeAction(
+					AnalyzeSyntaxNode,
+					SyntaxKind.MethodDeclaration,
+					SyntaxKind.ClassDeclaration
+				);
+		}
+
+		private void AnalyzeSyntaxNode( SyntaxNodeAnalysisContext context ) {
+
+			SyntaxList<AttributeListSyntax> attributeList;
+
+			switch( context.Node ) {
+
+				case MethodDeclarationSyntax method:
+					attributeList = method.AttributeLists;
+					break;
+
+				case ClassDeclarationSyntax @class:
+					attributeList = @class.AttributeLists;
+					break;
+
+				default:
+					return;
+			}
+
+			AttributeSyntax[] attributes = attributeList
+				.SelectMany( x => x.Attributes )
+				.ToArray();
+
+			// Method has no attributes
+			if( !attributes.Any() ) {
+				return;
+			}
+
+			foreach( AttributeSyntax attribute in attributes ) {
+
+				ISymbol symbol = context
+					.SemanticModel
+					.GetSymbolInfo( attribute )
+					.Symbol;
+
+				if( symbol == null ) {
+					continue;
+				}
+
+				// Not a [ConfigTestSetup()]
+				string displayString = symbol.ToDisplayString(
+						SymbolDisplayFormat.FullyQualifiedFormat
+					);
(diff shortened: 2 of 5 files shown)


src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
```diff
@@ -0,0 +1,107 @@
...
+			}
+
+			foreach( AttributeSyntax attribute in attributes ) {
+
+				ISymbol symbol = context
+					.SemanticModel
+					.GetSymbolInfo( attribute )
+					.Symbol;
+
+				if( symbol == null ) {
```
j3parker: There are some other cases that might be worth checking http://search.dev.d2l/source/xref/NuGet/D2L.CodeStyle/src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs#219

(these come up as you type, typically.)

src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
```diff
@@ -0,0 +1,107 @@
...
+					continue;
+				}
+
+				// Not a [ConfigTestSetup()]
```
j3parker: The prefered approach is to look up the symbol for ConfigTestSetup in the setup and not register callbacks if it doesn't exist. If it does, compare the strongly typed things rather than strings.

Here's an analyzer that does that kind of stuff: https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs#L25

src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
```diff
@@ -0,0 +1,107 @@
...
+				SeparatedSyntaxList<AttributeArgumentSyntax> arguments =
+					attribute.ArgumentList.Arguments;
+
+				if( arguments.Count != 1 ) {
+					continue;
+				}
+
+				ExpressionSyntax argExpression = arguments.First().Expression;
+
+				// Not [ConfigTestSetup( "foo" )]
```
j3parker: I'm imagining people who love to make `const`s hiding from this analyzer.

Should we just force that its a `nameof` directly?

tests/D2L.CodeStyle.TestAnalyzers.Tests/NUnit/ConfigTestSetupStringsAnalyzerTests.cs
```diff
@@ -0,0 +1,133 @@
+﻿using D2L.CodeStyle.TestAnalyzers.Common;
```
j3parker: Tests like this are fine, but our prefered way to write tests these days is.....
nevermind, not applicable to test analyzers. arghhhhhhhhhhhhhhhhhhh.

j3parker:
Looks cool; might want to check the other analyzer @omsmith has and see if there could be a generalized analyzer. Maybe it makes sense to separate them, I haven't looked.

ppaskaris-d2l:
I wrote these using the existing TestCaseSource, ValueSource, etc. attributes as inspiration. The feedback you've left also applies to them. Generalizing seems neat, once these issues are addressed (next week).

j3parker:
Makes sense! No worries!
Yeah, over time there's a few things we/I've grown nittier about (e.g. strings-vs-symbols)

omsmith:
Yeah - those ones are real old. ~first ones I did I think. Could probably be improved.

src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
ppaskaris-d2l: I did this initially, but I changed it after looking at the adjacent analyzers.

j3parker: Ok, your call

ppaskaris-d2l: I am changing it back now, but it is proving difficult because I didn't actually test my original version (it didn't work, `nameof` is tragic in Roslyn).

src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -0,0 +1,42 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Microsoft.CodeAnalysis;
+
+namespace D2L.CodeStyle.TestAnalyzers.Extensions {
+	public static class RoslynExtensions {
```
ppaskaris-d2l: These projects don't reference each other, and I didn't want to add such a reference without proposing just copying these two helpers first.

src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
```diff
@@ -0,0 +1,146 @@
...
+				Diagnostic diagnostic = Diagnostic.Create(
+						Diagnostics.ConfigTestSetupStrings,
+						argExpression.GetLocation(),
+						messageArg
+					);
+				context.ReportDiagnostic( diagnostic );
+			}
+		}
+
+		private static bool IsNameOfSyntax( ExpressionSyntax expression ) {
```
ppaskaris-d2l: `nameof` is ambiguous, so this sucks a bunch.

```
class Test {

    [Test]
    [ConfigTestSetup( nameof("uwotm8") )]
    public void TestMethod() {
        // Even though this invocation is invalid, the compiler
        // prefers this interpretation if `nameof` resolves to
        // something.
    }

    string nameof( string discard ) {
        return "Foo";
    }
}
```

dotnet/roslyn#35599

src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
```diff
@@ -0,0 +1,146 @@
...
+				(InvocationExpressionSyntax)expression;
+
+			if( !invocation.Expression.IsKind( SyntaxKind.IdentifierName ) ) {
+				return false;
+			}
+
+			IdentifierNameSyntax identifier =
+				(IdentifierNameSyntax)invocation.Expression;
+
+			if( !identifier.Identifier.IsKind( SyntaxKind.IdentifierToken ) ) {
```
omsmith: Give `IsKind( SyntaxKind.NameOfKeyword )` a whirl?

ppaskaris-d2l: Explained in https://github.com/Brightspace/D2L.CodeStyle/pull/490#discussion_r283943964

src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
j3parker: Eventually we plan to roll everything into one DLL. This is fine.

src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
j3parker: Yikers

j3parker: Linking that issue in a comment in this function would be helpful

j3parker:
Do you know how to test this locally with the LMS first?

ppaskaris-d2l:
I do not - is there a document somewhere that explains it?

omsmith:
Copy `bin/{Debug,Release]/D2L.CodeStyle.TestAnalyzers.dll` to `checkout/packatges/D2L.CodeStyle.TestAnalyzers/analyzers/dotnet/cs` and run `full_all_noschema_noupdate.bat` and/or open a sln (you could try `_temp/Lms.sln`, which can be finicky and will take a while to analyze, but can be helpful if trying to apply a codefix).

ppaskaris-d2l:
It sure does work locally ✔️

PR #493, opened 2019-05-13
Allow static methods as IdentiferNames for StatelessFuncs

travisMiehm:


src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -267,6 +275,18 @@ CancellationToken ct
 			return false;
 		}
 
+		private static bool IsStaticMethodIdentifier(
```
omsmith: note: I don't think the differences between this and `IsStaticMemberAccess` are particularly important, and so could be combined in some fashion (member access one doesn't actually do anything specific to member access, for instance).

PR #502, opened 2019-05-23
TooManyUnnamedArgs: actually interpolate message

omsmith:
Alternatively could make this {0} and pass in the arg, but just fixing
the intention here.

- [ ] Fix


j3parker:
Ha!

PR #503, opened 2019-05-23
Mark dependencies as private so this package can be referenced

mthjones:


omsmith:
It can be an attribute OR a nested node? Neat I guess :P

omsmith:
nuspec looks good :+1:

mthjones:
> It can be an attribute OR a nested node?

Yep. All boils down to a `Dictionary<string, string>` of metadata in the end.

mthjones:
Should I be bumping the version in this PR? Seems like if we merge without bumping it'll overwrite the old packages of that version? 🤔

omsmith:
I'll bump it after. We don't autopublish. I suppose there's a project feed buuut, unused

PR #504, opened 2019-05-24
Stop packaging output dll to lib and fix tools path

mthjones:
Packing the build output to lib causes package references to these
analyzers to also pull in the dll to their build output folder like a
standard package. This isn't desired behaviour, since the analyzers
should never impact the user's build output.

Also, the install/uninstall scripts were being packed in `tools/tools`,
so were never picked up properly. They were also missing from the test
analyzer project. These have also been fixed.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,6 +12,9 @@ build:
   parallel: true
   verbosity: minimal
 
+test_script:
+  - dotnet test --no-build
+
 artifacts:
   - path: src\**\*.nupkg
     name: NuGet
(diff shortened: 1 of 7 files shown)


mthjones:
Packaging is broken. Don't merge yet!

mthjones:
`GeneratePackageOnBuild` was giving me grief the first time, so I removed it. But AppVeyor doesn't run a separate pack step, so that means the packages weren't getting created. Not seeing the warnings I was seeing before now though, I think a `dotnet clean` fixed it up.

Should consider switching Appveyor to use `dotnet build`, `dotnet pack` etc. explicitly. It's way less noisy compared to the msbuild output.

PR #508, opened 2019-06-17
D2L0055: Require awaited tasks are configured.

omsmith:
This is unopinionated on true vs false (modulo the CodeFix using false). Just requires something `await`ed must be configured in one direction. Handles `Task` and `ValueTask`.

Uses `SafeAsync()` if `D2L.Core` is referenced. Otherwise uses `ConfigureAwait( continueOnCapturedContext: false )`.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Language/AwaitedTasksAnalyzer.Codefix.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/AwaitedTasksAnalyzer.Codefix.cs
@@ -1,0 +1,118 @@
+﻿using System.Collections.Immutable;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CodeActions;
+using Microsoft.CodeAnalysis.CodeFixes;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace D2L.CodeStyle.Analyzers.Language {
+	partial class AwaitedTasksAnalyzer {
+
+		[ExportCodeFixProvider(
+			LanguageNames.CSharp,
+			Name = nameof( ConfigureAwaitedTaskCodeFix )
+		)]
+		public sealed class ConfigureAwaitedTaskCodeFix : CodeFixProvider {
+
+			public override ImmutableArray<string> FixableDiagnosticIds
+				=> ImmutableArray.Create( Diagnostics.AwaitedTaskNotConfigured.Id );
+
+			public override FixAllProvider GetFixAllProvider()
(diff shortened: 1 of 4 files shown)


src/D2L.CodeStyle.Analyzers/Language/AwaitedTasksAnalyzer.Codefix.cs
```diff
@@ -0,0 +1,118 @@
...
+							.Argument( SyntaxFactory.LiteralExpression( SyntaxKind.FalseLiteralExpression ) )
+							.WithNameColon( SyntaxFactory.NameColon( "continueOnCapturedContext" ) )
+					) ) );
+
+			private static InvocationExpressionSyntax SafeAsyncify( ExpressionSyntax task ) =>
+				SyntaxFactory
+					.InvocationExpression( SyntaxFactory.MemberAccessExpression(
+						SyntaxKind.SimpleMemberAccessExpression,
+						task,
+						SyntaxFactory.IdentifierName( "SafeAsync" )
```
omsmith: There isn't a `SafeAsync` for `ValueTask` yet. Will add to D2L.Core

omsmith: Err, forgot ValueTask is in a nuget package and messing with such dependencies is awful, soooo

Probably not going to see many ValueTasks in the wild still, so going to pretend this isn't a problem.

j3parker: Could we just not use `SafeAsync`? Like if we just required `ConfigureAwait( continueOnCapturedContext: false )` (i.e. including the argument name) then

1) no references to d2l DLLs/vanilla C#
2) all the "safety"
3) its more clear what its doing

omsmith: Are you saying that's something you want to do _now_? If we're going to do that, I think we should switch existing usages as well and then remove the extension methods if we're going to do that.

In our current state, I'd prefer the codefix to produce code that approximates existing code.

```
osmith@KLX1-OSMITH MINGW64 /c/d2l/full/checkout (master)
$ git grep 'ConfigureAwait' | wc -l
84

osmith@KLX1-OSMITH MINGW64 /c/d2l/full/checkout (master)
$ git grep 'SafeAsync' | wc -l
415

osmith@KLX1-OSMITH MINGW64 /c/d2l/full/checkout (master)
$ git grep 'SafeWait' | wc -l
103
```

j3parker: Nah, just it would be cool in the future.

PR #516, opened 2019-06-25
IInstancePlugins immutability and singleton checks

cbortos-d2l:


cbortos-d2l:
For reference, here are the interface definitions: http://search.dev.d2l/source/xref/Lms/lp/framework/core/D2L.LP/LP/Extensibility/Plugins/DI/IInstancePlugins.cs

PR #522, opened 2019-08-28
Added check for [Test] attribute on all public methods in a [TestFixture]

jbheard:


j3parker:
It looks like there are still some whitespace issues. I'll come over and we can see if we need to poke at our code style settings.

j3parker:
I think there was some desire to report errors when we had `[Test]`s in non-`[TestFixture]`s... (maybe that gets hard with those crummy test base classes?) so the control flow would need to change a bit for that, but we can do that later. @travisMiehm thoughts?

travisMiehm:
I think test attributes in non fixtures can be a next step, and the case covered by this change is good.

src/D2L.CodeStyle.TestAnalyzers/Common/Diagnostics.cs
```diff
@@ -51,5 +51,15 @@ public static class Diagnostics {
 			isEnabledByDefault: true,
 			description: "Custom service locators are expensive to instantiate and slow down tests significantly. Use the default locator instead."
 		);
-	}
+
+        public static readonly DiagnosticDescriptor TestAttributeMissed = new DiagnosticDescriptor(
+            id: "D2LTESTS006",
+            title: "Method not labelled as [Test], [TestCase], or [TestCaseSource]",
```
travisMiehm: Ah, sorry, we also have some tests that only use `[Theory]`.

I've skimmed [NUnit's list of attributes](https://github.com/nunit/docs/wiki/Attributes), and I think this is the last "hey, this method is a test" attribute that we use.

jbheard: Added the theory attribute

j3parker:
Ok, we need to wittle down the in-flight analyzer changes :) Let's focus on getting the LMS up-to-snuff now. We can't merge the analyzers until it's ready.

j3parker:
You could make your new diagnostics be info if you want.

PR #524, opened 2019-08-30
Changes to RequireNamedArgumentsAnalyzer

jbheard:
- Fixed bug with params
- Added/fixed some test cases
- Changed literal param severity from error to info


src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -1,223 +1,247 @@
...
+				// which case it is "unnamed", so ignore it.
+				if( param.Name == "" ) {
+					continue;
+				}
+
+				// C# allows us to create variables with the same names as reserved keywords,
+				// as long as we prefix with @ (e.g. @int is a valid identifier)
+				// So any parameters which are reserved must have the @ prefix
+				string paramName;
+				SyntaxKind paramNameKind = SyntaxFacts.GetKeywordKind( param.Name );
```
j3parker: When there is no match this helper returns `SyntaxKind.None`: http://source.roslyn.io/#Microsoft.CodeAnalysis.CSharp/Syntax/SyntaxKindFacts.cs,960

Could we just check for that, and if so add `@`?

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -1,223 +1,247 @@
...
+				if( SyntaxFacts.GetReservedKeywordKinds().Any( reservedKind => reservedKind == paramNameKind ) ) {
```
j3parker: I couldn't find anything on `ISymbol`/`IParameterSymbol` that indicates that the name is a verbatim name. Looking at where that detection happens the data might still be there... check `MetadataName`? etc., we can investigate on Monday.

j3parker:
Let's break this up into smaller PRs, there are multiple unrelated things happening here. We can walk through that on Monday!

jbheard:
This has been split into multiple smaller PRs.

Issue #521, opened 2019-08-27
Require named arguments when any two arguments have interchangable types


j3parker:
This is a bit more of a challenge than #520

Code goes here: https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs#L91
Tests go here: https://github.com/Brightspace/D2L.CodeStyle/blob/master/tests/D2L.CodeStyle.Analyzers.Test/Specs/RequireNamedArgumentsAnalyzer.cs

When a function takes parameters like `int width, int height` or `OrgUnitId src, OrgUnitId dst` it's easy to mix up the order of arguments and it can be very hard to spot this in code reviews or otherwise. So instead of:

```cs
SomeCssThing( top, bottom, left, right );
```

we want:

```cs
SomeCssThing(
  top: top,
  right: bottom,
  bottom: left,
  right: right
);
```

but since we consider arguments where the argument is an identifier which is equivalent to the parameters identifier to be named, we actually only want to be forced to write:

```cs
SomeCssThing(
  top,
  right: bottom,
  bottom: left,
  right
);
```

which really highlights the issue to human readers. Sometimes this is intentionally, but usually not.

Regarding implementation, (this is a guess) we can use [`ClassifyConversion`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.csharpextensions.classifyconversion?view=roslyn-dotnet) to query the semantic model and ask if the two argument expressions could in principle be implicitly swapped, and require that named parameters are used.

We need to be careful to not do too many queries here. We require named arguments at 5 or more arguments, and the check is symmetric so that caps the number of checks we should ever do to 8. We could measure how many checks on average we'd do in our code base.

TimothyJCowen:
Change was deemed too complicated (with exclusions and such), and so is being closed.
Related pull requests are also closed:
https://github.com/Brightspace/lms/pull/5678
https://github.com/Brightspace/lms/pull/5684

TimothyJCowen:
Analyzer changes were here:
https://github.com/Brightspace/D2L.CodeStyle/tree/TJ-RequireNamedArgs

PR #528, opened 2019-09-03
Added check for duplicate argument types

jbheard:
To address issue #521


jbheard:
I should add test cases and fix current breaking tests before I PR

PR #532, opened 2019-10-09
Remove CORS Header analyzer (class being blacklisted has been removed)

grant-cleary:


grant-cleary:
I don't have write access to this repo, so I can't merge it myself, but from my perspective it should be good basically whenever it's convenient for you. :)

PR #541, opened 2020-03-05
More behaviour preserving suggestion

azielinski:
The current suggestion to use JsonConvertParamBinder has caused a number of our APIs to no longer match our published API contracts. StrictJsonParamBinder was created to address this issue and should be used instead of JsonConvertParamBinder in almost all cases.


omsmith:
If this is correct, then https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/ApiUsage/JsonParamBinderAttribute/JsonParamBinderAnalyzerFixer.cs should be updated as well

j3parker:
@ryantmer sound good?

azielinski:
> LGTM - `StrictJsonParamBinder` didn't exist at the time this was added, but if that's the new/correct one to use, then good by me 😄

Yeah, this is a relatively newer thing. The JsonConvertParamBinder is very permissive, in most situations it'll even bind {} and attempt to do "something" with it. In a lot of the old domains this results in weird behaviour if parameters are included twice, left out, etc. The strict binder is... stricter, closely matching the legacy binder. Our Valence docs all say that all claim we do strict validation on the structures and the API team wrote the strict binder to try to match that more closely but it seems like it doesn't get much usage right now. Hoping this will help nudge people the right way.

Issue #515, opened 2019-06-24
Prevent losing Exception context on rethrow


mthjones:
## Problem
When exceptions are rethrown from `catch` blocks, there are two typical ways:

#### Option 1
```csharp
try {
    // do work
} catch( Exception e ) {
    // log, etc.
    throw e;
}
```

#### Option 2
```csharp
try {
    // do work
} catch( Exception ) {
    // log, etc.
    throw;
}
```

The problem is that Option 1 will lose all stacktrace context when rethrowing the exception, making it difficult to track down issues in the code if those errors do happen, while Option 2 retains the complete original exception context.

## Analyzer
I propose adding a new analyzer that looks for uses of Option 1 in the code and suggests switching them to Option 2. To make it simple, it could look for cases where the caught exception variable is rethrown and without being re-assigned in the `catch` block.

### Example
```csharp
try {
    // do work
} catch( Exception e ) {
    // log, etc.
    // D2LXXXX: This exception rethrow will lose all stacktrace context of the original exception.
    // Replace with `throw;` to rethrow the original exception with the complete stacktrace.
    throw e;
}
```

## Additional Considerations
There are potentially cases where the exception is instead captured into a variable that's rethrown outside of the `catch` block, or it's desired to rethrow an existing exception from elsewhere (e.g. unwrapping an exception's `InnerException`). These scenarios can be handled by using `ExceptionDispatchInfo.Capture( e )`, but these scenarios may be much more difficult to create an analyzer for.

omsmith:
I was bored and whipped together a start to this in https://github.com/Brightspace/D2L.CodeStyle/pull/542. Anyone can feel free to bring it over the line.

PR #542, opened 2020-03-10
D2LXXXX: should rethrow exception

omsmith:
Ref: https://github.com/Brightspace/D2L.CodeStyle/issues/515

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
@@ -1,0 +1,80 @@
+﻿using System;
+using System.Collections.Immutable;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Operations;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.Analyzers.Language {
+
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal sealed partial class RethrowAnalyzer : DiagnosticAnalyzer {
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics
+			=> ImmutableArray.Create( Diagnostics.ShouldRethrow );
+
+		public override void Initialize( AnalysisContext context ) {
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( RegisterAnalyzer );
+		}
+
+		private static void RegisterAnalyzer(
+			CompilationStartAnalysisContext context
+		) {
+			context.RegisterOperationAction(
+				Analyze,
+				OperationKind.CatchClause
+			);
+		}
+
+		private static void Analyze(
+			OperationAnalysisContext context
+		) {
+			ICatchClauseOperation operation = context.Operation as ICatchClauseOperation;
+
+			if( operation.ExceptionDeclarationOrExpression == null ) {
+				// try {} catch {}
+				// try {} catch( Exception ) {}
+				return;
+			}
+
+			var throwOperations = operation.Handler.Operations.OfType<IThrowOperation>();
+			if( !throwOperations.Any() ) {
+				// try {} catch( Exception e ) { m_log.Error( e ); }
+				return;
+			}
+
+			SyntaxToken capturedIdentifier = ( operation.ExceptionDeclarationOrExpression.Syntax as CatchDeclarationSyntax ).Identifier;
+
+			SemanticModel model = context.Compilation.GetSemanticModel( operation.Syntax.SyntaxTree );
+			DataFlowAnalysis dataFlow = model.AnalyzeDataFlow( operation.Handler.Syntax );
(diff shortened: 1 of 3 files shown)


src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
```diff
@@ -0,0 +1,80 @@
...
+
+		public override void Initialize( AnalysisContext context ) {
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( RegisterAnalyzer );
+		}
+
+		private static void RegisterAnalyzer(
+			CompilationStartAnalysisContext context
+		) {
+			context.RegisterOperationAction(
```
j3parker: Double check that we can use operation analyzers in the LMS :) At one point they required a feature toggle.

src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
```diff
@@ -0,0 +1,80 @@
...
+			);
+		}
+
+		private static void Analyze(
+			OperationAnalysisContext context
+		) {
+			ICatchClauseOperation operation = context.Operation as ICatchClauseOperation;
+
+			if( operation.ExceptionDeclarationOrExpression == null ) {
+				// try {} catch {}
```
j3parker: So this in particular is evil. Do we want to ban it separately? Is there a Microsoft analyzer for that, or a codestyle?

src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
```diff
@@ -0,0 +1,80 @@
...
+				// try {} catch( Exception ) {}
```
j3parker: This pattern could be bad too - is it cool that we're ignoring it? Future thing to consider or what?

j3parker: I see... we are explicitly looking for `throw`s here, got it

src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
```diff
@@ -0,0 +1,80 @@
...
+				// try {} catch( Exception e ) { m_log.Error( e ); }
+				return;
+			}
+
+			SyntaxToken capturedIdentifier = ( operation.ExceptionDeclarationOrExpression.Syntax as CatchDeclarationSyntax ).Identifier;
+
+			SemanticModel model = context.Compilation.GetSemanticModel( operation.Syntax.SyntaxTree );
+			DataFlowAnalysis dataFlow = model.AnalyzeDataFlow( operation.Handler.Syntax );
+			if( dataFlow.WrittenInside.Any( s => s.Name == capturedIdentifier.ValueText ) ) {
+				// try {} catch( Exception e ) { e = new Exception(); throw e; }
```
j3parker: What use case is this meant to permit?

src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
```diff
@@ -0,0 +1,80 @@
...
+			}
+
+			foreach( var throwOperation in throwOperations ) {
+				if( throwOperation.Exception == null ) {
+					// throw;
+					continue;
+				}
+
+				if( !( throwOperation.Exception.Syntax is IdentifierNameSyntax thrownIdentifier ) ) {
+					// throw new Exception();
```
j3parker: Is it OK that we're losing the stack trace here? Or would that be a different analyzer?

src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
```diff
@@ -0,0 +1,80 @@
...
+					// throw MakeException();
+					continue;
+				}
+
+				if( thrownIdentifier.Identifier.ValueText != capturedIdentifier.ValueText ) {
```
j3parker: nit: I liked the comments above, maybe one here too?

src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
```diff
@@ -0,0 +1,80 @@
...
+					continue;
+				}
+
+				context.ReportDiagnostic( Diagnostic.Create(
```
j3parker: Nit: an example comment here might be nice for symmetry

src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
omsmith: Probably `new WrappedException( e )` or the like?

Assigning it to `e` is weird, but just trying not to be too picky.

Just simply trying to find things that can super-obviously be replaced with `throw;`

src/D2L.CodeStyle.Analyzers/Language/RethrowAnalyzer.cs
mthjones: Enforcing the original exception is defined as the InnerException would be cool here, but since the constructors are only by-convention it might be hard to enforce. There may also be cases where you wouldn't want to store the InnerException, but I can't think of any off-hand.

PR #543, opened 2020-03-23
DE38371 - Adding a move controller to the legacy list of JsonParamBin…

davewagler-d2l:
…der classes


davewagler-d2l:
This class was moved, and as such was updated to use the proper `JsonConvertParamBinder` attribute. This has now created issues for clients that were relying on a value of `"null"` to fail. Although this is a weird issue, the fact that we supported means we should probably continue supporting it until we can communicate that we don't.

davewagler-d2l:
Related PR that introduced this: https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/8088/overview

PR #544, opened 2020-03-29
Omsmith/obsl type whitelist

omsmith:


src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+
+namespace D2L.CodeStyle.Analyzers.Helpers {
+
+	internal static class WhitelistHelper {
+
+		public static WhitelistHelper<INamedTypeSymbol> TypeLevel(
```
j3parker: What's another `T` you have in mind?
The name `TypeLevel` didn't make any sense to me without reading this function.

src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+		public void EntryIsUnnecessary( T entry, Location location, ReportDiagnostic report ) {
+			report( Diagnostic.Create(
+				descriptor: Diagnostics.UnnecessaryWhitelistEntry,
+				location: location,
+				EntryFormatter( entry ),
+				WhitelistFileName
+			) );
+		}
+
+		private ImmutableHashSet<string> LoadWhitelist(
```
j3parker: Does this need to be `string` based?

src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+
+			AdditionalText whitelistFile = additionalFiles.FirstOrDefault(
+				file => Path.GetFileName( file.Path ) == WhitelistFileName
+			);
+
+			if( whitelistFile == null ) {
+				return whitelist.ToImmutableHashSet();
+			}
+
+			SourceText whitelistText = whitelistFile.GetText();
```
j3parker: whiteListFile.GetText().Select( x = > x.Trim().ToImmutableHashSet();

src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+			WhitelistFileName = whitelistFileName;
+			EntryFormatter = entryFormatter;
+		}
+
+		public string WhitelistFileName { get; }
+		public Func<T, string> EntryFormatter { get; }
+
+		public Func<T, bool> LoadWhitelist(
+			AnalyzerOptions options
+		) {
```
j3parker:
```suggestion
		public Func<T, bool> LoadWhitelist( AnalyzerOptions options ) {

```

src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+using System.Collections.Immutable;
+using System.IO;
+using System.Linq;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+
+namespace D2L.CodeStyle.Analyzers.Helpers {
+
+	internal static class WhitelistHelper {
```
j3parker: I don't think the helper suffix is adding anything here

src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
```diff
@@ -0,0 +1,83 @@
...
+			AnalyzerOptions options
+		) {
+			ImmutableHashSet<string> whitelist = LoadWhitelist( options.AdditionalFiles );
+			return IsInWhitelist;
+
+			bool IsInWhitelist( T candidate ) {
+				string entry = EntryFormatter( candidate );
+				bool whitelisted = whitelist.Contains( entry );
+				return whitelisted;
+			}
```
j3parker: The idiomatic way to do this in C# is to have a `static Whitelist.Create-` method which loads the whitelist and returns a `Whitelist` object holding on to the hash set and exposing some method like `IsInWhitelist`. Returning a Func like this is unusual... you've given it a name but it's only a local name which is pretty pointless.

j3parker: The API I'd expect for something like this:

```
public sealed class Whitelist {
  public Whitelist( ImmutableHashSet<INamedTypeSymbol> );

  public static Whitelist CreateFromAdditionalFiles( AdditionalFiles );

  public bool IsInWhitelist( INamedTypeSymbol );
```

src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
omsmith: I assume you mean trying to load them into type infos? I don't have an intuitive sense of how expensive that would be to do each project.

src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
omsmith: :+1:

j3parker: (If the string stuff is necessary (e.g. ambiguous parsing for generic types) then I'd make the constructor `private` and still keep `Contains` taking a `INamedTypeSymbol` and doing the lowering to strings so that the analyzers don't need to do that.)

src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
omsmith: Figured the assembly-level whitelist I'm replacing here coooould be desirable at some point?

src/D2L.CodeStyle.Analyzers/Helpers/WhitelistHelper.cs
j3parker: At least just the symbol part, but ya dno.

omsmith: I don't think whether the whitelist contains uses the formatted string or loaded symbols is a blocker, but avoiding strings as much as possible is cool. Definitely think it's something to continue looking at afterward.

PR #551, opened 2020-06-15
Fix symbol equality comparisons

mthjones:
Prior to MSBuild 16.5, these symbol reference equality comparisons worked. This has been confirmed by analyzing known-broken code on Visual Studio 2019 16.4. With the update to MSBuild 16.5+, these started silently failing instead.

The 3.6.0 release of the CodeAnalysis packages includes an analyzer that catches these invalid equality checks, which was used to identify all of the broken cases.

This also updates the test projects to use the 3.6.0 release of the CodeAnalysis packages. These packages align with the Roslyn version provided with [Visual Studio 2019 and .NET Core 3.1][0].

[0]: https://github.com/dotnet/roslyn/blob/master/docs/wiki/NuGet-packages.md#versioning

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/Domain/DependencyRegistry.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/Domain/DependencyRegistry.cs
@@ -66,7 +66,7 @@ namespace D2L.CodeStyle.Analyzers.ApiUsage.DependencyInjection.Domain {
 		}
 
 		public bool IsRegistationMethod( IMethodSymbol method ) {
-			if( method.ContainingType != m_dependencyRegistryType && method.ReceiverType != m_dependencyRegistryType ) {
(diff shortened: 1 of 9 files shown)


src/D2L.CodeStyle.Analyzers/ApiUsage/DependencyInjection/Domain/DependencyRegistry.cs
```diff
@@ -66,7 +66,7 @@ INamedTypeSymbol dependencyRegistryType
 		}
 
 		public bool IsRegistationMethod( IMethodSymbol method ) {
-			if( method.ContainingType != m_dependencyRegistryType && method.ReceiverType != m_dependencyRegistryType ) {
+			if( !method.ContainingType.Equals( m_dependencyRegistryType ) && !method.ReceiverType.Equals( m_dependencyRegistryType ) ) {
```
mthjones, line 69: This is the only one that causes failing tests.

PR #554, opened 2020-06-16
Add a dependabot configuration

mthjones:
Just a barebones configuration for now. Can tweak it as needed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "nuget"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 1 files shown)


omsmith:
Use the `.github/dependabot.yml` if we don't need any preview features?

j3parker:
There's a UI to switch... I've kinda just been adding the old config and clicking that button haha

j3parker:
The new format is more verbose and I have fewer things to copy and paste it from

mthjones:
Oops, missed the big yellow box at the top here: https://dependabot.com/docs/config-file/

I'll switch it over.

PR #561, opened 2020-06-27
Including this and base constructor calls in RequireNamedArgumentsAna…

JeffAshton:
…lyzer

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
@@ -1,8 +1,6 @@
 ﻿using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Collections.Immutable;
-using System.Diagnostics;
 using System.Linq;
 using D2L.CodeStyle.Analyzers.Extensions;
 using Microsoft.CodeAnalysis;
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -49,26 +47,35 @@ ArgumentSyntax syntax
 				AnalyzeCallSyntax,
 				SyntaxKind.ObjectCreationExpression
 			);
+
```
JeffAshton, line 50: https://www.youtube.com/watch?v=6iCPIUGnHQ8

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -89,22 +96,22 @@ SyntaxNodeAnalysisContext ctx
 				ctx.ReportDiagnostic(
 					Diagnostic.Create(
 						descriptor: Diagnostics.TooManyUnnamedArgs,
-						location: expr.GetLocation(),
+						location: ctx.Node.GetLocation(),
```
JeffAshton, line 99: `expr` was `(ExpressionSyntax)ctx.Node` previously

j3parker:
This is a good fix, but be prepared for a bit of backlog rolling this out. It might make sense to roll this change out on its own.
(cases where more than 5 args used in `this`/`base`)

You can use the code fix and "fix all in solution" to fix those. I often use `Lms.sln` to do it all at once but that thing seems to only be usable every-other VS release 🤷

JeffAshton:
> This is a good fix, but be prepared for a bit of backlog rolling this out. It might make sense to roll this change out on its own.
> (cases where more than 5 args used in `this`/`base`)
> 
> You can use the code fix and "fix all in solution" to fix those. I often use `Lms.sln` to do it all at once but that thing seems to only be usable every-other VS release 🤷

I got a successful build with only fixing 2 spots:

https://git.dev.d2l/projects/CORE/repos/lms/pull-requests/12981/diff

PR #564, opened 2020-06-27
Adding support for requiring named arguments

JeffAshton:


src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -127,6 +152,18 @@ SyntaxNodeAnalysisContext ctx
 			// all the args instead to avoid craziness with overloading.
 		}
 
+		private static ImmutableDictionary<string, string> CreateFixerContext( ImmutableArray<ArgParamBinding> unnamedArgs ) {
```
JeffAshton, line 155: Pulled fixer context generation out into common helper

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -209,19 +248,20 @@ ArgumentListSyntax args
...
-				if( psuedoName != null ) {
-					bool matchesParamName = string.Equals(
-						psuedoName,
-						param.Name,
-						StringComparison.OrdinalIgnoreCase
-					);
-
-					if( matchesParamName ) {
-						continue;
+				if( !requireNamedArguments ) {
```
JeffAshton, line 251: Skip the psuedo pass if named arguments are required

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
j3parker: Thanks

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
j3parker: Hmmm, why though?

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -80,19 +85,38 @@ SyntaxNodeAnalysisContext ctx
 				return;
 			}
 
+			bool requireNamedArguments = HasRequireNamedArgumentsAttribute( ctx.SemanticModel, args );
```
j3parker, line 88: This would be more efficient after the unnamedArgs calculation (this code runs for every fn/etc. call) but only if we stick with the pseudo naming logic (right now this is being passed into that fn so can’t be moved. )

The intent of the pseudo naming logic is that as a reader (that knows about this behaviour) there’s no ambiguity... adding names is just noise. 

It’s more important when we apply this widely i.e. not just for methods with an attribute because it cuts down on noise (think adapters/etc. ) so I don’t feel super strongly in this case.

JeffAshton:
> Consider keeping the pseudo logic for consistency but it’s not a big deal

So I was definitely expecting to see all arguments named.  Obviously a personal preference, but generally speaking if some of the arguments are named, I'd prefer them all named.  

image

I find the anomalies more taxing on my brain then the bonus names.  Everything having the `: ` whitespace makes them all look the same.

JeffAshton:
@j3parker ,  Change my mind and allowed the pseudo named.  Doesn't stop me from doing it my way.  :)

j3parker:
Ya, pseudo-names is definitely not intuitive. There’s the “it makes proxying arguments less noisy thing”, but I think it also makes some kinds of bugs more obvious to the reader  consider this code with all args named:

 ```cs
Foo( int a, int b, int c, int d ) : base(
  a: a,
  b: b,
  c: d,
  d: c
) { ... }
```

vs. this with pseudo names allowed:


```cs
Foo( int a, int b, int c, int d ) : base(
  a,
  b,
  c: d,
  d: c
) { ... }
```

It could be argued that the second one highlights the problem a bit better. We hit a few of these in practice with code that dealt with CSS and didn’t order the top/right/bottom/left (TRBL = trouble) right...

We thought about adding an independent analyzer that looked at relatedness of the names... like if you passed a var named “m_width” (edit distance 2 to width) to height and vice versa then maybe you meant those to go in the other order (idea based on an idea in a Java analyzer from google)... but decided against it because it was too complex. 

The downside is that as a code reader named arguments can be very reassuring, and if you don’t know about this pseudo naming stuff you don’t get that reassurance.

tests/D2L.CodeStyle.Analyzers.Test/Specs/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -212,22 +247,34 @@ public SomeInheritedClass( int a1, bool _ )
...
+				/* NamedArgumentsRequired */ : base( b1, b2, b3 ) /**/{ }
+
 			public SomeInheritedClass( int b1, int b2, int b3, int b4, string _ )
 				: base( b1, b2, b3, b4 ) { }
 
-			public SomeInheritedClass( int b1, int b2, int b3, int b4, int b5, string _ )
+			public SomeInheritedClass( int b1, int b2, int b3, int b4, int b5, string _ )
+				: base( a1: b1, a2: b2, a3: b3, a4: b4, a5: b5 ) { }
+
+			public SomeInheritedClass( int b1, int b2, int b3, int b4, int b5, string _ )
```
j3parker, line 277: Thanks for tests!

PR #568, opened 2020-07-16
Jashton/serializer attribute analyzer

JeffAshton:
Makes sure that `ExternalSerializer` : `ITrySerializer`

```
	[Serializer( typeof( ExternalSerializer ) )]
	public sealed class External { }
	public sealed class ExternalSerializer : ITrySerializer { }
```

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/SerializerAttributeAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/SerializerAttributeAnalyzer.cs
@@ -1,0 +1,122 @@
+﻿using System.Collections.Immutable;
+using System.Linq;
+using D2L.CodeStyle.Analyzers.Extensions;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.Analyzers.ApiUsage.Serialization {
+
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal sealed class SerializerAttributeAnalyzer : DiagnosticAnalyzer {
+
+		private static readonly SymbolDisplayFormat TypeDisplayFormat = new SymbolDisplayFormat(
+				memberOptions: SymbolDisplayMemberOptions.IncludeContainingType,
+				localOptions: SymbolDisplayLocalOptions.IncludeType,
+				typeQualificationStyle: SymbolDisplayTypeQualificationStyle.NameAndContainingTypesAndNamespaces
+			);
+
+		private const string SerializerAttributeFullName = "D2L.LP.Serialization.SerializerAttribute";
+		private const string ITrySerializerFullName = "D2L.LP.Serialization.ITrySerializer";
+
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+			Diagnostics.InvalidSerializerType
+		);
+
+		public override void Initialize( AnalysisContext context ) {
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( RegisterScopeBuilderAnalyzer );
+		}
+
+		private void RegisterScopeBuilderAnalyzer( CompilationStartAnalysisContext context ) {
+
+			Compilation comp = context.Compilation;
+			if( !comp.TryGetTypeByMetadataName( SerializerAttributeFullName, out INamedTypeSymbol serializerAttributeType ) ) {
+				return;
+			}
+			if( !comp.TryGetTypeByMetadataName( ITrySerializerFullName, out INamedTypeSymbol trySerializerType ) ) {
+				return;
+			}
+
+			context.RegisterSyntaxNodeAction(
+					c => AnalyzeAttributeSyntax(
+						c,
+						(AttributeSyntax)c.Node,
+						serializerAttributeType,
+						trySerializerType
+					),
+					SyntaxKind.Attribute
+				);
+		}
+
+		private void AnalyzeAttributeSyntax(
+				SyntaxNodeAnalysisContext context,
+				AttributeSyntax attributeSyntax,
+				INamedTypeSymbol serializerAttributeType,
+				INamedTypeSymbol trySerializerType
+			) {
+
+			SemanticModel model = context.SemanticModel;
+			if( !IsAttributeOfType( model, attributeSyntax, serializerAttributeType ) ) {
+				return;
+			}
+
+
+			AttributeArgumentSyntax typeArgumentSyntax = attributeSyntax.ArgumentList.Arguments[ 0 ];
+			if( !( typeArgumentSyntax.Expression is TypeOfExpressionSyntax typeofSyntax ) ) {
+
+				ReportInvalidSerializerType(
+						context,
+						typeArgumentSyntax,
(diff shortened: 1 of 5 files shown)


src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/SerializerAttributeAnalyzer.cs
```diff
@@ -0,0 +1,121 @@
...
+				INamedTypeSymbol serializerAttributeType,
+				INamedTypeSymbol trySerializerType
+			) {
+
+			SemanticModel model = context.SemanticModel;
+			if( !IsAttributeOfType( model, attributeSyntax, serializerAttributeType ) ) {
+				return;
+			}
+
+			AttributeArgumentSyntax typeArgumentSyntax = attributeSyntax.ArgumentList.Arguments[ 0 ];
```
mthjones: I believe this would crash the analyzer if an attempt was made to compile without specifying args (which would also fail to compile, at least with the current definition, but I believe analyzers will run first).

Consider handling this with some null propagation and `.FirstOrDefault()`.

```suggestion
			AttributeArgumentSyntax typeArgumentSyntax = attributeSyntax.ArgumentList?.Arguments?.FirstOrDefault();
			if( typeArgumentSyntax is null ) {
			    // compilation will fail
			    return;
			}
```

(and test cases for `[Serializer]` and `[Serializer()]`)

src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/SerializerAttributeAnalyzer.cs
```diff
@@ -0,0 +1,121 @@
...
+
+				ReportInvalidSerializerType(
+						context,
+						typeArgumentSyntax,
+						messageArg: typeArgumentSyntax.Expression.ToString()
+					);
+				return;
+			}
+
+			ITypeSymbol serializerType = model.GetTypeInfo( typeofSyntax.Type ).Type;
```
mthjones, line 81: Potential NRE if the type doesn't exist in the model. Consider handling null here as well.

```suggestion
			ITypeSymbol serializerType = model.GetTypeInfo( typeofSyntax.Type )?.Type;
			if( serializerType is null ) {
				// compilation will fail
			    return;
			}
```

tests/D2L.CodeStyle.Analyzers.Test/Specs/SerializerAttributeAnalyzer.cs
```diff
@@ -0,0 +1,39 @@
...
+	[Serializer( /* InvalidSerializerType(System.String) */ typeof( string ) /**/ )]
+	public sealed class InvalidSerializerType { }
+
+	[Serializer( /* InvalidSerializerType(null) */ null /**/ )]
+	public sealed class NullSerializerType { }
+
+	[Serializer( /* InvalidSerializerType(123) */ 123 /**/ )]
+	public sealed class InvalidSerializerTypeArgument { }
+
+	[Serializer( )]
```
mthjones, line 34: Could you add a case for `[Serializer]` as well? It might produce an NRE on accessing `.ArgumentList`.

src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/SerializerAttributeAnalyzer.cs
mthjones: Oops, this was already resolved.

JeffAshton:
> Seems good to me overall. I'm a bit curious why we're defining our own attributes instead of using `ISerializable`/`[Serializable]`, but that's tangential to the analyzer impl.

This is the `D2L.LP.Serialization` framework DJ wrote that does Json & Xml:

https://code.d2l.dev/#D2L.LP/LP/Serialization/Reflection/SerializerAttribute.cs,a906af49a592fd8e,references

This is the custom version of `[ReflectionSerializer]`

tests/D2L.CodeStyle.Analyzers.Test/Specs/SerializerAttributeAnalyzer.cs
JeffAshton: Good catch.  `ArgumentList` is `null` in this case

src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/SerializerAttributeAnalyzer.cs
```diff
@@ -0,0 +1,131 @@
...
+		public override void Initialize( AnalysisContext context ) {
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( RegisterScopeBuilderAnalyzer );
+		}
+
+		private void RegisterScopeBuilderAnalyzer( CompilationStartAnalysisContext context ) {
+
+			Compilation comp = context.Compilation;
+			if( !comp.TryGetTypeByMetadataName( SerializerAttributeFullName, out INamedTypeSymbol serializerAttributeType ) ) {
+				return;
```
j3parker, line 36: good stuff

src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/SerializerAttributeAnalyzer.cs
mthjones: Ah, you're right, my bad. Seems like it's the `Type` itself that would be null then, which would NRE below in `serializerType.AllInterfaces`. Adding a test case for it would be good.

https://github.com/dotnet/roslyn/blob/e704ca635bd6de70a0250e34c4567c7a28fa9f6d/src/Compilers/CSharp/Portable/Compilation/TypeInfo.cs#L13

tests/D2L.CodeStyle.Analyzers.Test/Specs/SerializerAttributeAnalyzer.cs
```diff
@@ -0,0 +1,47 @@
...
+	[Serializer()]
+	public sealed class IncompleteAttribute_WithParentheses { }
+
+	[System.ComponentModel.Category( "Not related" )]
+	public sealed class RandomAttribute { }
+
+	[Invalid]
+	public sealed class InvalidAttribute { }
+
+	[]
```
JeffAshton, line 45: This broke hard

JeffAshton: But fixed now

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.CSharp.Syntax.cs
```diff
@@ -33,5 +34,25 @@ internal static partial class RoslynExtensions {
 
 			return true;
 		}
+
```
JeffAshton, line 37: Going to use this in my next PR too.

src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/SerializerAttributeAnalyzer.cs
JeffAshton: Good feedback regardless.  Got me thinking about more `null` possibilities

tests/D2L.CodeStyle.Analyzers.Test/Specs/SerializerAttributeAnalyzer.cs
j3parker: It's always funky needing to handle broken code

PR #571, opened 2020-07-16
Adding support for ISerializer

JeffAshton:


src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/SerializerAttributeAnalyzer.cs
```diff
@@ -18,6 +19,7 @@ internal sealed class SerializerAttributeAnalyzer : DiagnosticAnalyzer {
 			);
 
 		private const string SerializerAttributeFullName = "D2L.LP.Serialization.SerializerAttribute";
+		private const string ISerializerFullName = "D2L.LP.Serialization.ISerializer";
```
JeffAshton, line 22: There is a 2nd interface used, `ITrySerializer` that gets wrapped into an `ITrySerializer`

https://search.d2l.dev/xref/lms/lp/framework/core/D2L.LP/LP/Serialization/Resolvers/Reflection/SerializerActivator.cs?r=9d051287#26

image

PR #589, opened 2020-10-28
remove Except support from [Immutable]

omsmith:


omsmith:
Oh oops. Updated to wrong remote. Thought some stuff was weird, lol

omsmith:
Fixed

j3parker (approved):
@mthjones -- we're looking to make some bigger changes to the immutability analyzer (allow for custom immutable container types and some misc fixes). Largely this is to allow it to integrate into static DI.

This feature wasn't used much and it's helpful to cut this stuff down a bit (e.g. we're making slight tweaks to how `[Immutable]` works and it would be weird with this feature).

RE: undiff, since we have so few `Unaudited` left Littlefoot has a good handle on which exist, what their role is etc., and people can't add new ones without us knowing... so in terms of making shipping safer I think this has mostly outlived its utility.

mthjones (approved):
Thanks for the context. Agreed that it's outlived its usefulness, and this much red to remove whats a pretty small feature seems like a good choice to me. :shipit:

PR #591, opened 2020-11-03
Add new diagnostics for immutability

j3parker:


src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -514,5 +514,77 @@ public static class Diagnostics {
 			isEnabledByDefault: true,
 			description: "We require that event types be sealed."
 		);
+
+		public static readonly DiagnosticDescriptor MutableBaseType = new DiagnosticDescriptor(
+			id: "D2L0061",
+			title: "Base class for immutable type must be [Immutable] (or [ImmutableBaseClass])",
+			messageFormat: "{0}'s base class is {1} which is missing [Immutable] (or, more weakly, [ImmutableBaseClass]).",
```
j3parker, line 521: "Foo's base class is Bar which is missing [Immutable] (or, more weakly, [ImmutableBaseClass])"

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -514,5 +514,77 @@ public static class Diagnostics {
...
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
+
+		public static readonly DiagnosticDescriptor ArraysAreMutable = new DiagnosticDescriptor(
+			id: "D2L0062",
+			title: "Arrays are mutable and thus can't be held by an immutable type.",
+			messageFormat: "{0}[]'s are mutable even if {0}'s are immutable because array items can be substituted. Prefer an immutable container type such as ImmutableArray<{0}>.",
```
j3parker, line 530: "Foo[]'s are mutable even if Foo's are immutable because array items can be substituted. Prefer an immutable container type such as ImmutableArray<Foo>."

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -514,5 +514,77 @@ public static class Diagnostics {
...
+			messageFormat: "Dynamic objects are mutable and thus can't be held by an immutable type.",
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
+
+		public static readonly DiagnosticDescriptor TypeParameterIsNotKnownToBeImmutable = new DiagnosticDescriptor(
+			id: "D2L0064",
+			title: "Type parameter is not known to be immutable.",
+			messageFormat: "The type parameter {0} is not known to be immutable. Add [Immutable] to it.",
```
j3parker, line 548: "The type parameter T is not known to be immutable. Add [Immutable] to it."

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -514,5 +514,77 @@ public static class Diagnostics {
...
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
+
+		public static readonly DiagnosticDescriptor UnexpectedTypeKind = new DiagnosticDescriptor(
+			id: "D2L0065",
+			title: "Unsupported type kind inside an immutable type declaration.",
+			messageFormat: "The type kind {0} is unexpected/unhandled by the immutability analyzer, and for safety we are emitting an error. This is a bug in the analyzer.",
```
j3parker, line 557: This "shouldn't happen" (until it does).

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -514,5 +514,77 @@ public static class Diagnostics {
...
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
+
+		public static readonly DiagnosticDescriptor TypeNeedsImmutableAttribute = new DiagnosticDescriptor(
+			id: "D2L0066",
+			title: "Type may be mutable.",
+			messageFormat: "The {0} {1} is missing the [Immutable]{2} attribute, so it isn't safe to be held by an immutable type",
```
j3parker, line 566: "The interface IFoo is missing the [Immutable] attribute, so it isn't safe to be held by an immutable type."
"The class Bar is missing the [Immutable] (or [ImmutableBaseClass]) attribute, so it isn't safe to be held by an immutable type."

Isn't there already an error for base classes? Yes. But the logic for `[ImmutableBaseClass]` is used for other stuff too (it's a very bad name). e.g.:

```
private static readonly IFoo m_foo = new Foo( ... );
```

In this case (modulo the #190 bug) things are allowed if `Foo` has `[ImmutableBaseClass]`.

(I'd like to change the attribute up a bit, but I'll leave that talk for the next PR (which doesn't touch the attribute but will better contextualize the discussion))

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
omsmith: I guess that's better than throwing.

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -514,5 +514,77 @@ public static class Diagnostics {
...
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
+
+		public static readonly DiagnosticDescriptor MemberIsNotReadOnly = new DiagnosticDescriptor(
+			id: "D2L0067",
+			title: "Member must be readonly inside immutable type.",
+			messageFormat: "The {0} {1} must be readonly because {2} is immutable.",
```
j3parker, line 575: "The field m_foo must be readonly because ParentType is immutable."

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -514,5 +514,77 @@ public static class Diagnostics {
...
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
+
+		public static readonly DiagnosticDescriptor UnexpectedMemberKind = new DiagnosticDescriptor(
+			id: "D2L0068",
+			title: "Unsupported member type in immutable type declaration.",
+			messageFormat: "The member {0} is of type {1} which is not supported by the immutability analyzer. This is a bug.",
```
j3parker, line 584: This "shouldn't happen" (until it does).

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
j3parker: Yeah. Right now we sometimes throw and sometime diagnostic. Diagnostic is nicer because in VS if your analyzer throws VS bans it (or offers to do so).

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -514,5 +514,77 @@ public static class Diagnostics {
...
+		public static readonly DiagnosticDescriptor UnexpectedTypeKind = new DiagnosticDescriptor(
+			id: "D2L0065",
+			title: "Unsupported type kind inside an immutable type declaration.",
+			messageFormat: "The type kind {0} is unexpected/unhandled by the immutability analyzer, and for safety we are emitting an error. This is a bug in the analyzer.",
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
+
+		public static readonly DiagnosticDescriptor TypeNeedsImmutableAttribute = new DiagnosticDescriptor(
```
omsmith, line 563: The naming of this implies to me it would be ~ D2L0040 MissingTransitiveImmutableAttribute vs what it is

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -514,5 +514,77 @@ public static class Diagnostics {
...
+		public static readonly DiagnosticDescriptor MutableBaseType = new DiagnosticDescriptor(
+			id: "D2L0061",
+			title: "Base class for immutable type must be [Immutable] (or [ImmutableBaseClass])",
+			messageFormat: "{0}'s base class is {1} which is missing [Immutable] (or, more weakly, [ImmutableBaseClass]).",
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
+
+		public static readonly DiagnosticDescriptor ArraysAreMutable = new DiagnosticDescriptor(
```
omsmith, line 527: Does this need a special diagnostic? Or are you calling it out separately because you can't `[Immutable]` so the language in `D2L0066` doesn't apply?

Does it need to be separate from `D2L0063` though? "This is a mutable thing that can't be made immutable, choose something else"

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
j3parker: Ah, these are two different cases:

1. If you need `[Immutable]` because you implement an `[Immutable]` interface (or etc.) then you should get D2L0040
2. If you are held onto a `Foo` and `Foo` isn't immutable, that's an error

Ideally we only show (1) and not (2) if applicable.

If you can think of a better name for the diagnostic var LMK.

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
j3parker: > Does this need a special diagnostic? Or are you calling it out separately because you can't [Immutable] so the language in D2L0066 doesn't apply?

It's a separate branch of code, but the reason you gave is good.

> Does it need to be separate from D2L0063 though? "This is a mutable thing that can't be made immutable, choose something else"

D2L0063 is about like `private readonly dynamic m_magic`.

j3parker: (The error messages don't _have_ to be this granular (they're very much not right now) but the steps to fix are very different so I think it's worth it.)

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
omsmith: Given the way we're going with the container analyzer, the wording on this seems off to me.

> Type parameter is not known to be immutable.

This implies to me that if I add `<T>` -> `<[Immutable] T>`, then `T` is guaranteed to be immutable. This aligns with current behaviour of immutable generics.

> Type parameter is not considered to be immutable.

Splitting hairs, but the slightly different wording I think alludes to the real situation a bit better. "This may or may not be immutable, but if you want it to be considered to be immutable for the purpose of analyzing your class declaration, add `[Immutable]` to it."

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
omsmith: Yeah for sure. Maybe `MutableTypeHeldByImmutableType`

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
omsmith: I guess since you're suggesting `ImmutableArray`, k

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
j3parker: Done

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
j3parker: > This aligns with current behaviour of immutable generics.

Yeah, I was kinda trying to not step on your toes :)

Agreed the wording is not fantastic. Want to fix this when you do the container stuff?

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
mthjones: Consider telling users what they can do if they experience this bug (e.g. file an issue in this repo). [`DiagnosticDescriptor.HelpLinkUri`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.diagnosticdescriptor.helplinkuri) may be useful here, even if its just a link to the repo itself.

mthjones (dismissed):

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
j3parker: Cool -- I deeplinked to the create issue page with a reasonable title for these two diagnostics.

omsmith (dismissed):

j3parker:
(fixed a typo: `helpLinkuri` -> `helpLinkUri`)

mthjones (approved):

PR #592, opened 2020-11-09
Add support for querying if a type is known to be immutable.

j3parker:
This organizes the logic to ask "do we know if this type is immutable?".
This is a different question than "is it OK for this declaration to have
[Immutable]?", or "does this declaration need to have [Immutable]?".

This is missing support for "generic container types" like
ImmutableArray<T>, where the type is conditionally immutable.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
@@ -1,0 +1,182 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+
+namespace D2L.CodeStyle.Analyzers.Immutability {
+	/// <summary>
+	/// Used to answer "do we know if this type immutable?" (in whichever sense).
+	/// Generally this means checking for [Immutable] etc., but there are also
+	/// some pre-approved types (like int) and other logic that this class
+	/// organizes.
+	///
+	/// This doesn't do anything particularly smart: e.g. for a user defined
+	/// class it is mostly just looking for [Immutable]. If this judges a type
+	/// as not known to be immutable that doesn't mean it is mutable.
+	/// </summary>
+	internal sealed class ImmutabilityContext {
+		private readonly ImmutableDictionary<ITypeSymbol, ImmutableTypeInfo> m_extraImmutableTypes = null;
+
+		// Hard code this to avoid looking up the ITypeSymbol to include it in m_extraImmutableTypes
+		private static readonly ImmutableHashSet<SpecialType> m_totallyImmutableSpecialTypes = ImmutableHashSet.Create(
+			SpecialType.System_Enum,
+			SpecialType.System_Boolean,
+			SpecialType.System_Char,
+			SpecialType.System_SByte,
+			SpecialType.System_Byte,
+			SpecialType.System_Int16,
+			SpecialType.System_UInt16,
+			SpecialType.System_Int32,
+			SpecialType.System_UInt32,
+			SpecialType.System_Int64,
+			SpecialType.System_UInt64,
+			SpecialType.System_Decimal,
+			SpecialType.System_Single,
+			SpecialType.System_Double,
+			SpecialType.System_String,
+			SpecialType.System_IntPtr,
+			SpecialType.System_UIntPtr,
+			SpecialType.System_ValueType
+		);
+
+		public ImmutabilityContext( IEnumerable<ImmutableTypeInfo> extraImmutableTypes ) {
+			m_extraImmutableTypes = extraImmutableTypes
+				.ToImmutableDictionary(
+					info => info.Type,
+					info => info
+				);
+		}
+
+		/// <summary>
+		/// Determines if a type is known to be immutable.
+		/// </summary>
+		/// <param name="type">The type to check</param>
+		/// <param name="kind">The degree of immutability</param>
+		/// <param name="diag">If this method returns false, an explaination for why its not known to be immutable.</param>
+		/// <returns>Is the type immutable?</returns>
+		public bool IsImmutable(
+			ITypeSymbol type,
+			ImmutableTypeKind kind,
+			Location location,
+			out Diagnostic diagnostic
+		) {
+			if ( kind == ImmutableTypeKind.None ) {
+				throw new ArgumentException(
+					"ImmutabilityKind.None is not a valid question to ask this function",
+					nameof( kind )
+				);
+			}
+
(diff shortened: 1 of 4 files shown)


j3parker:
(gotta write some tests)

j3parker:
> (gotta write some tests)

We're now planning on getting stuff in and iterating (without it running etc.)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableTypeKind.cs
```diff
@@ -0,0 +1,28 @@
...
+		/// an immutable value.
+		/// </summary>
+		Instance = 1,
+
+
+		/// <summary>
+		/// A type T is totally immutable if all objects of type T are
+		/// immutable. This is a stronger condition than instance immutability.
+		/// </summary>
+		Total = 2 | Instance,
```
j3parker, line 26: This+`ImmutableTypeInfo` is the stuff I care most about pinning down: we have a shocking amount of reimplementation in the immutability stuff due to how it was developed over time. Stuff like `[ImmutableGeneric]` and the many ways that we "bless" types.

The comments in here I think capture what we're trying to do. To make all this happen we need:

* A way to tell if a type is known to be immutable in one of these two ways (`ImmutabilityContext` -- it is context dependent, e.g. based on type arguments, and when you're analyzing `Foo`s decl you assume that `Foo` is immutable, etc.)
* A way to validate that an `[Immutable]` decl is valid (looks at fields, makes sure they're readonly, etc.... `ImmutableDeclarationChecker` in next PR)
* If a type is `[Immutable]` (in this `Total` sense) then any subclass/implementation also needs to be `[Immutable]`. We already have an analyzer for this (it was done awhile ago with this refactoring in mind)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableTypeInfo.cs
```diff
@@ -0,0 +1,35 @@
...
+using Microsoft.CodeAnalysis;
+
+namespace D2L.CodeStyle.Analyzers.Immutability {
+	/// <summary>
+	/// A representation of the immutability of some type.
+	/// </summary>
+	internal readonly struct ImmutableTypeInfo : IEquatable<ImmutableTypeInfo> {
+		public ImmutableTypeInfo(
+			ImmutableTypeKind kind,
+			ITypeSymbol type
```
j3parker, line 11: To support generic container classes like `ImmtuableArray<T>` (either "blessed" or user-defined) this needs to have a list of `ITypeParameterSymbol`s that the immutability is conditional upon.

There won't be support for saying `X<A,B>` is `ImmutableTypeKind.Total` if `A`&`B` are `ImmutableKind.Total`, but `ImmutableTypeKind.Instance` if only `A` is, etc.... in other words we're only going to bother having at most one `ImmutableTypeInfo` judgement per `ITypeSymbol`.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -0,0 +1,178 @@
...
+						Diagnostics.ArraysAreMutable,
+						location,
+						type.Name
+					);
+
+					return false;
+
+				case TypeKind.Delegate:
+					// Delegate members just give a name to a function type.
+					return true;
```
j3parker, line 119: This is a difference from the current analyzer. Seems fine to me?

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableTypeInfo.cs
```diff
@@ -0,0 +1,35 @@
...
+		) {
+			Kind = kind;
+			Type = type;
+		}
+
+		public ImmutableTypeKind Kind { get; }
+
+		public ITypeSymbol Type { get; }
+
+		public bool Equals( ImmutableTypeInfo other ) => this.Equals( other );
```
j3parker, line 21: Uh this stuff is probably not needed. Initially I had a hashset of these types but now its `Dictionary<ITypeSymbol, ImmutableTypeInfo>`.

(`ISymbol` is `IEquatable<ISymbol?>`)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
mthjones: Would this change allow something like this to pass the analyzer?

```cs
public delegate string SillyDelegate( string key );

[Immutable]
public class Foo {
    public SillyDelegate m_func;

    public Foo( Dictionary<string, string> sillyCache, IConfigViewer viewer ) {
        m_func = key => {
			if( !sillyCache.TryGetValue( key, out string value ) ) {
				sillyCache[key] = viewer.GetValue( key );
			}
			return value;
        };
    }

	public string GetValue( string key ) => m_func( key );
}
```

j3parker: Yes, this is wrong:

```cs
const string src = @"
	class X {
		delegate void Y();
		public Y m_y;
		event Y m_z;
	}";

var tree = CSharpSyntaxTree.ParseText( src );
var compilation = CSharpCompilation.Create( "assembly", new [] { tree } );
var model = compilation.GetSemanticModel( tree );
var decl = tree.GetRoot().ChildNodes().First();
var xtype = (INamedTypeSymbol)model.GetDeclaredSymbol( decl );
var members = xtype.GetMembers();

foreach( var member in members ) {
	var memberTypeInfo =
		member is IPropertySymbol prop ? $" of type {prop.Type.Name} which is a {prop.Type.TypeKind}"
		: member is IFieldSymbol field ? $" of type {field.Type.Name} which is a {field.Type.TypeKind}"
		: null;

	Console.WriteLine( $"{member.Name} is a {member.Kind}{memberTypeInfo} (implicit: {member.IsImplicitlyDeclared})" );
}
```

outputs:

```
Y is a NamedType (implicit: False)
m_y is a Field of type Y which is a Delegate (implicit: False)
add_m_z is a Method (implicit: True)
remove_m_z is a Method (implicit: True)
m_z is a Event (implicit: False)
.ctor is a Method (implicit: True)
```

This code will get used to look at a fields type, and the only time we see TypeKind == delegate is when we have a field/property that uses a delegate, not when a delegate is declared (that one is just a NamedType member)... I got mixed up.

j3parker: Fixed; thanks.

mthjones (approved):

omsmith (approved):

PR #594, opened 2020-11-11
amalgamate checking of attributes and extra types

omsmith:


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -165,6 +160,29 @@ internal sealed class ImmutabilityContext {
 			}
 		}
 
+		private bool TryGetImmutableTypeInfo(
```
omsmith, line 163: Slight difference, in that `m_extraImmutableTypes` isn't considered if a type already has `[Immutable*]` attributes

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -165,6 +160,29 @@ internal sealed class ImmutabilityContext {
...
+			if( fromAttributes != ImmutableTypeKind.None ) {
+				info = new ImmutableTypeInfo(
+					kind: fromAttributes,
+					type: type
+				);
+				return true;
+			}
+
+			// Check if we were otherwise told that this type is immutable
+			if( m_extraImmutableTypes.TryGetValue( type, out info ) ) {
```
omsmith, line 178: Should probably eventually look at `OriginalDefinition` as well, but will save making such adjustments for later

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -165,6 +160,29 @@ internal sealed class ImmutabilityContext {
...
+				return true;
+			}
+
+			info = default;
+			return false;
```
j3parker, line 183:
```suggestion
			return m_extraImmutableTypes.TryGetValue( type, out info );
```

I think?

j3parker (dismissed):
Nice

j3parker:
@mthjones this is to support immutable container types -- we want to ask "how" a type is immutable and then check if that's sufficiently immutable for us. It shouldn't matter if its done with attributes or done via `m_extraImmutableTypes`; `ImmutableTypeInfo` abstracts this away.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -84,15 +84,10 @@ internal sealed class ImmutabilityContext {
...
-
-			// Check if we were otherwise told that this type is (sufficiently)
-			// immutable.
-			if ( m_extraImmutableTypes.TryGetValue( type, out var info ) && info.Kind.HasFlag( kind ) ) {
-				return true;
+			if( TryGetImmutableTypeInfo( type, out ImmutableTypeInfo info ) ) {
+				if( info.Kind.HasFlag( kind ) ) {
+					return true;
+				}
 			}
```
j3parker, line 91: @mthjones this is to support immutable container types better -- we want to ask "how" a type is immutable and then check if that's sufficiently immutable for us. It shouldn't matter if its done with attributes or done via m_extraImmutableTypes; ImmutableTypeInfo abstracts this away.

mthjones (dismissed):

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
omsmith: Yeah, just depends on your feelings about breakpoints

j3parker (approved):

PR #595, opened 2020-11-11
only consider a type immutable if every type argument for an [Immutab…

omsmith:
…le] parameter is immutable

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
@@ -88,25 +88,12 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			if( type is INamedTypeSymbol namedType ) {
 				if( TryGetImmutableTypeInfo( namedType, out ImmutableTypeInfo info ) ) {
 					if( info.Kind.HasFlag( kind ) ) {
-						if( namedType.TypeArguments.Length != info.ImmutableTypeParameters.Length ) {
-							// could this happen?
-							throw new Exception();
-						}
-
-						IEnumerable<(ITypeSymbol, bool)> pairs = namedType
-							.TypeArguments
-							.Zip( info.ImmutableTypeParameters, ( a, relevant ) => (a, relevant) );
-						foreach( (ITypeSymbol argument, bool isRelevant) in pairs ) {
-							if( !isRelevant ) {
-								continue;
-							}
-
-							if( !IsImmutable( argument, ImmutableTypeKind.Total, location, out diagnostic ) ) {
-								return false;
-							}
-						}
-
-						return true;
+						return info.IsImmutableDefinition(
+							context: this,
+							definition: namedType,
+							location: location,
+							out diagnostic
+						);
 					}
 				}
 			}
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Linq;
 using Microsoft.CodeAnalysis;
 
 namespace D2L.CodeStyle.Analyzers.Immutability {
@@ -15,7 +16,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 	/// as not known to be immutable that doesn't mean it is mutable.
 	/// </summary>
 	internal sealed class ImmutabilityContext {
-		private readonly ImmutableDictionary<ITypeSymbol, ImmutableTypeInfo> m_extraImmutableTypes = null;
+		private readonly ImmutableDictionary<INamedTypeSymbol, ImmutableTypeInfo> m_extraImmutableTypes = null;
 
 		// Hard code this to avoid looking up the ITypeSymbol to include it in m_extraImmutableTypes
 		private static readonly ImmutableHashSet<SpecialType> m_totallyImmutableSpecialTypes = ImmutableHashSet.Create(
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -84,9 +85,29 @@ internal sealed class ImmutabilityContext {
 				return true;
 			}
 
-			if( TryGetImmutableTypeInfo( type, out ImmutableTypeInfo info ) ) {
-				if( info.Kind.HasFlag( kind ) ) {
-					return true;
+			if( type is INamedTypeSymbol namedType ) {
```
omsmith, line 88: I don't thiiiink this is unreasonable. But tests will confirm down the road (this might break analyzing the declaration of an immutable container, when trying to store `T`)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -84,9 +85,29 @@ internal sealed class ImmutabilityContext {
...
 			}
 
-			if( TryGetImmutableTypeInfo( type, out ImmutableTypeInfo info ) ) {
-				if( info.Kind.HasFlag( kind ) ) {
-					return true;
+			if( type is INamedTypeSymbol namedType ) {
+				if( TryGetImmutableTypeInfo( namedType, out ImmutableTypeInfo info ) ) {
+					if( info.Kind.HasFlag( kind ) ) {
+						if( namedType.TypeArguments.Length != info.ImmutableTypeParameters.Length ) {
+							// could this happen?
```
omsmith, line 92: 🤷

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableTypeInfo.cs
```diff
@@ -1,34 +1,76 @@
...
+				.ToImmutableArray();
+
+			return new ImmutableTypeInfo(
+				kind: kind,
+				type: type,
+				immutableTypeParameters: immutableTypeParameters
+			);
+		}
+
+		public static ImmutableTypeInfo CreateWithAllImmutableTypeParameters(
```
omsmith, line 44: Imagine this would be used when loading in the "extra types". i.e we won't support adding an exception equivalent to `MyType<A, [Immutable] B>`

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -84,9 +85,29 @@ internal sealed class ImmutabilityContext {
...
+							throw new Exception();
+						}
+
+						IEnumerable<(ITypeSymbol, bool)> pairs = namedType
+							.TypeArguments
+							.Zip( info.ImmutableTypeParameters, ( a, relevant ) => (a, relevant) );
+						foreach( (ITypeSymbol argument, bool isRelevant) in pairs ) {
+							if( !isRelevant ) {
```
omsmith, line 100: For a "container" type

```csharp
[Immutable] class MyType<[Immutable] A, B> {}
```

Which might get used like

```csharp
[Immutable]
class Holder {
  readonly MyType<int, object> m_held;
}
```

Only `A`/`int` is relevant when considering where it is immutable (this generally implies `B`/`object` isn't "stored").


Also note that for the current analyzers, `MyType<object, object>` is not allowed since `<[Immutable] A` means `A` is _required_ to be immutable. For the new analyzers, `MyType<object, object>` is just not considered to be immutable, where as `MyType<int, object>` is.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
j3parker: I don't think so, `throw new Exception()` is fine.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableTypeInfo.cs
```diff
@@ -1,34 +1,76 @@
...
 	/// A representation of the immutability of some type.
 	/// </summary>
 	internal readonly struct ImmutableTypeInfo : IEquatable<ImmutableTypeInfo> {
-		public ImmutableTypeInfo(
+
+		private ImmutableTypeInfo(
 			ImmutableTypeKind kind,
-			ITypeSymbol type
+			INamedTypeSymbol type,
+			ImmutableArray<bool> immutableTypeParameters
```
j3parker, line 15: Maybe add a comment somewhere to explain how this works? I assume `[Immutable]` -> `true` here?

omsmith: Yes from https://github.com/Brightspace/D2L.CodeStyle/pull/595/files#diff-e61de6c1c372a9fe62f866659862da11c2f471672e110100560ab084e4183ccdR34

doc comment on `public ImmutableArray<bool> ImmutableTypeParameters { get; }`?

j3parker: 🤷 dno

j3parker: Hmmm.... I wonder if some of the logic could move into `ImmutableTypeInfo` and more of it could just become private?

That makes me thing.... `m_extraImmutableTypes`... the keys, should they be the generic type instead?

j3parker:
```cs
public sealed class ImmutableTypeInfo {
  INamedType Type { get; }
  ImmutableType Kind { get; }

  private ImmutableTypeInfo( ... ) {
  }

  // your two static helper Create- methods

  public bool IsImmutable( ImmutabilityContext ctx, INamedType other ) {
    // say we are ImmutableArray<[Immutable]T>
    // other is ImmutableArray<int>
    // we need to check that T=int is immutable
  }
}
```

that's what I mean

omsmith: > That makes me thing.... `m_extraImmutableTypes`... the keys, should they be the generic type instead?

I think we'll need both. We want `ImmutableArray<T>` to be considered immutable in general and so yeah we'd use the OriginalDefinition as a key, but maybe also `IFactory<IMyFoo>` (for replacing `ImmutableGeneric`)?

j3parker: That way there is no need to document stuff. `ImmutableTypeInfo.IsImmutable`/`ImmutabilityContext.IsImmutable` would be co-recursive to support `ImmutableArray<ImmutableArray<int>>` etc.

j3parker: Ah `OriginalDefinition`, ok now I get your other comment.

omsmith: Yeah, only concern with the co-recursiveness for me would be splitting the logic up too terribly, but some of it is via the `Create` and knowing how to use the properties anyway, so :shrug: seems good

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableTypeInfo.cs
omsmith: I've successfully used this for value tuples.

j3parker (approved):

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableTypeInfo.cs
j3parker: Oh nice, value tuple support would be good.

PR #597, opened 2020-11-12
support value tuples in ImmutabilityContext

omsmith:
It's impractical and/or impossible (I forget which) to refer to value tuples by name and thus they wouldn't do well as an "extra type" style thing.

Add specific logic, filling them in as "all types must be immutable"

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
@@ -185,7 +185,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			if( type.IsTupleType ) {
 				info = ImmutableTypeInfo.CreateWithAllImmutableTypeParameters(
 					kind: ImmutableTypeKind.Total,
-					type: type
+					type: type.OriginalDefinition
 				);
 				return true;
 			}
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
@@ -182,6 +182,14 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				return true;
 			}
 
+			if( type.IsTupleType ) {
+				info = ImmutableTypeInfo.CreateWithAllImmutableTypeParameters(
+					kind: ImmutableTypeKind.Total,
+					type: type
+				);
+				return true;
+			}
+
 			// Check if we were otherwise told that this type is immutable
 			return m_extraImmutableTypes.TryGetValue( type, out info );
 		}


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -182,6 +182,14 @@ out diagnostic
 				return true;
 			}
 
+			if( type.IsTupleType ) {
+				info = ImmutableTypeInfo.CreateWithAllImmutableTypeParameters(
+					kind: ImmutableTypeKind.Total,
```
omsmith, line 187: There could theoretically be some funkiness in this zone given `readonly (object, int) m_foo = (new object(), 1);` is theoretically okay...

Decide if we want to care/support that down the road?

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -182,6 +182,14 @@ out diagnostic
 				return true;
 			}
 
+			if( type.IsTupleType ) {
+				info = ImmutableTypeInfo.CreateWithAllImmutableTypeParameters(
+					kind: ImmutableTypeKind.Total,
+					type: type
```
j3parker, line 188: Is it "weird" that we don't do `.OriginalDefinition` here? It works regardless, though?

omsmith: Doesn't really matter. This is basically JIT adding `[Immutable]` attributes to the type definition.

j3parker: Yeah, it's just interesting to me that type could be like, `ValueType<object>`, no?

j3parker: `[Immutable]ValueType<[Immutable]T1>`, `[Immutable]ValueType<[Immutable]T1, [Immutable]T2>`... are things I'd expect to express with an `ImmutableTypeInfo` for sure.

omsmith: I can switch it to OrignialDefinition, I've no strong feels.

```csharp
			readonly (ClassMarkedImmutableImplementingRegularInterface) m_bar = default;
			readonly (object) /* NonImmutableTypeHeldByImmutable(Class, Object, ) */ m_foo /**/ = default;
```

```
type.Name
"ValueTuple"
type.TypeParameters
Length = 2
    [0]: {T1}
    [1]: {T2}
    Results View: Expanding the Results View will enumerate the IEnumerable
type.TypeArguments
Length = 2
    [0]: {SpecTests.Classes.ClassMarkedImmutableImplementingRegularInterface}
    [1]: {?}
    Results View: Expanding the Results View will enumerate the IEnumerable
type.Name
"ValueTuple"
type.TypeParameters
Length = 2
    [0]: {T1}
    [1]: {T2}
    Results View: Expanding the Results View will enumerate the IEnumerable
type.TypeArguments
Length = 2
    [0]: {object}
    [1]: {?}
    Results View: Expanding the Results View will enumerate the IEnumerable
type.ToMinimalDisplayString()
error CS1501: No overload for method 'ToMinimalDisplayString' takes 0 arguments
type.ToDisplayString()
"(object, ?)"
type.OriginalDefinition.ToDisplayString()
"(T1, T2)"
```

omsmith: The `?` (ErrorType) is because I suppose single-valued tuples aren't valid, need a minimum of two :P

Still works though!

j3parker: Yeah, I don't feel strongly about it and suspected it would still work. I'm just "surprised" to see it not used, even though in practice it doesn't matter.

It might matter more if we want to support what `[ImmutableGeneric]` is doing, where we say "I know `IFactory<T>` in general is mutable, but I want any `IFactory<Foo>` to be immutable". I don't think the logic of that stuff is 100% sound though, like when you have `Bar<X>` implementing `IFactory<X>` then `Bar<Foo>` would also need to be immutable too and I'm not sure we currently do that right (so I'm happy to stop supporting that feature)

j3parker (approved):

PR #601, opened 2020-11-12
Add support for initializers in ImmutableDefinitionChecker

j3parker:


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -190,18 +191,87 @@ ExpressionSyntax initializer
...
 
-			// TODO: narrow inspection with initializer
-			//       possibly a narrower type (using initializers type) and
-			//       possibly ImmutableTypeKind.Instances if its a new T( ... ).
+			var stuff = GetStuffToCheckForMember( type, typeSyntax, initializer );
 
-			if( !m_context.IsImmutable( type, ImmutableTypeKind.Total, typeSyntax.GetLocation(), out var diagnostic ) ) {
+			// null is a signal that there is nothing fruther that needs to be checked
+			if ( stuff == null ) {
+				return immutable;
```
j3parker, line 198: It is important that this isn't `return true;` because of the `readonly` check above.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -190,18 +191,84 @@ ExpressionSyntax initializer
...
+
+			if( !m_context.IsImmutable( typeToCheck, checkKind, diagnosticLocation, out var diagnostic ) ) {
 				diagnosticSink( diagnostic );
 				immutable = false;
 			}
 
 			return immutable;
 		}
 
+		/// <summary>
```
j3parker, line 211: This stuff is also basically copy-and-paste from the MutabilityInspector, of course.

omsmith (approved):
I think I like it

PR #603, opened 2020-11-13
place the diagnostic on the member type if both member and initialize…

omsmith:
…r arent immutable

This is pretty roundabout... 😞 

```csharp
sealed class SomeClass {}

[Immutable]
sealed class AnalyzedClass {
  readonly SomeClass m_foo = new SomeClass();
}
```

Current behaviour:
```
readonly SomeClass m_foo = new SomeClass();
                               ~~~~~~~~~
// The Class SomeClass is missing the [Immutable] (or [ImmutableBaseClass]) attribute, so it isn't safe to be held by an immutable type
```

New behaviour:
```
readonly SomeClass m_foo = new SomeClass();
         ~~~~~~~~~
// The Class SomeClass is missing the [Immutable] attribute, so it isn't safe to be held by an immutable type
```

This is maybe pretty preferential though, so I could ignore it for the sake of the less complicated code.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
@@ -209,7 +209,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				if( !m_context.IsImmutable( typeToCheck, checkKind, diagnosticLocation, out var diagnostic ) ) {
 					immutable = false;
 
-					// depending on the situation, it may be preferable to place a diagnostic on the held member type instead
+					// if the initializer _and_ the member type aren't immutable, prefer the diagnostic on the member type
 					if( !m_context.IsImmutable( type, ImmutableTypeKind.Instance, typeSyntax.GetLocation(), out _ ) ) {
 						// This returning false is implied by not being Instance immutable
 						// Running this is mostly just a hack to get a diagnostic without the " (or [ImmutableBaseClass])" though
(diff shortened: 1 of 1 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -191,18 +191,34 @@ ExpressionSyntax initializer
...
-			var (typeToCheck, checkKind, diagnosticLocation) = stuff.Value;
+				var (typeToCheck, checkKind, diagnosticLocation) = stuff.Value;
 
-			if( !m_context.IsImmutable( typeToCheck, checkKind, diagnosticLocation, out var diagnostic ) ) {
-				diagnosticSink( diagnostic );
-				immutable = false;
+				if( !m_context.IsImmutable( typeToCheck, checkKind, diagnosticLocation, out var diagnostic ) ) {
+					immutable = false;
+
+					// depending on the situation, it may be preferable to place a diagnostic on the held member type instead
```
omsmith, line 212:
```suggestion
					// if the initializer _and_ the member type aren't immutable, prefer the diagnostic on the member type
```

j3parker:
> This is maybe pretty preferential though, so I could ignore it for the sake of the less complicated code.

Yeah uh, this seems like maybe not a good trade. 

Honestly once we do fancier stuff with constructors we’re probably going to want to go back to just putting the diagnostic on the field type and pretending the initializer logic doesn’t exist (like we talked about). So maybe let’s not worry about this right yet.

j3parker:
Let's hold off on this because it will change when #319 is fixed anyway.

PR #607, opened 2020-11-13
look for OriginalDefinition in ExtraTypes

omsmith:


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -201,7 +201,7 @@ out diagnostic
 			}
 
 			// Check if we were otherwise told that this type is immutable
-			return m_extraImmutableTypes.TryGetValue( type, out info );
+			return m_extraImmutableTypes.TryGetValue( type.OriginalDefinition, out info );
```
omsmith, line 204: As stated before, we may eventually want to look for both, but this is good for current cases.

j3parker (approved):

PR #610, opened 2020-11-16
Place mutable base class diagnostic better

j3parker:


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -285,5 +282,49 @@ ExpressionSyntax initializer
...
+			// Our goal is to get one of the mentions of Base after "Foo :" to
+			// place the diagnostic. There may be multiple, but for valid code
+			// it will be the first type in the list for one of the decls.
+			// (interfaces always go after the base class, if it is listed.)
+
+			var candidates = type.DeclaringSyntaxReferences
+				.Select( r => r.GetSyntax() )
+				.Cast<ClassDeclarationSyntax>()
+				// Take _at most_ the first item from each BaseList.Types
+				.SelectMany( r => r.BaseList.Types.Take( 1 ) );
```
omsmith, line 307: Even though ` Foo: IA, Base` won't compile, our analyzer may still run?

🤷

omsmith (approved):

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Yep, so my thought is that the semantic model already told us that our base type is something other than `System.Object`, so it seems likely that `Foo : IA, Base` either isn't there or there is another `Foo : Base` decl.

That may not be true, so I added a fallback.

PR #611, opened 2020-11-16
Remove unnecessary check

j3parker:
Need to validate this first.


omsmith:
I agree it should be unnecessary given object is already special cased for instance.

j3parker:
Oh yeah, this is pretty obviously safe.

omsmith (approved):

PR #614, opened 2020-11-17
check immutability annotation of type parameters

omsmith:
The "above" this referred to was updated to only check named types

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
@@ -134,7 +134,10 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 					return false;
 
 				case TypeKind.TypeParameter:
-					// We already checked m_extraImmutableTypes above.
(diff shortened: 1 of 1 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -134,7 +134,10 @@ out diagnostic
 					return false;
 
 				case TypeKind.TypeParameter:
-					// We already checked m_extraImmutableTypes above.
+					if( GetImmutabilityFromAttributes( type ).HasFlag( ImmutableTypeKind.Total ) ) {
```
omsmith, line 137: Could also do `Attributes.Objects.Immutable.IsDefined( type )`

j3parker: I’m fine either way. Originally I was thinking about putting ImmutableTypeKind into annotations and eventually adding an optional arg to `[Immutable]` that defaulted to Total because I think “ImmutableBaseClass” is a misnomer. I didn’t do this because the analyzer DLL can’t reference annotations... but we could define an equivalent enum...

Anyway, either way is fine with me for now :P

j3parker (approved):

PR #621, opened 2020-11-18
Omsmith/definition analyzer

omsmith:


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -112,7 +112,7 @@ out diagnostic
 					diagnostic = Diagnostic.Create(
 						Diagnostics.ArraysAreMutable,
 						location,
-						type.Name
+						( type as IArrayTypeSymbol ).ElementType.Name
```
j3parker, line 115: Rebase?

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -0,0 +1,82 @@
...
+			context.EnableConcurrentExecution();
+			context.RegisterCompilationStartAction( CompilationStart );
+		}
+
+		public static void CompilationStart(
+			CompilationStartAnalysisContext context
+		) {
+			ImmutabilityContext immutabilityContext = ImmutabilityContext.Create( context.Compilation );
+
+			context.RegisterSymbolAction(
```
j3parker, line 32: Does this only run once if we have two decls for a partial class?

j3parker: (If it runs twice and emits the same diagnostics I wouldn't be surprised if they were de-duped, so check carefully)

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -0,0 +1,82 @@
...
+		}
+
+		private static bool ShouldAnalyze(
+			INamedTypeSymbol analyzedType
+		) {
+			if( analyzedType.IsImplicitlyDeclared ) {
+				return false;
+			}
+
+			if( analyzedType.TypeKind == TypeKind.Interface ) {
```
j3parker, line 67: So we actually only really know how to handle `class` and `struct`... so I'm guessing this will try to check a `record` and then crash inside the checker, which is good I think.

Just thinking out-loud.

tests/D2L.CodeStyle.Analyzers.Test/Specs/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -0,0 +1,598 @@
+﻿// analyzer: D2L.CodeStyle.Analyzers.Immutability.TypeDeclarationImmutabilityAnalyzer
```
j3parker, line 1: Taking a little extra time to look at these; writing some notes by hand about what I think needs to be tested and comparing.

I find these tests good but typically "write/append-only", so this is the only way I can think to review it.

omsmith: Yeah. Some of the groupings are a little inconsistent, largely due to impatience and evolution over time, but the idea was to just try to exercise all the patterns for each "type", instead of assuming/knowing that read-only field checks are done top-level and only need to be tested once, for instance.

It's annoying, but trying to blackbox it.

Was theoretically going to copy-paste them to be in a `struct` and an `[ImmutableBaseClass]` as well 😧

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
omsmith: Can double check, but no I don't think so. Symbols are a different layer than syntax, so I would be very surprised.

tests/D2L.CodeStyle.Analyzers.Test/Specs/TypeDeclarationImmutabilityAnalyzer.cs
omsmith: Need to add an auto-implemented `set` for properties.

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
j3parker: Yeah, the docs say:

> Register an action to be executed at completion of semantic analysis of an ISymbol with an appropriate Kind. A symbol action reports Diagnostics about ISymbols.

but I don't know enough to say for sure based on that, personally. I know semantic models are somewhat tied to documents and/or SyntaxTrees which is where my concern comes from.

j3parker: Like worst case we live with it or we add a cache, whatever.

j3parker (approved):
Ok -- still reviewing the tests, but I don't need to block the PR merge for that.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
omsmith: Done. Had opened from my phone. Was hoping it just wouldn't show.

j3parker: Yeah, I wish GitHub hadn't switched how they calculated diffs :(

PR #622, opened 2020-11-19
defer computing location of immutable type diagnostics

omsmith:
avoid paying for GetLocationOfBaseClass

| Function Name | Time, ms | Avg Time, ms | Own Time, ms | Calls |
| --- | --- | --- | --- | --- |
| D2L.CodeStyle.Analyzers.Immutability.ImmutableDefinitionChecker.GetLocationOfBaseClass(ITypeSymbol) | 226 | 21 | 8 | 11 |

(dotTrace when running current spec tests)

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
@@ -58,7 +58,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		public bool IsImmutable(
 			ITypeSymbol type,
 			ImmutableTypeKind kind,
-			Location location,
+			Func<Location> locationGetter,
 			out Diagnostic diagnostic
 		) {
 			if ( kind == ImmutableTypeKind.None ) {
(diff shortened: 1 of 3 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
```diff
@@ -58,7 +58,7 @@ internal sealed partial class ImmutabilityContext {
 		public bool IsImmutable(
 			ITypeSymbol type,
 			ImmutableTypeKind kind,
-			Location location,
+			Func<Location> locationGetter,
```
j3parker, line 61: nit: getLocation to match `GetLocation` + the active voice for function names (IsImmutable, CheckDefinition, etc.)

j3parker (approved):
cool

j3parker (approved):

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.cs
j3parker: It's not a method, but I try (but sometimes probably fail) to name my funcs as if they were: https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-type-members#names-of-methods

PR #626, opened 2020-11-19
analyze immutable method typeargs

omsmith:


src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -57,26 +80,33 @@ INamedTypeSymbol typeSymbol
 			checker.CheckDeclaration( typeSymbol );
 		}
 
-		private static bool ShouldAnalyze(
-			INamedTypeSymbol analyzedType
+		private static void AnalyzeGenericMethodTypeArguments(
```
omsmith, line 83: Not sure where the logic in this should live in the long run.

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -57,26 +80,33 @@ INamedTypeSymbol typeSymbol
...
 
-			if( Attributes.Objects.ImmutableBaseClass.IsDefined( analyzedType ) ) {
-				return true;
+				if( !immutabilityContext.IsImmutable(
+					type: argument,
+					kind: ImmutableTypeKind.Total,
+					getLocation: () => syntax.TypeArgumentList.Arguments[position].GetLocation(),
+					out Diagnostic diagnostic
+				) ) {
+					// TODO: not necessarily a good diagnostic for this use-case
```
omsmith, line 106: This is maybe okay for type parameters (as demonstrated in the spec test) but won't be for actual types (nothing is being held by an immutable type)

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -100,26 +122,33 @@ INamedTypeSymbol typeSymbol
...
+			int i = 0;
+			var paramArgPairs = method.TypeParameters.Zip( method.TypeArguments, ( p, a ) => (p, a, i++) );
+			foreach( var (parameter, argument, position) in paramArgPairs ) {
+				if( !Attributes.Objects.Immutable.IsDefined( parameter ) ) {
+					continue;
+				}
+
+				if( !immutabilityContext.IsImmutable(
+					type: argument,
+					kind: ImmutableTypeKind.Total,
```
omsmith, line 144: could over time change this to check the value being passed to an argument (assuming the type arg is used for an argument) similar to checking initializers of members.

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -100,26 +122,33 @@ INamedTypeSymbol typeSymbol
...
-				return true;
+			SymbolInfo info = ctx.SemanticModel.GetSymbolInfo( syntax, ctx.CancellationToken );
+			if( !( info.Symbol is IMethodSymbol method ) ) {
+				return;
 			}
 
-			if( Attributes.Objects.ImmutableBaseClass.IsDefined( analyzedType ) ) {
-				return true;
+			int i = 0;
+			var paramArgPairs = method.TypeParameters.Zip( method.TypeArguments, ( p, a ) => (p, a, i++) );
```
j3parker, line 136: There is an overload to `.Select` which gives you an index but 🤷

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
j3parker: Interesting. Open an issue? (Will forget about this personally)

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
j3parker: Yeah I think we'll need to do another pass generally...

j3parker (approved):

PR #627, opened 2020-11-20
Add checks for static vars to TypeDeclarationImmutabilityAnalyzer

j3parker:


src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -37,6 +37,38 @@ CompilationStartAnalysisContext context
...
+				return;
+			}
+
+			var checker = new ImmutableDefinitionChecker(
+				compilation: ctx.Compilation,
+				diagnosticSink: ctx.ReportDiagnostic,
+				context: immutabilityContext
+			);
+
+			checker.CheckMember( ctx.Symbol );
```
j3parker, line 71: Finally... if you look at UnsafeStaticsAnalyzer it has a lot of duplication with the MutabilityInspector.
It is because it came first, and the MutabilityInspector was pulled out from it and generalized.
I was (very) slowly walking that back.

Rewriting stuff was the right call.

j3parker:
(rename because its not just about type declarations now?)

omsmith:
I figured we'd rename it to ImmutabilityAnalyzer, lol

j3parker:
Cool. I didn't even rename it correctly so I backed that out.

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -37,6 +37,38 @@ CompilationStartAnalysisContext context
...
+			// We only care about checking static fields/properties. These
+			// are global variables, so we always want them to be immutable.
+			// The fields/properties of [Immutable] types get handled via
+			// AnalyzeTypeDeclaration.
+			if ( !ctx.Symbol.IsStatic ) {
+				return;
+			}
+
+			// Ignore const things, which include enum names.
+			if ( ctx.Symbol is IFieldSymbol f && f.IsConst ) {
```
omsmith, line 61: Can you send xplain this one? I don't think I understand.

j3parker:
```cs
enum Foo {
  Bar
}
```

Bar is a static field that is not implicitly declared. Critically it has zero variable declarators, so if passed it to CheckMember it would throw on an invalid cast in some code that looks up syntax locations.

j3parker: Also probably just regular consts but their types are ok

omsmith (approved):

PR #628, opened 2020-11-20
Remove old static analyzer

j3parker:
Needs #627


omsmith:
Lets remove it at the same time as the old ImmutabilityAnalyzer maybe?

Might be good to do the new statics checking to run parallel to this like for the non-statics?

j3parker:
🤷 ok

PR #631, opened 2020-11-20
Exclude generated resource file code from static var checks

j3parker:


src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -62,6 +63,14 @@ ImmutabilityContext immutabilityContext
 				return;
 			}
 
+			// We want to run our analyzer to run for generated code too, but
+			// the generated code from resx files contain static variables
+			// which don't contribute to mutability but can't be analyzed as
+			// safe (e.g. CultureInfos, only some of those are read-only).
+			if( IsFromGeneratedResourcesFile( ctx.Symbol ) ) {
+				return;
+			}
```
j3parker, line 72: https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Immutability/UnsafeStaticsAnalyzer.cs#L26-L32

This is less crazy than what we're doing now. I'm open to other ideas in the future.

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -110,5 +119,12 @@ INamedTypeSymbol analyzedType
 
 			return false;
 		}
+
+		private static bool IsFromGeneratedResourcesFile( ISymbol symbol ) {
+			var tree = symbol.DeclaringSyntaxReferences.Single().SyntaxTree;
+
+			return tree.FilePath.EndsWith( ".Designer.cs" )
+				&& tree.ToString().StartsWith( "//------------------------------------------------------------------------------\n// <auto-generated>" );
```
j3parker, line 127: Just this marker text is enough to exclude the file in `UnsafeStaticsAnalyzer` right now. Restricting it to just `*.Designer.cs` so we analyze more things at least.

src/D2L.CodeStyle.Analyzers/Immutability/TypeDeclarationImmutabilityAnalyzer.cs
```diff
@@ -110,5 +119,12 @@ INamedTypeSymbol analyzedType
 
 			return false;
 		}
+
+		private static bool IsFromGeneratedResourcesFile( ISymbol symbol ) {
+			var tree = symbol.DeclaringSyntaxReferences.Single().SyntaxTree;
```
j3parker, line 124: `Single()` -- I don't know how this could break for properties/fields.
It doesn't break in our codebase.

omsmith, line 124: C#9 adds support for partial methods (separating declaration and definition). Don't recall if that applied to fields and members as well.

j3parker: Doesn't look like it.

omsmith (approved):

PR #632, opened 2020-11-20
More aggressively ignore static vars in generated code

j3parker:
This is unfortunate, but we've got a backlog of things to fix up in our
generated code too.

Everything looks practically safe though.

This matches the behaviour of UnsafeStaticsAnalyzer more closely.


omsmith (approved):

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -212,5 +214,69 @@ string name
 			typeSymbol = compilation.GetTypeByMetadataName( fullyQualifiedMetadataName );
 			return ( typeSymbol != null );
 		}
+
+		// Inspired by https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/Portable/InternalUtilities/GeneratedCodeUtilities.cs
```
j3parker, line 218: Boo. This code is similar to that. They don't expose this stuff publicly. We can't do the same thing UnsafeStaticsAnalyzer does (disable analysis for generated code) because that's a per-analyzer setting (we don't have to, but I think it's good to have these analyzers be combined... at least we get to share the `ImmutabilityContext`).

There is some ongoing debate about making generated code exclusions more flexible in Roslyn (the discussions didn't look like they're going anywhere fast.)

I want to poke at our codegen and make it generate cleaner code.

PR #637, opened 2020-11-26
Upgrade .NET

j3parker:


src/D2L.CodeStyle.Analyzers/D2L.CodeStyle.Analyzers.csproj
```diff
@@ -1,8 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard1.3</TargetFramework>
-    <LangVersion>7.3</LangVersion>
+    <TargetFramework>netstandard2.0</TargetFramework>
```
j3parker, line 4: 2.0 and not 2.1 because we want to reference this DLL from static DI. Cool for now?

j3parker:
Interesting failure, going to do this in pieces.

PR #651, opened 2020-11-27
Make helper function name more accurate

j3parker:


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -157,7 +157,7 @@ ImmutabilityContext context
 				return true;
 			}
 
-			var propInfo = GetPropertySyntax(
+			var propInfo = GetPropertyStuff(
```
j3parker, line 160: When I was writing it, it started off being just syntax but it grew to be random other stuff too.

Issue #582, opened 2020-10-20
Assess recent C# features wrt mutability analyzer


j3parker:
Off the top of my head we probably don't deal with record types well.

I think the analyzer has long-standing trouble with `ValueTuple` syntax (e.g. `(int, string)`) -- we should add some spec tests there

omsmith:
> I think the analyzer has long-standing trouble with ValueTuple syntax (e.g. (int, string)) -- we should add some spec tests there

Nah, ValueTuple should already be working in the existing analyzers. https://github.com/Brightspace/D2L.CodeStyle/pull/507

j3parker:
I'll do a quick audit of the changelogs, but records were the big one.
Went from C# 7.3 -> 9.0.

j3parker:
* Mark these types immutable:
  * System.Index
  * System.Range
* [Wider support for init-only setters](https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9#init-only-setters)
* Test cases for:
  * Target-typed new: `Foo x = new();`

j3parker:
@omsmith could you double check C# 7.2 - 9.0 and see if anything else looks relevant?
https://docs.microsoft.com/en-us/dotnet/csharp/whats-new/csharp-9

j3parker:
He looked. Theres maybe some stuff for the stateless func stuff but o/w we look good.

PR #654, opened 2020-11-27
Fully support init-only properties

j3parker:
Part of #582

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
@@ -375,27 +375,40 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			ExpressionSyntax Initializer,
 			bool IsAutoImplemented,
 			bool IsReadOnly
-		) GetPropertyStuff( IPropertySymbol symbol, SyntaxNode syntax )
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -373,29 +372,35 @@ ExpressionSyntax initializer
...
+			if ( symbol.IsReadOnly ) {
+				return true;
+			}
+
+			// For our purposes this is as good as readonly.
+			// Properties in a concisely-defined record like
+			// "record Foo(int X);" will always be init-only rather than
+			// readonly, but there is also regular properties with
+			// "int X { get; init; }".
+			return symbol.SetMethod != null && symbol.SetMethod.IsInitOnly;
```
j3parker, line 403: Init-only stuff is actually a bit different:

```cs
using System;

public class Program {
    class Foo { public object X { get; init; } = new object();}
	
    public static void Main() {
        var l = new Foo(){
            X = new Program()
        };
        
        Console.WriteLine(l.X.GetType().FullName);
    }
}
```

Prints "Program". i.e. callers can overwrite the initializers (not just constructors like `readonly`) so we should never narrow our analysis based on initializers.

I'll work on this a bit.

omsmith (approved):

PR #678, opened 2020-12-15
allow multiple expectations on the same node

omsmith:
Split by a pipe

```csharp
/* Foo() | Bar() */ Thing /**/
```


omsmith:
Ignore whitespace
https://github.com/Brightspace/D2L.CodeStyle/pull/678/files?w=1

omsmith:
Currently trying to nest them

```csharp
/* Foo() */ /* Bar() */ Thing /**/ /**/
```

Has Foo's expected location including Bar's comment. We could look at the syntax tree and filter those our when choosing the real location instead of having a separator, but, :shrug:.

j3parker (approved):
cool

Issue #677, opened 2020-12-11
Analysis of [Immutable] method type parameters should consider implied names


omsmith:
```csharp
class Foo {
  public static void Bar<[Immutable] T>( T value ) {}

  public static void Baz() {
    object value = null;
    Bar<object>( value );
    Bar( value ); // Analyzer doesn't catch this
  }
} 
```

The current analyzer is registered against `GenericNameSyntax`. This doesn't catch the _implied_ `Bar<object>` call. The analyzer should instead, I expect, look at `InvocationExpression`.

The same analysis path is used to restrict type parameters of named types, and so they will likely need to be split.

PR #679, opened 2020-12-15
Analysis of [Immutable] method type parameters should consider implied names #677

TimothyJCowen:
- Changed GenericNameSyntax references to SimpleNameSyntax, and trigger it with GenericNameSyntax AND IdentifierNameSyntax
- Added tests to support new case

Closes https://github.com/Brightspace/D2L.CodeStyle/issues/677

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
@@ -143,11 +143,9 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			SymbolInfo info = ctx.SemanticModel.GetSymbolInfo( syntax, ctx.CancellationToken );
 
 			// Ignore anything that cannot have type arguments/parameters
-			if( info.Symbol is not INamedTypeSymbol && info.Symbol is not IMethodSymbol ) {
+			if( !GetTypeParamsAndArgs( info.Symbol, out var typeParameters, out var typeArguments ) ) {
 				return;
 			}
-
-			var (typeParameters, typeArguments) = GetTypeParamsAndArgs( info.Symbol );
 
 			int i = 0;
 			var paramArgPairs = typeParameters.Zip( typeArguments, ( p, a ) => (p, a, i++) );
@@ -160,11 +158,11 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 					continue;
 				}
 
-				// If the syntax is a GenericName (has explicit type arguments) then the error should be on the argument
-				// Otherwise, it should be on the identifier itself
 				if( !immutabilityContext.IsImmutable(
 					type: argument,
 					kind: ImmutableTypeKind.Total,
+					// If the syntax is a GenericName (has explicit type arguments) then the error should be on the argument
+					// Otherwise, it should be on the identifier itself
 					getLocation: () => syntax is GenericNameSyntax genericSyntax
 						? genericSyntax.TypeArgumentList.Arguments[position].GetLocation()
 						: syntax.Identifier.GetLocation(),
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -153,15 +160,21 @@ GenericNameSyntax syntax
...
+					getLocation: () => syntax is GenericNameSyntax genericSyntax
+						? genericSyntax.TypeArgumentList.Arguments[position].GetLocation()
+						: syntax.Identifier.GetLocation(),
 					out Diagnostic diagnostic
 				) ) {
 					// TODO: not necessarily a good diagnostic for this use-case
 					ctx.ReportDiagnostic( diagnostic );
 				}
+
+
```
omsmith, line 177: nit: extra diff

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -129,17 +130,23 @@ INamedTypeSymbol typeSymbol
...
 		) {
 			if( syntax.IsFromDocComment() ) {
 				// ignore things in doccomments such as crefs
 				return;
 			}
 
 			SymbolInfo info = ctx.SemanticModel.GetSymbolInfo( syntax, ctx.CancellationToken );
+
+			// Ignore anything that cannot have type arguments/parameters
+			if( info.Symbol is not INamedTypeSymbol && info.Symbol is not IMethodSymbol ) {
```
j3parker, line 146: I understand why this check is necessary but it's not totally obvious why... it's at least partially because `GetTypeParamsAndArgs` throws. It's a bit confusing to duplicate this check in two places.

Maybe we should change things a bit to use out vars and the `TryX` pattern instead of destructuring assignment: 

```cs
if ( !GetTypeParamsAndArgs( info.Symbol, out var typeParameters, out var typeArguments ) {
  // blah
  return;
}
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -153,15 +160,21 @@ GenericNameSyntax syntax
 					continue;
 				}
 
+				// If the syntax is a GenericName (has explicit type arguments) then the error should be on the argument
+				// Otherwise, it should be on the identifier itself
```
j3parker, line 164: Thanks for the comment. Maybe it should go beside the `getLocation:` argument though?

j3parker (approved):

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -173,20 +171,23 @@ SimpleNameSyntax syntax
...
+				case IMethodSymbol method:
+					typeParameters = method.TypeParameters;
+					typeArguments = method.TypeArguments;
+					return true;
+				case INamedTypeSymbol namedType:
+					typeParameters = namedType.TypeParameters;
+					typeArguments = namedType.TypeArguments;
+					return true;
+				default:
+					return false;
```
omsmith, line 189: I'm surprised this compiles since the out parameters aren't assigned, but, :shrug:

omsmith (approved):

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -165,15 +174,20 @@ GenericNameSyntax syntax
...
-		) GetTypeParamsAndArgs( ISymbol type )
-			=> type switch {
-				IMethodSymbol method => (method.TypeParameters, method.TypeArguments),
-				INamedTypeSymbol namedType => (namedType.TypeParameters, namedType.TypeArguments),
-
-				_ => throw new NotImplementedException(),
-			};
+		private static bool GetTypeParamsAndArgs( ISymbol type, out ImmutableArray<ITypeParameterSymbol> typeParameters, out ImmutableArray<ITypeSymbol> typeArguments ) {
+			switch (type)
+			{
```
j3parker, line 179: nit: whitespace
(is your editor picking up our .editorconfig settings?)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
j3parker: I think it's because they are value types which are always "definitely assigned" (term from spec)

PR #680, opened 2020-12-15
General formatting update

TimothyJCowen:
Updated .editorconfig to remove suggestions for using explicit types rather than 'var'.

Ran 'ImmutabilityAnalyzer.cs' through formatter to fix the codestyle throughout.


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -9,185 +8,184 @@
 
 namespace D2L.CodeStyle.Analyzers.Immutability {
 	[DiagnosticAnalyzer( LanguageNames.CSharp )]
-	public sealed class ImmutabilityAnalyzer : DiagnosticAnalyzer {
```
j3parker, line 12: This change is switch to spaces for indent but our editorconfig says tabs... ?

j3parker (changes requested):

j3parker (approved):

Issue #663, opened 2020-11-27
Add diagnostic to block use of [OnlyIf] in method declarations


j3parker:

omsmith:
```csharp
public sealed class Foo {

  public void SomeMethod<[CondtionallyImmutable.OnlyIf] T>() {}

}
```

The diagnostic should be placed on `T` and note that `[ConditionallyImmutable.OnlyIf]` is only valid on parameters of named types.

PR #681, opened 2020-12-16
Add diagnostic to block use of [OnlyIf] in method declarations #663

TimothyJCowen:
- Added syntax checking for MethodDeclaration and LocalFunctionStatement
  - Checks if any type parameters have the [ConditionallyImmutable.OnlyIf] attribute
- Added supporting tests

Closes https://github.com/Brightspace/D2L.CodeStyle/issues/663

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -528,7 +528,7 @@ namespace D2L.CodeStyle.Analyzers {
 
 		public static readonly DiagnosticDescriptor UnexpectedConditionalImmutability = new DiagnosticDescriptor(
 			id: "D2L0071",
-			title: "The [ConditionallyImmutable.OnlyIf] attribute cannot be applied to this member.",
+			title: "The [ConditionallyImmutable.OnlyIf] attribute cannot be used with methods.",
 			messageFormat: "The [ConditionallyImmutable.OnlyIf] attribute is only valid on parameters of named types.",
 			category: "Immutability",
 			defaultSeverity: DiagnosticSeverity.Error,
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -525,5 +525,14 @@ namespace D2L.CodeStyle.Analyzers {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor UnexpectedConditionalImmutability = new DiagnosticDescriptor(
+			id: "D2L0071",
+			title: "The [ConditionallyImmutable.OnlyIf] attribute cannot be applied to this member.",
+			messageFormat: "The [ConditionallyImmutable.OnlyIf] attribute is only valid on parameters of named types.",
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
 	}
 }
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
@@ -183,6 +183,17 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		private static void AnalyzeConditionalImmutability( SyntaxNodeAnalysisContext ctx ) {
 			var syntax = ctx.Node;
 
+			// Get the symbol for the method
+			if( ctx.SemanticModel.GetDeclaredSymbol( ctx.Node ) is not IMethodSymbol symbol ) {
+				return;
+			}
+
+			// Exit if this somehow is not possible
+			// We don't care about arguments so throw them out
+			if( !GetTypeParamsAndArgs( symbol, out var typeParameters, out _ ) ) {
+				return;
+			}
+
 			// Retrieve the relevant parameter list
 			TypeParameterListSyntax paramListSyntax;
 			switch( syntax ) {
@@ -196,25 +207,20 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 					return;
 			}
 
-			if( paramListSyntax == null ) {
-				return;
-			}
+			// Iterate through the parameter symbols
+			for( int i = 0; i < typeParameters.Length; i++ ) {
+				var parameter = typeParameters[i];
 
-			// Iterate through the individual parameters of the method
-			foreach( TypeParameterSyntax parameter in paramListSyntax.Parameters ) {
-
-				// Iterate through the attributes on the current parameter
-				foreach( AttributeListSyntax attribute in parameter.AttributeLists ) {
-
-					// Check if the current attribute is the conditional immutability attribute
-					if( attribute.Attributes.ToString() == "ConditionallyImmutable.OnlyIf" ) {
-						var diagnostic = Diagnostic.Create(
-							Diagnostics.UnexpectedConditionalImmutability,
-							attribute.GetLocation() );
-						ctx.ReportDiagnostic( diagnostic );
-						break;
-					}
+				// Check if the parameter has the [OnlyIf] attribute
+				if( !Attributes.Objects.OnlyIf.IsDefined( parameter ) ) {
+					continue;
 				}
+
+				// Create the diagnostic on the parameter (including the attribute)
+				var diagnostic = Diagnostic.Create(
+					Diagnostics.UnexpectedConditionalImmutability,
+					paramListSyntax.Parameters[i].GetLocation() );
+				ctx.ReportDiagnostic( diagnostic );
 			}
 		}
 
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
@@ -181,8 +181,6 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		}
 
 		private static void AnalyzeConditionalImmutability( SyntaxNodeAnalysisContext ctx ) {
-			var syntax = ctx.Node;
-
 			// Get the symbol for the method
 			if( ctx.SemanticModel.GetDeclaredSymbol( ctx.Node ) is not IMethodSymbol symbol ) {
 				return;
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
@@ -61,10 +61,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			);
 
 			context.RegisterSyntaxNodeAction(
-				ctx => AnalyzeConditionalImmutability(
-					ctx,
-					ctx.Node
-				),
+				AnalyzeConditionalImmutability,
 				SyntaxKind.MethodDeclaration,
 				SyntaxKind.LocalFunctionStatement
 			);
@@ -183,32 +180,22 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			}
 		}
 
-		private static void AnalyzeConditionalImmutability(
-			SyntaxNodeAnalysisContext ctx,
-			SyntaxNode syntax
-		) {
-			// Ignore references in DocComments such as crefs
-			if( syntax.IsFromDocComment() ) {
-				return;
-			}
+		private static void AnalyzeConditionalImmutability( SyntaxNodeAnalysisContext ctx ) {
+			var syntax = ctx.Node;
 
-			// Methods can be either of the following depending on where they are declared
-			MethodDeclarationSyntax methodSyntax = null;
-			LocalFunctionStatementSyntax functionSyntax = null;
-
+			// Retrieve the relevant parameter list
+			TypeParameterListSyntax paramListSyntax;
 			switch( syntax ) {
 				case MethodDeclarationSyntax:
-					methodSyntax = (MethodDeclarationSyntax)syntax;
+					paramListSyntax = ( (MethodDeclarationSyntax)syntax ).TypeParameterList;
 					break;
 				case LocalFunctionStatementSyntax:
-					functionSyntax = (LocalFunctionStatementSyntax)syntax;
+					paramListSyntax = ( (LocalFunctionStatementSyntax)syntax ).TypeParameterList;
 					break;
 				default:
 					return;
 			}
 
-			// Get the parameter list syntax and make sure it is not null
-			TypeParameterListSyntax paramListSyntax = methodSyntax?.TypeParameterList ?? functionSyntax?.TypeParameterList;
 			if( paramListSyntax == null ) {
 				return;
 			}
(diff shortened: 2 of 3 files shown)


src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -525,5 +525,14 @@ public static class Diagnostics {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor UnexpectedConditionalImmutability = new DiagnosticDescriptor(
+			id: "D2L0071",
+			title: "The [ConditionallyImmutable.OnlyIf] attribute cannot be applied to this member.",
```
j3parker, line 531:
```suggestion
			title: "The [ConditionallyImmutable.OnlyIf] attribute cannot be used with methods.",
```

(more specific, so that it's easy to understand what's wrong without also looking at the line of code)

TimothyJCowen: Sounds good. I wasn't sure if there were other things it might be applied to in the future, which is why I left it generic.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -58,6 +59,15 @@ CompilationStartAnalysisContext context
 				SyntaxKind.IdentifierName,
 				SyntaxKind.GenericName
 			);
+
+			context.RegisterSyntaxNodeAction(
+				ctx => AnalyzeConditionalImmutability(
+					ctx,
+					ctx.Node
+				),
```
j3parker, line 67: nit: splitting here is fine but we usually spill only when it gets to 80 chars or just there is a lot to read/understand.

less-nit: you can probably drop the second argument here and just get it from ctx inside the method declartion. We pass lambdas as the action argument when we need to inject something from the stack into the action (like `immutabilityContext`). This happens because these analyzers can be reused in parallel for different compilations, so we don't want to store this stuff in a member variable. So what I'm suggesting:

```suggestion
				AnalyzeConditionalImmutability,
```

(but you'd also need to tweak the defn of that method)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -173,6 +183,54 @@ SimpleNameSyntax syntax
...
 
+		private static void AnalyzeConditionalImmutability(
+			SyntaxNodeAnalysisContext ctx,
+			SyntaxNode syntax
+		) {
+			// Ignore references in DocComments such as crefs
+			if( syntax.IsFromDocComment() ) {
+				return;
+			}
+
```
j3parker, line 194:
```suggestion
```

I don't think this is necessary, because neither of these kinds of syntax can show up inside a doc comment?

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -173,6 +183,54 @@ SimpleNameSyntax syntax
...
+			// Methods can be either of the following depending on where they are declared
+			MethodDeclarationSyntax methodSyntax = null;
+			LocalFunctionStatementSyntax functionSyntax = null;
```
j3parker, line 197: Since the only usage of these is in the expression ` methodSyntax?.TypeParameterList ?? functionSyntax?.TypeParameterList` it would be simpler to instead have a `TypeParameterListSyntax` and use the switch to pull it out.

Basically there's some duplication here: the `switch` is handling the difference between `MethodDeclarationSyntax` and `LocalFunctionStatementSyntax`, but then it has to get handled a second time in the expression (which is harder to read IMO because of multiple `?.` and a `??`).

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
j3parker: At this time the only things that can be generic are type decls and methods. If they add something else to the language we can adjust the diagnostic later :)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -173,6 +183,54 @@ SimpleNameSyntax syntax
...
+			}
+
+			// Iterate through the individual parameters of the method
+			foreach( TypeParameterSyntax parameter in paramListSyntax.Parameters ) {
+
+				// Iterate through the attributes on the current parameter
+				foreach( AttributeListSyntax attribute in parameter.AttributeLists ) {
+
+					// Check if the current attribute is the conditional immutability attribute
+					if( attribute.Attributes.ToString() == "ConditionallyImmutable.OnlyIf" ) {
```
j3parker, line 223: The downside to using strings to compare symbols is that it's pretty fragile, e.g.

```cs
using static D2L.CodeStyle.Analayzers.Something.ConditionallyImmutable;

// ...

void Foo<[OnlyIf]T>() {}
```

wouldn't get caught. It's not a big deal here but I'll link to an example of how to do this purely with symbols in a sec.

TimothyJCowen: I appreciate that, I was expecting you to suggest something like that

j3parker: https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs#L22

You can grab a reference to the symbol from the compilation inside the wireup like this, and pass it down into this method and use it for comparison without conversion to strings.

j3parker: Yeah, it's hard to know what the right thing to do for this stuff is -- we've learned this stuff the hard way haha.

j3parker: I guess reading a lot of documentation would help but (A) who does that (B) it is quite scattered/non-existent.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -173,6 +183,54 @@ SimpleNameSyntax syntax
 			}
 		}
 
+		private static void AnalyzeConditionalImmutability(
```
omsmith, line 186: Rather than work at the syntax level here, I would try switching to symbols as quickly as possible.

```csharp
if( !( ctx.SemanticModel.GetDeclaredSymbol( ctx.Node ) is IMethodSymbol symbol ) ) {
  return;
}

foreach( IParameterSymbol parameter in symbol.TypeParameters ) {
  if( !Attributes.Objects.OnlyIf.IsDefined( parameter ) ) {
    continue;
  }

  // diagnostic
}
```

LocalFunction and MethodDeclaration should both provide IMethodSymbol even though GetDeclaredSymbol doesn't have an overload for LocalFunction specifically. You may still need to work with the input syntax node directly for placing the diagnostic.

omsmith: `Attributes.Objects.OnlyIf.IsDefined` does work with strings, but it does so in a well-defined way, and is relatively idiomatic in our codebase.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
omsmith: The pattern in our analyzers of pulling out the node early here is when we're registering for one kind and we can confidently cast the `SyntaxNode` to the type we've registered for.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
j3parker: Why symbols? These actions run later, its nice to do syntax when reasonable because it shows up in the editor faster.

j3parker: > Attributes.Objects.OnlyIf.IsDefined does work with strings, but it does so in a well-defined way, and is relatively idiomatic in our codebase.

I've been slowly removing it because it's bad :) It is used in these analyzers though.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
j3parker: Yeah, but we're not casting here, and can't because it could be one of two things.

j3parker: Other examples:

https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Language/ClassShouldBeSealedAnalyzer.cs#L25
https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Language/DefaultValueConsistencyAnalyzer.cs#L31
https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs#L46-L64
https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Language/StructShouldBeReadonlyAnalyzer.cs#L29

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
omsmith: My suggestion is still registered for syntax analysis, the symbols just tend to be easier to work with?

j3parker: Oh yes, I misread.

(Also to be fair: waiting for the semantic model (which is common) mostly negates the latency benefits of syntax analyzers)

TimothyJCowen: I'm playing with Owen's idea. The problem is that I still need to hold some syntax to get the location of where to put the error, and it's actually less intuitive that way.

omsmith: > (Also to be fair: waiting for the semantic model (which is common) mostly negates the latency benefits of syntax analyzers)

Which would be true if comparing the attribute by symbol anyway, since we'd be calling `GetDeclaredSymbol`/`GetSymbolInfo` on each type parameter in order to do the comparison

omsmith: > I've been slowly removing it because it's bad :) It is used in these analyzers though.

Yep, I think it would be reasonable to make them non-static and add some sort of wrapper type that gets them lazily from the compilation. Having them easily reference-able by name and not repeated everywhere is pretty nice. I definitely did what you were suggesting outside of the immutability analyzers.

j3parker: I'm good either way here

omsmith: > I'm playing with Owen's idea. The problem is that I still need to hold some syntax to get the location of where to put the error, and it's actually less intuitive that way.

To avoid having to manage "syncing" the input syntax and the symbol, you should be able to use `parameter.DeclarationSyntaxReferences[0].GetSynax()` or similar. I wouldn't be terribly worried about the hop back to syntax land in this manner being maybe unperformant (comparatively), since it only happens in the error case. In the general case, the analyzers should produce no diagnostics.

TimothyJCowen: I've updated with another commit, it still uses the syntax reference from before for this purpose.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -173,6 +180,50 @@ SimpleNameSyntax syntax
...
+			// Get the symbol for the method
+			if( ctx.SemanticModel.GetDeclaredSymbol( ctx.Node ) is not IMethodSymbol symbol ) {
+				return;
+			}
+
+			// Exit if this somehow is not possible
+			// We don't care about arguments so throw them out
+			if( !GetTypeParamsAndArgs( symbol, out var typeParameters, out _ ) ) {
+				return;
+			}
```
omsmith, line 195:
```suggestion
```

No need for this call/check, we know we've got an `IMethodSymbol` which has a `TypeParameters`, and as noted we don't need the arguments.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -173,6 +180,50 @@ SimpleNameSyntax syntax
...
+
+			// Retrieve the relevant parameter list
+			TypeParameterListSyntax paramListSyntax;
+			switch( syntax ) {
+				case MethodDeclarationSyntax:
+					paramListSyntax = ( (MethodDeclarationSyntax)syntax ).TypeParameterList;
+					break;
+				case LocalFunctionStatementSyntax:
+					paramListSyntax = ( (LocalFunctionStatementSyntax)syntax ).TypeParameterList;
+					break;
```
omsmith, line 205: Can avoid the extra cast

```suggestion
				case MethodDeclarationSyntax methodSyntax:
					paramListSyntax = methodSyntax.TypeParameterList;
					break;
				case LocalFunctionStatementSyntax localSyntax:
					paramListSyntax = localSyntax.TypeParameterList;
					break;
```


Though it would probably read even better as a switch expression:

```csharp
			// Retrieve the relevant parameter list
			TypeParameterListSyntax paramListSyntax = syntax switch {
				MethodDeclarationSyntax methodSyntax => methodSyntax.TypeParameterList,
				LocalFunctionStatementSyntax localSyntax => localSyntax.TypeParameterList,
				_ => throw new InvalidOperationException(),
			};
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
TimothyJCowen: Ah I read this wrong, give me a second to update it again...

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -173,6 +180,28 @@ SimpleNameSyntax syntax
 			}
 		}
 
+		private static void AnalyzeConditionalImmutability( SyntaxNodeAnalysisContext ctx ) {
+			var syntax = ctx.Node;
+
```
omsmith, line 185: nit: unused variable

```suggestion
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -173,6 +180,26 @@ SimpleNameSyntax syntax
 			}
 		}
 
+		private static void AnalyzeConditionalImmutability( SyntaxNodeAnalysisContext ctx ) {
```
omsmith, line 183: Last suggestion: maybe rename to mention this is about methods?

`AnalyzeMethodDeclarationConditionalImmutabilityAnnotations` or something...

omsmith (approved):

Issue #661, opened 2020-11-27
Add diagnostic for audit attribute conflicts


j3parker:
E.g. you shouldn't be Unaudited and Audited (right now you can be both), and you should use the right one of Mutability or Statics (the wrong ones just get ignored)

omsmith:
```csharp
[Immutable]
public sealed class Foo {
  [Statics.Audited( "Owen Smith", "2020-12-09", "Because I said so" )]
  [Statics.Unaudited( Because.ItHasntBeenLookedAt )]
  private static bool m_bar;

  [Mutability.Audited( "Owen Smith", "2020-12-09", "Because I said so" )]
  [Mutability.Unaudited( Because.ItHasntBeenLookedAt )]
  private bool m_baz;
}
```

A diagostic should be placed on `m_bar` and on `m_baz`, noting that `Audited` and `Unaudited` are mutually exclusive.

PR #683, opened 2020-12-16
Add diagnostic for audit attribute conflicts #661

TimothyJCowen:
- Added new method to analyze if fields, properties and events have both the [Audited] and [Unaudited] attributes
  - This checks against the Mutability attributes AND the static attributes at once
- Added supporting tests

Closes https://github.com/Brightspace/D2L.CodeStyle/issues/661

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -537,17 +537,17 @@ namespace D2L.CodeStyle.Analyzers {
 
 		public static readonly DiagnosticDescriptor ConflictingImmutability = new DiagnosticDescriptor(
 			id: "D2L0072",
-			title: "The [Immutable] and [ConditionallyImmutable.OnlyIf] attributes are mutually exclusive.",
-			messageFormat: "The [Immutable] and [ConditionallyImmutable.OnlyIf] attributes cannot be used on the same parameter.",
+			title: "Conflicting immutability attributes.",
+			messageFormat: "The [{0}] and [{1}] attributes cannot be used on the same {2}.",
 			category: "Immutability",
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
 
-		public static readonly DiagnosticDescriptor ConflictingAuditing = new DiagnosticDescriptor(
+		public static readonly DiagnosticDescriptor InvalidAuditType = new DiagnosticDescriptor(
 			id: "D2L0073",
-			title: "The [Audited] and [Unaudited] attributes are mutually exclusive.",
-			messageFormat: "The [Audited] and [Unaudited] attributes cannot be used on the same property.",
+			title: "Wrong type of auditing was used.",
+			messageFormat: "A {0} {1} should audit using the [{2}] attributes.",
 			category: "Immutability",
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -543,5 +543,14 @@ namespace D2L.CodeStyle.Analyzers {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor ConflictingAuditing = new DiagnosticDescriptor(
+			id: "D2L0073",
+			title: "The [Audited] and [Unaudited] attributes are mutually exclusive.",
+			messageFormat: "The [Audited] and [Unaudited] attributes cannot be used on the same property.",
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
 	}
 }
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -547,7 +547,7 @@ namespace D2L.CodeStyle.Analyzers {
 		public static readonly DiagnosticDescriptor InvalidAuditType = new DiagnosticDescriptor(
 			id: "D2L0073",
 			title: "Wrong type of auditing was used.",
-			messageFormat: "A {0} {1} should audit using the [{2}] attributes.",
+			messageFormat: "A {0} {1} should be audited using the [{2}] attributes.",
 			category: "Immutability",
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
@@ -70,7 +70,14 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		/// values.
 		/// </remarks>
 		public bool CheckMember( ISymbol member ) {
-			if ( MutabilityAuditor.IsAudited( member, out var location ) ) {
+			if ( MutabilityAuditor.IsAudited( member, out var location, out var diagnostic ) ) {
+
+				// If they have one of the auditing attributes,
+				// error if a diagnostic occurred
+				if( diagnostic is not null ) {
+					m_diagnosticSink( diagnostic );
+				}
+
 				// If they have one of the auditing attributes, run the
 				// checks anyway and error if they are unnecessary
 				if( CheckMember( diagnosticSink: _ => { }, member ) ) {
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
@@ -70,13 +70,10 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		/// values.
 		/// </remarks>
 		public bool CheckMember( ISymbol member ) {
-			if ( MutabilityAuditor.IsAudited( member, out var location, out var diagnostic ) ) {
-
-				// If they have one of the auditing attributes,
-				// error if a diagnostic occurred
-				if( diagnostic is not null ) {
-					m_diagnosticSink( diagnostic );
-				}
+			if( MutabilityAuditor.CheckAudited(
+				member,
+				m_diagnosticSink,
+				out var location ) ) {
 
 				// If they have one of the auditing attributes, run the
 				// checks anyway and error if they are unnecessary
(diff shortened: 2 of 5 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -70,7 +70,14 @@ ImmutabilityContext context
 		/// values.
 		/// </remarks>
 		public bool CheckMember( ISymbol member ) {
-			if ( MutabilityAuditor.IsAudited( member, out var location ) ) {
+			if ( MutabilityAuditor.IsAudited( member, out var location, out var diagnostic ) ) {
+
+				// If they have one of the auditing attributes,
+				// error if a diagnostic occurred
```
j3parker, line 76: super nit: we wrap at around 80 chars :)

So if we have a conflict we will report it but carry on as if there was no conflict (i.e. assume this is audited, but also check that an audit is even warranted in the first place)... that makes sense to me.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityAuditor.cs
```diff
@@ -3,15 +3,24 @@
...
 namespace D2L.CodeStyle.Analyzers.Immutability {
 	internal sealed class MutabilityAuditor {
-		public static bool IsAudited( ISymbol symbol, out Location location ) {
+		public static bool IsAudited( ISymbol symbol, out Location location, out Diagnostic diagnostic ) {
 			AttributeData attr;
+			diagnostic = null;
+
+			// Check if there is a conflicting audit AND unaudit
+			if( ( Attributes.Statics.Audited.IsDefined( symbol ) || Attributes.Mutability.Audited.IsDefined( symbol ) )
+				&& ( Attributes.Statics.Unaudited.IsDefined( symbol ) || Attributes.Mutability.Unaudited.IsDefined( symbol ) ) ) {
```
j3parker, line 12: It makes sense to me that the diagnostic goes on the identifier rather than an attribute, because we can't really choose which attribute to put it on.

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -543,5 +543,14 @@ public static class Diagnostics {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor ConflictingAuditing = new DiagnosticDescriptor(
+			id: "D2L0073",
+			title: "The [Audited] and [Unaudited] attributes are mutually exclusive.",
```
j3parker, line 549: Avoid speaking of `[Audited]` and `[Unaudited]` because "those aren't things" (it's `[Statics.Audited]` etc.)

For the title you can make it generic, and then in the `messageFormat` use `[{0}.Audited]` etc. and pass that in from the analyzer.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityAuditor.cs
```diff
@@ -3,15 +3,24 @@
...
+			}
 
 			if( symbol.IsStatic ) {
 				attr = Attributes.Statics.Audited.GetAll( symbol ).FirstOrDefault()
-				    ?? Attributes.Statics.Unaudited.GetAll( symbol ).FirstOrDefault();
+					?? Attributes.Statics.Unaudited.GetAll( symbol ).FirstOrDefault();
 			} else {
 				attr = Attributes.Mutability.Audited.GetAll( symbol ).FirstOrDefault()
-				    ?? Attributes.Mutability.Unaudited.GetAll( symbol ).FirstOrDefault();
+					?? Attributes.Mutability.Unaudited.GetAll( symbol ).FirstOrDefault();
```
j3parker, line 23: We're still allowing

```cs
[Statics.Audited( ... )]
object x;
```

even though `x` is non-static, right? (this is mentioned in #661 as an undesirable thing)

Maybe we could do something like:

```cs
var hasStaticsAudited = ...;
var hasStaticsUnaudited = ...;
var hasBothStaticsAttributes = hasStaticsAudited  && hasStaticsUnaudited;
var hasEitherStaticsAttribute = hasStaticsAudited  || hasStaticsUnaudited;

var hasMutabilityAudited = ...;
var hasMutabilityUnaudited = ...;
var hasBothMutabilityAttribute = hasMutabilityAudited  && hasMutabilityUnaudited ;
var hasEitherMutabilityAttribute = hasMutabilityAudited  || hasMutabilityUnaudited ;

if( !hasEitherStaticsAttribute && !hasEitherMutabilityAttribute ) {
  // no attributes at all is ok
  return false;
}

if( hasBothStaticsAttributes ) {
  // diagnostic: can't have both audited and unaudited
  return true;
}

if( hasBothMutabilityAttributes ) {
  // diagnostic: can't have both audited and unaudited
  return true;
}

if ( symbol.IsStatic && hasEitherMutabilityAttribute ) {
  // diagnostic: wrong kind
  return true;
}

if ( !symbol.IsStatic && hasEitherStaticsAttribute ) {
  // diagnostic: wrong kind
  return true;
}

return true;
```

This would give you the ability to customize the diagnostic via `messageFormat` to the scenario.
You'd need an extra diagnostic for "wrong kind" (which would also need some `messageFormat`ing)

j3parker:
Looks good! My main feedback is that last comment RE: handling all the cases, and making the diagnostics more ergonomic.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityAuditor.cs
```diff
@@ -3,15 +3,24 @@
 
 namespace D2L.CodeStyle.Analyzers.Immutability {
 	internal sealed class MutabilityAuditor {
-		public static bool IsAudited( ISymbol symbol, out Location location ) {
+		public static bool IsAudited( ISymbol symbol, out Location location, out Diagnostic diagnostic ) {
 			AttributeData attr;
+			diagnostic = null;
```
omsmith, line 8: Early assignment of out parameters is an anti-pattern. This is usually done to defeat CS0177, however the compiler error is also there to help you. Assigning at the top means it won't catch if you forgot to assign a proper value in a given case.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityAuditor.cs
```diff
@@ -3,15 +3,24 @@
...
 			AttributeData attr;
+			diagnostic = null;
+
+			// Check if there is a conflicting audit AND unaudit
+			if( ( Attributes.Statics.Audited.IsDefined( symbol ) || Attributes.Mutability.Audited.IsDefined( symbol ) )
+				&& ( Attributes.Statics.Unaudited.IsDefined( symbol ) || Attributes.Mutability.Unaudited.IsDefined( symbol ) ) ) {
+				diagnostic = Diagnostic.Create(
+					Diagnostics.ConflictingAuditing,
+					symbol.DeclaringSyntaxReferences[0].GetSyntax().GetLastToken().GetLocation() );
+			}
```
omsmith, line 16: This will raise a diagnostic for the following code

```csharp
[Statics.Unaudited( Because.Foo )]
[Mutability.Audited( "Owen Smith", "2020-12-17", "Foo" )]
private bool m_disposed;
```

While it's still invalid code _regardless_, these issues are about trying to improve (or get back to good) the developer experience and give the correct suggestions.

We don't want `ConflictingAudit` to be raised for this case - instead this would just have a diagnostic for the `Statics.Unaudited` being used on a non-static member. That particular diagnostic doesn't need to happen in this PR though.

`ConflictingAudit` should be raised for Audit+Unaudited of the same `Statics`/`Mutability`.

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityAuditor.cs
```diff
@@ -3,15 +3,24 @@
 
 namespace D2L.CodeStyle.Analyzers.Immutability {
 	internal sealed class MutabilityAuditor {
-		public static bool IsAudited( ISymbol symbol, out Location location ) {
+		public static bool IsAudited( ISymbol symbol, out Location location, out Diagnostic diagnostic ) {
```
omsmith, line 6: As referred to in https://github.com/Brightspace/D2L.CodeStyle/pull/683/files#r545201100, this area of code will eventually also want to raise a diagnostic for using the wrong audit type as well. While the chances are _slim_ that the following code would occur, we'd want to be able to provide _both_ diagnostics. With that in mind, it may be a better API to accept `diagnosticSink` (`internal delegate void DiagnosticSink( Diagnostic diagnostic );` / `Action<Diagnostic>`) so that we can raise diagnostics as needed internally. `IsAudited( ISymbol symbol, DiagnosticSink diagnosticSink, out Location location )`. With that in mind, it may also make sense to rename it to `CheckAudited`.

```csharp
[Statics.Unaudited( Because.Foo )]
[Mutability.Unaudited( Because.Foo )]
[Mutability.Audited( "Owen Smith", "2020-12-17", "Foo" )]
private bool m_disposed;
```

omsmith:
Ah. Hadn't refreshed and seen @j3parker's comments. Sorry for parroting some of it at you again TJ!

src/D2L.CodeStyle.Analyzers/Immutability/MutabilityAuditor.cs
```diff
@@ -1,24 +1,94 @@
...
+				}
 			}
 
+			// Everything looks good, so collect information to determine if
+			// auditing is necessary
+			AttributeData attr
+				= Attributes.Statics.Audited.GetAll( symbol ).FirstOrDefault()
+				?? Attributes.Statics.Unaudited.GetAll( symbol ).FirstOrDefault()
+				?? Attributes.Mutability.Audited.GetAll( symbol ).FirstOrDefault()
+				?? Attributes.Mutability.Unaudited.GetAll( symbol ).FirstOrDefault();
```
omsmith, line 86: This group means that the following will be considered audited:

```csharp
[Mutability.Audited( "Owen Smith", "2020-12-17", "Foo" )]
private static bool m_initialized = false;
```

Would prefer for it to be _not_ considered audited. Meaning we should have a diagnostic about the incorrect audit being used, and about the lack of readonly.

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -537,8 +537,17 @@ public static class Diagnostics {
...
+			messageFormat: "The [{0}] and [{1}] attributes cannot be used on the same {2}.",
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
+
+		public static readonly DiagnosticDescriptor InvalidAuditType = new DiagnosticDescriptor(
+			id: "D2L0073",
+			title: "Wrong type of auditing was used.",
+			messageFormat: "A {0} {1} should audit using the [{2}] attributes.",
```
omsmith, line 550:
```suggestion
			messageFormat: "A {0} {1} should be audited using the [{2}] attributes.",
```

omsmith (approved):

PR #684, opened 2020-12-17
Make TestAttributeMissed an Error

omsmith:
We run our build with `werr`, so make Visual Studio reflect that


mthjones (approved):
LGTM! (**edit**: CI failures not withstanding :))

(We should probably replace this diagnostic with [`nunit.analyzers`](https://github.com/nunit/nunit.analyzers) at some point)

mthjones (approved):

PR #685, opened 2020-12-17
Require named arguments when any two arguments have interchangable types #521

TimothyJCowen:
- Added check for how many times each data type is used in the parameters
- Added diagnostic when the current data type has been used more than once to force naming
- Updated test cases

Closes https://github.com/Brightspace/D2L.CodeStyle/issues/521

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -556,7 +556,7 @@ namespace D2L.CodeStyle.Analyzers {
 		public static readonly DiagnosticDescriptor SwappableArgsShouldBeNamed = new DiagnosticDescriptor(
 			id: "D2L0075",
 			title: "Swappable arguments should be named for readability.",
-			messageFormat: "The arguments for the {0} and {1} parameters may be accidentally swapped. Use named arguments for readability.",
+			messageFormat: "The arguments for the \"{0}\" and \"{1}\" parameters may be accidentally swapped. Use named arguments for readability.",
 			category: "Readability",
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true,
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -554,7 +554,7 @@ namespace D2L.CodeStyle.Analyzers {
 		);
 
 		public static readonly DiagnosticDescriptor SwappableArgsShouldBeNamed = new DiagnosticDescriptor(
-			id: "D2L0075",
+			id: "D2L0074",
 			title: "Swappable arguments should be named for readability.",
 			messageFormat: "The arguments for the \"{0}\" and \"{1}\" parameters may be accidentally swapped. Use named arguments for readability.",
 			category: "Readability",
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -483,7 +483,7 @@ namespace D2L.CodeStyle.Analyzers {
 		public static readonly DiagnosticDescriptor NonImmutableTypeHeldByImmutable = new DiagnosticDescriptor(
 			id: "D2L0066",
 			title: "Type may be mutable.",
-			messageFormat: "The {0} {1} is missing the [Immutable]{2} attribute, so it isn't safe to be held by an immutable type",
+			messageFormat: "The {0} \"{1}\" is missing the [Immutable]{2} attribute, so it isn't safe to be held by an immutable type",
 			category: "Immutability",
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -553,14 +553,14 @@ namespace D2L.CodeStyle.Analyzers {
 			isEnabledByDefault: true
 		);
 
-		public static readonly DiagnosticDescriptor ImplicitUnnamedArgs = new DiagnosticDescriptor(
-			id: "D2L0074",
-			title: "Name arguments to prevent implicit swapping.",
-			messageFormat: "Multiple arguments share the {0} data type. Use named arguments to prevent mistakes.",
+		public static readonly DiagnosticDescriptor SwappableArgsShouldBeNamed = new DiagnosticDescriptor(
+			id: "D2L0075",
+			title: "Swappable arguments should be named for readability.",
+			messageFormat: "The arguments for the {0} and {1} parameters may be accidentally swapped. Use named arguments for readability.",
 			category: "Readability",
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true,
-			description: "When multiple arguments share a data type, they should be named so that they don't get switched accidentally."
+			description: "Swappable arguments should be named for readability."
 		);
 	}
 }
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -552,5 +552,15 @@ namespace D2L.CodeStyle.Analyzers {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor ImplicitUnnamedArgs = new DiagnosticDescriptor(
+			id: "D2L0074",
+			title: "Name arguments to prevent implicit swapping.",
+			messageFormat: "Multiple arguments share the {0} data type. Use named arguments to prevent mistakes.",
+			category: "Readability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true,
+			description: "When multiple arguments share a data type, they should be named so that they don't get switched accidentally."
+		);
 	}
 }
--- a/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
@@ -170,9 +170,12 @@ namespace D2L.CodeStyle.Analyzers.Language {
 			// If more than one argument can apply to multiple parameters,
 			// then require the arguments to be named
 			if( numOfSwappableArgs > 1 ) {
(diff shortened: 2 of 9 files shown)


src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -543,5 +543,15 @@ public static class Diagnostics {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor ImplicitUnnamedArgs = new DiagnosticDescriptor(
```
j3parker, line 547: Let's consolidate on `NamedArgumentsRequired`:

* as a user I don't really care why they are required, just that they are (i.e. the fix is the same in any case)
* the rules are very complex, actually: e.g. we would never require naming for an argument like `m_foo.Bar` being passed to a parameter named `bar`... we don't want people to have to internalize all these rules because they are uninteresting, so we don't need to explain it in the diagnostics
* the conventions at D2L are very strong and intuitive (and this analyzer does a good job of matching the intuition) so its unlikely anyone would ever be surprised or interested in exactly why an argument was required to be named

That diagnostic has a good title: "Named arguments are required for readability.". The description is a little bit off if we consolidate on it, but it's still pretty applicable.

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -123,34 +124,49 @@ SyntaxNodeAnalysisContext ctx
...
-			// Literal arguments should always be named
-			foreach( var arg in unnamedArgs ) {
-
-				// Check if the argument type is literal
-				if( arg.Syntax.Expression is LiteralExpressionSyntax ) {
-
+			}
+
+			// Generate a Dictionary of data type usages
+			Dictionary<string, int> dataTypeCount = new Dictionary<string, int>();
```
j3parker, line 130: This doesn't really handle all the cases unfortunately.

What we want to do is catch cases where arguments could have been swapped, e.g.:

```cs
interface IA {}
interface IB : IA {}
interface IC : IA {}
interface ID : IA, IB {}
// assume impls A, B, C, D

void Foo( IA x, IB y ) {}

// all of these compile. they all base a B (which impls IB) as the second arg,
// but we consider all cases for the first arg:
Foo( new A(), new B() ); // always safe (no argument swapping possible)
Foo( new B(), new B() ); // possible error
Foo( new C(), new B() ); // always safe
Foo( new D(), new B() ); // possible error
```

There is a way to ask the semantic model if a particular conversion exists.

j3parker: From DM: this is roughly what we want to be use, I think: [SemanticModel.ClassifyConversion](https://docs.microsoft.com/en-us/dotnet/api/microsoft.codeanalysis.csharp.csharpextensions.classifyconversion?view=roslyn-dotnet#Microsoft_CodeAnalysis_CSharp_CSharpExtensions_ClassifyConversion_Microsoft_CodeAnalysis_SemanticModel_Microsoft_CodeAnalysis_CSharp_Syntax_ExpressionSyntax_Microsoft_CodeAnalysis_ITypeSymbol_System_Boolean_).

TimothyJCowen: This one was interesting, but it should be good now.

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -123,34 +123,80 @@ SyntaxNodeAnalysisContext ctx
...
+
+					// Get the associated parameter of the argument to compare
+					IParameterSymbol param = compareArg.Syntax.DetermineParameter(
+						ctx.SemanticModel,
+						allowParams: false );
+
+					// Check if the argument could apply to the other parameter
+					if( ctx.SemanticModel.Compilation.ClassifyConversion(
+						argType,
+						param.Type ).Exists ) {
```
j3parker, line 160: This is kind of hard to read, so pulling it out into a local variable would be good

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -123,34 +123,80 @@ SyntaxNodeAnalysisContext ctx
...
+						// The argument can apply to multiple parameters,
+						// so take note of it
+						numOfSwappableArgs++;
```
j3parker, line 163: A few thoughts:

* I think we are complaining here if an argument could be used in the place of another parameter, but we actually care if two arguments could be swapped. So like with `Foo( object x, int x )` we aren't worried about `Foo( new object(), 6 )` even though `6` could go in place of the first parameter.
* Do we want to output a diagnostic for just these ambiguous arguments, or just say "screw it, name every argument"? I think we want to do it individually for these (like we do for literals). We'd need to keep track of the indexes of arguments that are (mutually) swappable, like a `List<int>` or something?
* Actually, now I wonder if we should just be building up a single diagnostic... like instead of immediately emitting a diagnostic for literals we should be adding to a shared `fixerContext` instance and keeping a `bool hadProblems` to emit one diagnostic at the end.

For context: the fixer takes possibly multiple arguments to name in the diagnostic. Ideally I think we'd have at most one diagnostic per function call that we are checking.

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -123,34 +123,80 @@ SyntaxNodeAnalysisContext ctx
...
+
+				ctx.ReportDiagnostic( Diagnostic.Create(
+						descriptor: Diagnostics.NamedArgumentsRequired,
+						location: ctx.Node.GetLocation(),
+						properties: fixerContext ) );
+			}
+
+			// TODO: if there are duplicate typed args then they should be named
+			// These will create a bit more cleanup. Fix should probably name
+			// all the args instead to avoid craziness with overloading.
```
j3parker, line 182: I think you can remove this now...

Oh hey, overloading.... so my previous comment about only naming some of the args.... Hmmm....

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -123,34 +123,80 @@ SyntaxNodeAnalysisContext ctx
...
 		}
 
-		private static ImmutableDictionary<string, string> CreateFixerContext( ImmutableArray<ArgParamBinding> unnamedArgs ) {
+		private static ITypeSymbol GetArgumentType(
+			SyntaxNodeAnalysisContext ctx,
+			ArgumentSyntax argSyntax ) {
```
j3parker, line 187:
```suggestion
			ArgumentSyntax argSyntax
		) {
```

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -123,34 +123,80 @@ SyntaxNodeAnalysisContext ctx
...
+			var baseSymbol = ctx.SemanticModel.GetSymbolInfo( argSyntax.Expression ).Symbol;
+			return baseSymbol switch {
+				IParameterSymbol symbol => symbol.Type,
+				IFieldSymbol symbol => symbol.Type,
+				ILocalSymbol symbol => symbol.Type,
+				_ => baseSymbol.ContainingType,
```
j3parker, line 193: I don't think this is right.... I think you want `ctx.SemanticModel.GetTypeInfo` instead, without this switch.
You can pass an expression like `2 + 2` as an argument... and I think this would crash?

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
```diff
@@ -160,20 +206,20 @@ SyntaxNodeAnalysisContext ctx
...
 			if( expressionType == null || expressionType.Kind == SymbolKind.ErrorType ) {
 				return false;
-			}
-
-			// the current call could be nested inside an expression tree, so
-			// check every call we are nested inside
+			}
+
+			// the current call could be nested inside an expression tree, so
+			// check every call we are nested inside
```
j3parker, line 222: In the future it would be preferable to do whitespace changes in their own PR :)

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
j3parker: I don't think this concern is valid. Inserting argument names won't change which overload you use. It could be that we suggest that you explicitly pick an overload that you didn't want to, but already accidentally had (this would only be in VERY unlikely scenarios I think) **and** you still don't notice... and then the next person assumes you knew what you were doing... but that's kind of a general "problem" with this analyzer, so eh I don't buy it.

Sorry, just thinking out loud. I wrote this comment originally but that doesn't mean I remembered why immediately :)

j3parker: TL;DR: ignore the overload thing, let's do max one diagnostic per function call, and the fixerContext should have whatever args we deem worthy of naming, which may be an incomplete list.

TimothyJCowen: So is it to be removed, or not?

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
TimothyJCowen: I don't know what is happening here... I tried to get rid of these but they don't show up on my local.

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
TimothyJCowen: Arguments are now flagged and reported individually. Your example is now handled.

src/D2L.CodeStyle.Analyzers/Language/RequireNamedArgumentsAnalyzer.cs
TimothyJCowen: Method was removed entirely.

TimothyJCowen:
Change was deemed too complicated (with exclusions and such), and so is being closed.

Issue #518, opened 2019-07-02
Require log messages are constant strings


omsmith:
The messages that are passed to `ILogger`'s logging methods are intended to be constant values. If the string includes details specific to that particular instance of the error, then it is much harder to track error counts, filter out "uninteresting" errors in alerting and other such things. Additional context is meant to be provided through the `Func<T> metadata` parameter.

An analyzer could be made which inspects `IInvocationOperation`s which have a `TargetMethod` matching one of `D2L.LP.Logging.ILogger`'s methods. `SemanticModel`'s `GetConstantValue` can be used to get the value of the `message` argument. If the model is able to return a constant value, all is well, but otherwise should raise a diagnostic.

```csharp
internal sealed class MyClass {

  private static readonly D2L.LP.Logging.ILogger m_log = D2L.LP.Logging.LogManager.GetLogger( ... );

  private static readonly string STATIC_STRING = "static_string";
  private const string CONST_STRING = "const_string";

  void Foo( string myVar ) {
    m_log.Error( "Foo" );
    m_log.Info( CONST_STRING );
    m_log.Warn( CONST_STRING + "Foo" );

    m_log.Error( myVar ); // D2LXXXX: log messages should be constant
    m_log.Error( "Foo" + myVar ); // D2LXXXX: log messages should be constant
    m_log.Error( $"{myVar}" ); // D2LXXXX: log messages should be constant
    m_log.Warn( STATIC_STRING ); // D2LXXXX: log messages should be constant
  }
}
```

PR #687, opened 2021-01-04
Added [Constant] attribute.

TimothyJCowen:
Can be used to fix https://github.com/Brightspace/D2L.CodeStyle/issues/518

- Added D2L.CodeStyle.Annotations.Contract.ConstantAttribute.cs
- Added D2L.CodeStyle.Analyzers.Helpers.ConstantAnalyzer.cs
- Added D2L.CodeStyle.Analyzers.Tests.Specs.ConstantAnalyzer.cs
- Moved common method from Analyzers.ApiUsage.NotNullAnalyzer.cs to Analyzers.ExtensionMethods.cs
  - Changed to extension method on ISymbol


tests/D2L.CodeStyle.Analyzers.Test/Specs/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+		public static void SomeMethodWithOneConstantParameter([Constant] string param1, string param2) { }
+		public static void SomeMethodWithOneOtherConstantParameter(string param1, [Constant] string param2) { }
+		public static void SomeMethodWithTwoConstantParameters([Constant] string param1, [Constant] string param2) { }
+	}
+
+	public sealed class Tests
+	{
+		void Method()
+		{
+			string variable = "This is a variable message";
```
omsmith, line 32: Most non-constant usage of the logger look like

```csharp
m_logger.Error( string.Format( ERROR_RELATIVE_MONTH_DATE_VALUE_INVALID, achievementId ) );
m_logger.Warn( $"The job {jobId} for the org unit {entity.TargetOrgUnitId} has/is being proccessed" );
```

I would personally include tests that cover those cases.

tests/D2L.CodeStyle.Analyzers.Test/Specs/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
...
+		public sealed class ConstantAttribute : Attribute { }
+	}
+}
+#endregion
+
+namespace SpecTests
+{
+	using static Contract;
+
+	public sealed class Types
```
omsmith, line 19: Since this `ConstantAnalyzer` is general, I would include examples with non-`string`s as well

tests/D2L.CodeStyle.Analyzers.Test/Specs/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,69 @@
+﻿// analyzer: D2L.CodeStyle.Analyzers.Helpers.ConstantAnalyzer
+
+using D2L.CodeStyle.Annotations;
+
+#region Relevant Types
+namespace D2L.CodeStyle.Annotations
```
omsmith, line 6: Don't know why you editor isn't applying it properly, but, some spacing issues here:

```csharp
namespace Foo {
	public class Bar {
		public void Method( object parameter ) {
			Method( "foo" );
		}
	}
}
```

omsmith:
I wonder if `[Constant]` should be restricted to parameters of SpectialTypes that can actually have a non-null constant value?

```csharp
public interface IFoo {}
public class Bar {
	void Do( [Constant] IFoo foo ) {}
}
```

TimothyJCowen:
> I wonder if `[Constant]` should be restricted to parameters of SpectialTypes that can actually have a non-null constant value?
> 
> ```cs
> public interface IFoo {}
> public class Bar {
> 	void Do( [Constant] IFoo foo ) {}
> }
> ```

Is there a way I can find this out in code? Like an "IfCanBeConstant" or something? I know that idealistic, but you never know!

omsmith:
Also dunno if @j3parker still has reservations of this approach vs just doing it specifically for logging? Only know what was in that initial slack thread before the break.

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -552,5 +552,15 @@ public static class Diagnostics {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor NonConstantPassedToConstantParameter = new DiagnosticDescriptor(
+			id: "D2L0074",
+			title: "Constant parameter cannot be passed a non-constant value.",
+			messageFormat: "Parameter \"{0}\" is marked with the [Constant] attribute, and so cannot be passed a non-constant value.",
```
j3parker, line 559: Having part of this be customizable @ the parameter defn would be neat. Like why does it need to be constant? etc.
Not in this PR, could just open an issue for later.

src/D2L.CodeStyle.Analyzers/ExtensionMethods.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System.Linq;
+using D2L.CodeStyle.Analyzers.Extensions;
+using Microsoft.CodeAnalysis;
+
+namespace D2L.CodeStyle.Analyzers {
+
+	internal static class ExtensionMethods {
```
j3parker, line 7: We have an extensions folder where we store our extension methods on roslyn types. There's a convention about how its organized by namespace -- could you move this into there?

src/D2L.CodeStyle.Analyzers/ExtensionMethods.cs
```diff
@@ -0,0 +1,27 @@
...
+
+		/// <summary>
+		/// Check if the symbol has a specific attribute attached to it.
+		/// </summary>
+		/// <param name="symbol">The symbol to check for an attribute on</param>
+		/// <param name="attributeClassName">The class of the attribute to check for</param>
+		/// <returns>True if the attribute exists on the symbol, false otherwise</returns>
+		public static bool HasAttribute(
+			this ISymbol symbol,
+			string attributeClassName
```
j3parker, line 17: Stringly comparisons kind of suck, so I'm a bit unexcited about making this a shared extension method. Checks like these should be done at the symbol level to avoid ambiguity.

Types can be renamed:

```cs
using CheatyType = TypeOurAnalyzerExpects;
```

Types can also be ambiguous (same name in different namespaces, or even the same name+namespace but different assembly)

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+			AnalyzeInvocationLikeThing(
+				context,
+				construction.ArgumentList
+			);
+		}
+
+
+		private static void AnalyzeInvocationLikeThing(
+			SyntaxNodeAnalysisContext context,
+			ArgumentListSyntax argumentList ) {
```
j3parker, line 67:
```suggestion
			context.RegisterSyntaxNodeAction(
				ctx => AnalyzeInvocation(
					ctx,
					((InvocationExpressionSyntax)ctx.Node).ArgumentList
				),
				SyntaxKind.InvocationExpression
			);

			context.RegisterSyntaxNodeAction(
				ctx => AnalyzeInvocation(
					ctx,
					((ObjectCreationExpressionSyntax)ctx.Node).ArgumentList
				),
				SyntaxKind.ObjectCreationExpression
			);
		}
		
		private static void AnalyzeInvocation(
			SyntaxNodeAnalysisContext context,
			ArgumentListSyntax argumentList
		) {
```

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,105 @@
...
+			}
+
+			// Get arguments from the list and iterate through them
+			SeparatedSyntaxList<ArgumentSyntax> arguments = argumentList.Arguments;
+			foreach( var argument in arguments ) {
+
+				// Get the associated parameter
+				var parameter = argument.DetermineParameter(
+					context.SemanticModel,
+					allowParams: true );
```
j3parker, line 81:
```suggestion
					allowParams: true
				);
```

j3parker (dismissed):
Left some notes -- looks good!

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
TimothyJCowen: https://github.com/Brightspace/D2L.CodeStyle/issues/696

tests/D2L.CodeStyle.Analyzers.Test/Specs/ConstantAnalyzer.cs
TimothyJCowen: I think spacing is good now.

tests/D2L.CodeStyle.Analyzers.Test/Specs/ConstantAnalyzer.cs
TimothyJCowen: Added integer and boolean.

src/D2L.CodeStyle.Analyzers/ExtensionMethods.cs
TimothyJCowen: I poked around a bit and saw that we actually do this a lot. Immutability does it, NonNull was doing it...
I think this should be its own issue.

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,151 @@
...
+				// We need the symbol
+				IParameterSymbol parameter = context.SemanticModel.GetDeclaredSymbol( parameterSyntax );
+
+				// Parameter is somehow null, so do nothing
+				if( parameter == null ) {
+					continue;
+				}
+
+				// Parameter is not [Constant], so do nothing
+				if( !parameter.HasAttribute( AttributeName ) ) {
```
omsmith, line 68: Check `StatelessFuncAnalyzer` for an easy reference of doing this with symbols.

https://github.com/Brightspace/D2L.CodeStyle/blob/c9bacec35a74b8568721dd25a9dad187a5d1c826/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs#L28

https://github.com/Brightspace/D2L.CodeStyle/blob/c9bacec35a74b8568721dd25a9dad187a5d1c826/src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs#L64

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,151 @@
...
+				SyntaxKind.InvocationExpression
+			);
+
+			context.RegisterSyntaxNodeAction(
+				ctx => AnalyzeArguments(
+					ctx,
+					( (ObjectCreationExpressionSyntax)ctx.Node ).ArgumentList
+				),
+				SyntaxKind.ObjectCreationExpression
+			);
```
omsmith, line 50: This can be simplified by just registering for `SyntaxKind.Argument`

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,151 @@
...
+		public static void CompilationStart(
+			CompilationStartAnalysisContext context
+		) {
+			context.RegisterSyntaxNodeAction(
+				ctx => AnalyzeParameters(
+					ctx,
+					(ParameterListSyntax)ctx.Node
+				),
+				SyntaxKind.ParameterList
+			);
```
omsmith, line 34: Consider registering for `SyntaxKind.Parameter` instead.

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,151 @@
...
+					context.ReportDiagnostic(
+						Diagnostic.Create(
+							descriptor: Diagnostics.InvalidConstantType,
+							location: argument.GetLocation(),
+							messageArgs: type.TypeKind
+						)
+					);
+
+					continue;
+				}
```
omsmith, line 135:
```suggestion
```

This check doesn't really make sense on the argument does it? Made sense on the parameter to ensure the attribute is being used correctly, but for the argument the constant check below seems like it's what we want.

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,151 @@
...
+
+				// Get the base type of the parameter
+				ITypeSymbol type = parameter.Type;
+
+				// Check that the parameter type is not a valid type,
+				// and that it is not a type parameter
+				if( type.SpecialType != SpecialType.None
+				 || type.Kind != SymbolKind.NamedType ) {
```
omsmith, line 78: nit: formatting
```suggestion
				if( type.SpecialType != SpecialType.None
					 || type.Kind != SymbolKind.NamedType
	 	 	 	 ) {
```

But also, I don't think I understand the `type.Kind != SymbolKind.NamedType`. Could you explain?

tests/D2L.CodeStyle.Analyzers.Test/Specs/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,174 @@
...
+using D2L.CodeStyle.Annotations;
+
+#region Relevant Types
+namespace D2L.CodeStyle.Annotations
+{
+	public static class Contract
+	{
+		public sealed class ConstantAttribute : Attribute { }
+	}
+}
```
omsmith, line 12:
```suggestion
namespace D2L.CodeStyle.Annotations.Contract
{
	public sealed class ConstantAttribute : Attribute { }
}

```

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
TimothyJCowen: I added a more descriptive comment.
Basically this is needed because when the parameter is generic (`SomeThing<T>([Constant] T obj) { }`) you can't tell if the type can be constant or not - that needs to be decided when the actual type is determined.

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
TimothyJCowen: I split those into two separate statements so that I could add more explanation.
Basically the same reason as the argument block you asked about, a "NamedType" means it's not generic.

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -216,5 +216,32 @@ string name
...
+		/// <param name="symbol">The symbol to check for an attribute on</param>
+		/// <param name="compilation">The contextual compilation to lookup types</param>
+		/// <param name="attributeMetadataName">The class of the attribute to check for</param>
+		/// <returns>True if the attribute exists on the symbol, false otherwise</returns>
+		public static bool HasAttribute(
+			this ISymbol symbol,
+			Compilation compilation,
+			string attributeMetadataName
+		) {
+			var attributeSymbol = compilation.GetTypeByMetadataName( attributeMetadataName );
```
omsmith, line 233: This is best done once during `CompilationStart`, not within an extension like this for every check. I'd suggest you revert the changes to extracting this extension method to keep the PR focused.

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
omsmith: But that's just confusing / uninteresting as a consumer of an API. It's already been marked as constant and there's nothing I can do about that. As a consumer the interesting thing is whether the value I'm passing is constant or not.

Checking the type the parameter is attached to and raising this diagnostic is just to keep people from applying to a parameter that can never be constant, and thus any usage of it would be an error.

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,136 @@
...
+			// Parameter is not [Constant], so do nothing
+			if( !parameter.HasAttribute( context.Compilation, AttributeName ) ) {
+				return;
+			}
+
+			var type = parameter.Type;
+
+			// If the parameter is a valid type, there are no issues
+			if( type.SpecialType != SpecialType.None
+			) {
```
omsmith, line 65: nit, whitespace

```suggestion
			if( type.SpecialType != SpecialType.None ) {
```

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,136 @@
...
+				return;
+			}
+
+			// If the parameter is a generic type, there aren't issues yet...
+			// There may be issues when the method is being called, but
+			// those are handled in the argument analyzer
+			if( type.Kind != SymbolKind.NamedType
+			) {
+				return;
+			}
```
omsmith, line 75: The terminology of "generic type" here isn't quite write. `IEnumerable<T>` is a generic type. `T` is a type parameter.

```suggestion
			// If the parameter's type is a type parameter then whether
			// it can be constant depends on the type that gets filled and
			// we can't determine that here
			if( type.Kind == SymbolKind.TypeParameter ) {
				return;
			}
```

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,136 @@
...
+
+			// Parameter is not [Constant], so do nothing
+			if( !parameter.HasAttribute( context.Compilation, AttributeName ) ) {
+				return;
+			}
+
+			var type = parameter.Type;
+
+			// If the parameter is a valid type, there are no issues
```
omsmith, line 63:
```suggestion
			// Special types (bool, enum, int, string, etc.) are the only types which might be
			// constant, aside from type parameters
```

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,167 @@
...
+		/// <param name="attributeSymbol">The symbol of the attribute</param>
+		/// <returns>True if the attribute exists on the symbol, false otherwise</returns>
+		public static bool HasAttribute(
+			ISymbol symbol,
+			ISymbol attributeSymbol
+		) {
+			if( attributeSymbol == null ) {
+				return false;
+			}
+
```
omsmith, line 158: constantAttribute should never be null when we're using it here.

```suggestion
```

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,167 @@
...
+				ctx => AnalyzeArguments(
+					ctx,
+					(ArgumentSyntax)ctx.Node,
+					constantAttribute
+				),
+				SyntaxKind.Argument
+			);
+		}
+
+		private static void AnalyzeParameters(
```
omsmith, line 50: nit: `AnalyzeParameter` (not plural anymore)

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,167 @@
...
+
+		private static void AnalyzeArguments(
+			SyntaxNodeAnalysisContext context,
+			ArgumentSyntax argument,
+			ISymbol constantAttribute
+		) {
+			// Get the associated parameter
+			var parameter = argument.DetermineParameter(
+				context.SemanticModel,
+				allowParams: true
```
omsmith, line 101: I think we can probably `allowParams: false` here?

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,167 @@
...
+			}
+
+			var type = parameter.Type;
+
+			// Special types (bool, enum, int, string, etc) are the only types
+			// which might be constant, aside from type parameters filled with
+			// special types
+			if( type.SpecialType != SpecialType.None ) {
+				return;
+			}
```
omsmith, line 74: Looking at `SpecialType` again, there's a number of them that aren't able to be constant.

```suggestion
			switch( type.SpecialType ) {
				case SpecialType.System_Enum:
				case SpecialType.System_Boolean:
				case SpecialType.System_Char:
				case SpecialType.System_SByte:
				case SpecialType.System_Byte:
				case SpecialType.System_Int16:
				case SpecialType.System_UInt16:
				case SpecialType.System_Int32:
				case SpecialType.System_UInt32:
				case SpecialType.System_Int64:
				case SpecialType.System_UInt64:
				case SpecialType.System_Decimal:
				case SpecialType.System_Single:
				case SpecialType.System_Double:
				case SpecialType.System_String:
					return;
				default:
					break;
			}
```

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -552,5 +552,24 @@ public static class Diagnostics {
...
+			category: "Safety",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true,
+			description: "The method being called has declared that this parameter must receive a constant, but a non-constant value is being passed."
+		);
+
+		public static readonly DiagnosticDescriptor InvalidConstantType = new DiagnosticDescriptor(
+			id: "D2L0075",
+			title: "Invalid data type marked as [Constant].",
+			messageFormat: "The [Constant] attribute cannot be used on \"{0}\" types, because they cannot produce compile-time constants.",
```
omsmith, line 569:
```suggestion
			messageFormat: "The [Constant] attribute cannot be used on \"{0}\" types, because they cannot be compile-time constants.",
```

omsmith (dismissed):

src/D2L.CodeStyle.Analyzers/Helpers/ConstantAnalyzer.cs
```diff
@@ -0,0 +1,169 @@
...
+			context.ConfigureGeneratedCodeAnalysis( GeneratedCodeAnalysisFlags.Analyze | GeneratedCodeAnalysisFlags.ReportDiagnostics );
+			context.RegisterCompilationStartAction( CompilationStart );
+		}
+
+		public static void CompilationStart(
+			CompilationStartAnalysisContext context
+		) {
+			var constantAttribute = context.Compilation.GetTypeByMetadataName(
+				"D2L.CodeStyle.Annotations.Contract.ConstantAttribute"
+			);
```
j3parker, line 29:
```suggestion
			);

			// The D2L.CodeStyle.Annotations reference is optional
			if ( constantAttribute == null || constantAttribute .Kind == SymbolKind.ErrorType ) {
				return;
			}
```

j3parker: This won't happen often, like if you transitively reference D2L.CodeStyle.Annotations (which you would have to to call a [Constant] method) then it won't get hit... but this isn't required.

j3parker (dismissed):
Nice!

j3parker (approved):

Issue #689, opened 2021-01-12
Allow test cases to support commas.


TimothyJCowen:
Test case diagnostics are defined in the test files as such:
```cs
static /* SomeDiagnostic(structure, SomeStruct) */ SomeStruct /**/ some_field;
```
in which case commas are used to delineate between diagnostic arguments.

We should be able to support cases where commas are included in the diagnostic, by escaping the commas as such:
```cs
static /* SomeDiagnostic(interface, SomeGenericInterface<int\\, int>) */ SomeGenericInterface<int, int> /**/ some_field;
```

PR #691, opened 2021-01-12
Allow test cases to support commas. #689

TimothyJCowen:
Will help make an upcoming issue more clear.

Closes https://github.com/Brightspace/D2L.CodeStyle/issues/689

### DIFF
--- a/tests/D2L.CodeStyle.Analyzers.Test/SpecRunner.cs
+++ b/tests/D2L.CodeStyle.Analyzers.Test/SpecRunner.cs
@@ -339,7 +339,7 @@ namespace D2L.CodeStyle.Analyzers {
 					yield return new DiagnosticExpectation(
 						name: name,
 						arguments: Regex.Split( arguments, @"(?<!\\\\)," )
-							.Select( arg => arg.Replace( "\\\\", string.Empty ) )
+							.Select( arg => arg.Replace( "\\\\,", "," ) )
 							.Select( RemoveSingleLeadingSpace )
 							.ToImmutableArray()
 					);
--- a/tests/D2L.CodeStyle.Analyzers.Test/SpecRunner.cs
+++ b/tests/D2L.CodeStyle.Analyzers.Test/SpecRunner.cs
@@ -339,9 +339,9 @@ namespace D2L.CodeStyle.Analyzers {
 					yield return new DiagnosticExpectation(
 						name: name,
 						arguments: Regex.Split( arguments, @"(?<!\\\\)," )
-						.Select( arg => arg.Replace( "\\\\", string.Empty ) )
-						.Select( RemoveSingleLeadingSpace )
-						.ToImmutableArray()
+							.Select( arg => arg.Replace( "\\\\", string.Empty ) )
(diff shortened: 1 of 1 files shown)


tests/D2L.CodeStyle.Analyzers.Test/SpecRunner.cs
```diff
@@ -338,7 +338,10 @@ ImmutableArray<string> arguments
 
 					yield return new DiagnosticExpectation(
 						name: name,
-						arguments: arguments.Split( ',' ).Select( RemoveSingleLeadingSpace ).ToImmutableArray()
+						arguments: Regex.Split( arguments, @"(?<!\\\\)," )
+						.Select( arg => arg.Replace( "\\\\", string.Empty ) )
+						.Select( RemoveSingleLeadingSpace )
+						.ToImmutableArray()
```
omsmith, line 344: nit

```suggestion
							.Select( arg => arg.Replace( "\\\\", string.Empty ) )
							.Select( RemoveSingleLeadingSpace )
							.ToImmutableArray()
```

tests/D2L.CodeStyle.Analyzers.Test/SpecRunner.cs
```diff
@@ -338,7 +338,10 @@ ImmutableArray<string> arguments
 
 					yield return new DiagnosticExpectation(
 						name: name,
-						arguments: arguments.Split( ',' ).Select( RemoveSingleLeadingSpace ).ToImmutableArray()
+						arguments: Regex.Split( arguments, @"(?<!\\\\)," )
+							.Select( arg => arg.Replace( "\\\\", string.Empty ) )
```
omsmith, line 342: I'm struggling to wrap my head around whether this should actually be `\\\\` or `\\`. Can you share an example of an expectation you've got working with this?

omsmith, line 342:
```suggestion
						arguments: Regex.Split( arguments, @"(?<!\\\\)," )
							.Select( arg => arg.Replace( "\\\\,", "," ) )
```

TimothyJCowen: I've tested with this and it works perfectly:
```cs
/* NonImmutableTypeHeldByImmutable(interface, SpecTests.Types.SomeGenericInterface<int\\, int>, ) */
```

TimothyJCowen: Is the reason for this just to make it only catch escaped commas?

omsmith: Right

omsmith: So given it's a comment, you don't actually need to be escaping the `\`. The following would be fine, and the regexes switched to `\\`. Sound reasonable?

`/* NonImmutableTypeHeldByImmutable(interface, SpecTests.Types.SomeGenericInterface<int\, int>, ) */`

omsmith (approved):

PR #698, opened 2021-01-28
Immutability: analyzer assumes initializers are the only writes to readonly members #319

TimothyJCowen:
- Added method to get all possible initializations from within constructors
- Changed single-initializer structures to use multiple initializers
- Added supporting tests

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
@@ -133,15 +133,14 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		}
 
 		private bool CheckField( DiagnosticSink diagnosticSink, IFieldSymbol field ) {
-			if( field.IsImplicitlyDeclared ) {
+			if ( field.IsImplicitlyDeclared ) {
 				// These correspond to auto-properties. That case gets handled
 				// in CheckProperty instead.
 				return true;
 			}
 
-			var decl = field.DeclaringSyntaxReferences
-			   .Single()
-			   .GetSyntax() as VariableDeclaratorSyntax;
+			var decl = field.DeclaringSyntaxReferences.Single()
+				.GetSyntax() as VariableDeclaratorSyntax;
 
 			var type = decl.FirstAncestorOrSelf<VariableDeclarationSyntax>().Type;
 
@@ -161,14 +160,14 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 
 
 		private bool CheckProperty( DiagnosticSink diagnosticSink, IPropertySymbol prop ) {
-			if( prop.IsIndexer ) {
+			if ( prop.IsIndexer ) {
 				// Indexer properties are just glorified method syntax and
 				// don't hold state.
 				// https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/indexers/
 				return true;
 			}
 
-			if( prop.IsImplicitlyDeclared ) {
+			if ( prop.IsImplicitlyDeclared ) {
 				// records have implicitly declared properties like
 				// EqualityContract which are OK but don't have a
 				// PropertyDeclarationSyntax etc.
@@ -180,7 +179,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				prop.DeclaringSyntaxReferences.Single().GetSyntax()
 			);
 
-			if( !propInfo.IsAutoImplemented ) {
+			if ( !propInfo.IsAutoImplemented ) {
 				// Properties that are auto-implemented have an implicit
 				// backing field that may be mutable. Otherwise, properties are
 				// just sugar for getter/setter methods and don't themselves
@@ -213,7 +212,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		) {
 			var immutable = true;
 
-			if( !isReadOnly ) {
+			if ( !isReadOnly ) {
 				diagnosticSink(
 					Diagnostic.Create(
 						Diagnostics.MemberIsNotReadOnly,
@@ -274,21 +273,21 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			// Retrieve a list of assignment expressions from any constructors
 			// within the class
 			var assignmentExpressions = ( memberSymbol.ContainingSymbol as INamedTypeSymbol )!
-			   .Constructors
-			   .Where( constructorSymbol => !constructorSymbol.IsImplicitlyDeclared )
-			   .Select( constructorSymbol => constructorSymbol.DeclaringSyntaxReferences
-				           .Single()
-				           .GetSyntax() )
-			   .SelectMany( constructorSyntax => constructorSyntax.DescendantNodes() )
-			   .OfType<AssignmentExpressionSyntax>();
+				.Constructors
+				.Where( constructorSymbol => !constructorSymbol.IsImplicitlyDeclared )
+				.Select( constructorSymbol => constructorSymbol.DeclaringSyntaxReferences
+					.Single()
+					.GetSyntax() )
+				.SelectMany( constructorSyntax => constructorSyntax.DescendantNodes() )
+				.OfType<AssignmentExpressionSyntax>();
 
 			// Add any assignments which act on the specific field/property
 			// to the list
 			return assignmentExpressions
-			   .Where( assignmentSyntax => IsAnAssignmentTo( assignmentSyntax, memberSymbol ) )
-			   .Select( assignmentSyntax => assignmentSyntax.Right )
-			   .Append( initializer )
-			   .ToImmutableArray();
+				.Where( assignmentSyntax => IsAnAssignmentTo( assignmentSyntax, memberSymbol ) )
+				.Select( assignmentSyntax => assignmentSyntax.Right )
+				.Append( initializer )
+				.ToImmutableArray();
 		}
 
 		private bool IsAnAssignmentTo(
@@ -300,7 +299,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			);
 
 			var leftSideSymbol = semanticModel.GetSymbolInfo( assignmentSyntax.Left )
-			   .Symbol;
+				.Symbol;
 
 			return SymbolEqualityComparer.Default.Equals(
 				memberSymbol,
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
@@ -146,9 +146,8 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 
 			var type = decl.FirstAncestorOrSelf<VariableDeclarationSyntax>().Type;
 
-			// Get all possible initializations of the field
-			ImmutableArray<ExpressionSyntax> initializers
-				= GetInitializers( field, decl.Initializer?.Value );
+			// Get all possible assignments of the field
+			var assignments = GetAssignments( field, decl.Initializer?.Value );
 
 			return CheckFieldOrProperty(
 				diagnosticSink,
@@ -157,7 +156,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				isReadOnly: field.IsReadOnly,
 				typeSyntax: type,
 				nameSyntax: decl.Identifier,
-				initializers: initializers
+				assignments
 			);
 		}
 
@@ -190,9 +189,8 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				return true;
 			}
 
-			// Get all possible initializations of the property
-			ImmutableArray<ExpressionSyntax> initializers
-				= GetInitializers( prop, propInfo.Initializer );
+			// Get all possible assignments of the property
+			var assignments = GetAssignments( prop, propInfo.Initializer );
 
 			return CheckFieldOrProperty(
 				diagnosticSink,
@@ -201,7 +199,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				isReadOnly: propInfo.IsReadOnly,
 				typeSyntax: propInfo.TypeSyntax,
 				nameSyntax: propInfo.Identifier,
-				initializers: initializers
+				assignments
 			);
 		}
 
@@ -212,7 +210,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			bool isReadOnly,
 			TypeSyntax typeSyntax,
 			SyntaxToken nameSyntax,
-			ImmutableArray<ExpressionSyntax> initializers
+			ImmutableArray<ExpressionSyntax> assignments
 		) {
 			var immutable = true;
 
@@ -230,16 +228,16 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				immutable = false;
 			}
 
-			// Check all initializations of the member for immutability
-			foreach( ExpressionSyntax initializer in initializers ) {
+			// Check all assignments of the member for immutability
+			foreach( var assignment in assignments ) {
 				var stuff = GetStuffToCheckForMember(
 					type,
 					typeSyntax,
-					initializer
+					assignment
 				);
 
 				if( stuff == null ) {
-					// Null is a signal that there is nothing further that needs to
+					// null is a signal that there is nothing further that needs to
 					// be checked
 					continue;
 				}
@@ -264,26 +262,25 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 
 
 		/// <summary>
-		/// A readonly field or property may be re-initialized in constructors.
-		/// This method finds all initializations of the field/property.
+		/// A readonly field or property may be re-assigned in constructors.
+		/// This method finds all assignments of the field/property.
 		/// </summary>
 		/// <param name="symbol">The field/property to search for</param>
-		/// <param name="initialInitializer">The initialization with the declaration</param>
-		/// <returns>An immutable list of initializations</returns>
-		private ImmutableArray<ExpressionSyntax> GetInitializers(
+		/// <param name="initializer">The initialization with the declaration</param>
+		/// <returns>An immutable list of assignments</returns>
+		private ImmutableArray<ExpressionSyntax> GetAssignments(
 			ISymbol symbol,
-			ExpressionSyntax initialInitializer
+			ExpressionSyntax initializer
 		) {
-			List<ExpressionSyntax> initializers = new() {initialInitializer};
+			List<ExpressionSyntax> assignments = new() {initializer};
 
 			// If the containing symbol isn't a named type (class) then there
 			// will be no constructors
 			if( symbol.ContainingSymbol is not INamedTypeSymbol namedTypeSymbol ) {
-				return initializers.ToImmutableArray();
+				return assignments.ToImmutableArray();
 			}
 
-			// Iterate through all constructors in the class
-			foreach( IMethodSymbol methodSymbol in namedTypeSymbol.Constructors ) {
+			foreach( var methodSymbol in namedTypeSymbol.Constructors ) {
 				// Make sure the syntax of the constructor exists
 				if( methodSymbol.DeclaringSyntaxReferences.Length == 0 ) {
 					continue;
@@ -299,11 +296,9 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 					continue;
 				}
 
-				// Retrieve the semantic model from the compilation
-				SemanticModel semanticModel = m_compilation.GetSemanticModel( methodSyntax.SyntaxTree );
+				var semanticModel = m_compilation.GetSemanticModel( methodSyntax.SyntaxTree );
 
-				// Iterate through every statement in the constructor
-				foreach( StatementSyntax statement in methodSyntax.Body.Statements ) {
+				foreach( var statement in methodSyntax.Body.Statements ) {
 					// If the current statement is not an expression,
 					// we don't care
 					if( statement is not ExpressionStatementSyntax expression ) {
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
 				diagnosticSink( diagnostic );
 				immutable = false;
 			}
 
 			return immutable;
 		}
 
+
+		/// <summary>
+		/// A readonly field or property may be re-initialized in constructors.
```
j3parker, line 267: Nit: I get what you mean here by initializers but it's a bit confusing because initializers are a specific thing in the language. Maybe reword this to writes? Like `GetWrites()` etc. which sounds more general/accurate.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
+					initializer
+				);
 
-			if ( stuff == null ) {
-				// null is a signal that there is nothing further that needs to
-				// be checked
-				return immutable;
-			}
+				if( stuff == null ) {
+					// Null is a signal that there is nothing further that needs to
```
j3parker, line 242:
```suggestion
					// null is a signal that there is nothing further that needs to
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
+		/// <summary>
+		/// A readonly field or property may be re-initialized in constructors.
+		/// This method finds all initializations of the field/property.
+		/// </summary>
+		/// <param name="symbol">The field/property to search for</param>
+		/// <param name="initialInitializer">The initialization with the declaration</param>
+		/// <returns>An immutable list of initializations</returns>
+		private ImmutableArray<ExpressionSyntax> GetInitializers(
+			ISymbol symbol,
+			ExpressionSyntax initialInitializer
```
j3parker, line 275: heh ya and rename this to initializer :)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
+		) {
+			List<ExpressionSyntax> initializers = new() {initialInitializer};
+
+			// If the containing symbol isn't a named type (class) then there
+			// will be no constructors
```
j3parker, line 280: Members can exist in other INamedTypeSymbols, like records or structs.
What is an example of a field that is not an INamedTypeSymbol?

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
+			if( symbol.ContainingSymbol is not INamedTypeSymbol namedTypeSymbol ) {
+				return initializers.ToImmutableArray();
+			}
+
+			// Iterate through all constructors in the class
```
j3parker, line 285:
```suggestion
```

Yes, its a for loop over the constructors, the code already says that :)

j3parker: https://blog.d2l.dev/2019/08/29/write-good-comments/

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
+			foreach( IMethodSymbol methodSymbol in namedTypeSymbol.Constructors ) {
+				// Make sure the syntax of the constructor exists
+				if( methodSymbol.DeclaringSyntaxReferences.Length == 0 ) {
+					continue;
+				}
+
+				// Retrieve the constructor syntax
+				var methodSyntax = methodSymbol.DeclaringSyntaxReferences
+				                               .Single()
```
j3parker, line 294: We don't align `.`s like this iirc

j3parker: If you put `.DeclaringSyntaxReferences on the next line then they will line up, that's pretty common.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
+				                               .GetSyntax() as ConstructorDeclarationSyntax;
+
+				// Make sure the constructor actually contains a body
```
j3parker, line 297: Are there constructors without a body?

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
+				if( methodSyntax?.Body?.Statements == null ) {
+					continue;
+				}
+
+				// Retrieve the semantic model from the compilation
```
j3parker, line 302:
```suggestion
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
TimothyJCowen: The check is because it is `INamedTypeSymbol` that has the `.Constructors` member.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
+				SemanticModel semanticModel = m_compilation.GetSemanticModel( methodSyntax.SyntaxTree );
+
+				// Iterate through every statement in the constructor
```
j3parker, line 305:
```suggestion
```

We don't need comments that just rephrase a line of code as english.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
TimothyJCowen: The IDE does. I've never seen it done a different way.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +230,112 @@ ExpressionSyntax initializer
...
+				foreach( StatementSyntax statement in methodSyntax.Body.Statements ) {
```
j3parker, line 306:
```suggestion
				foreach( var statement in methodSyntax.Body.Statements ) {
```

Can you go though the rest of the change and do this one? In the analyzer code we use `var` by default.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
TimothyJCowen: You wouldn't expect so, but the IDE was warning me of potential `NullReferenceException`s so I figured it's best to listen to it and be safe.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Actually, this doesn't work:

```cs
public MyType() {
  if ( condition ) // x
  { // y
    m_foo = bar; // z
  }
}
```

The statement `z` isn't in `methodSyntax.Body.Statements`, it's a child statement of the block `y` which is attached to the (if) statement `x`.

There is an easier way to do this than nested loops though (you'd need recursion/a stack to fix this up):

```cs
var writes = namedTypeSymbol
    .Constructors
    .SelectMany( c => c.DescendentNodes() )
    .OfType<AssignmentExpressionSyntax>();
```

It's not as efficient as it could be, because there are cases where we would never find a descendent `AssignmentExpressionSyntax` (like inside of a `case` pattern, like you can't do `case (m_foo = bar):`) but it would be hard to actually implement that and possible fragile (the language can change).

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Resharper? Mine doesn't 🤔

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Hm, I wouldn't listen to the IDE then.

The problem with quietly backing away like this in the analyzer when we find unexpected syntax is it creates room for bugs. The code here is saying "well, if it doesn't have a body I guess stuff is fine.". What if there is a form of constructor that doesn't have a body but does write to fields? It doesn't seem likely to me, but I would rather we crashed with a NullReferenceException :)

j3parker: In cases where we are less sure we have made diagnostics, e.g.: https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Diagnostics.cs#L473-L481

This one makes sense because C# may add a new type kind in the future.

For this case, it seems unlikely to me that there is going to be new syntax for bodyless constructors (I could be wrong, maybe it exists now!) so I'm happy with a NullReferenceException (its simpler.)

omsmith: I believe it's `Body` vs `ExpressionBody` (arrow syntax). They're mutually exclusive and the other is null.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -249,33 +339,29 @@ ExpressionSyntax initializer
...
 		/// <param name="memberType">The type of the field/property</param>
 		/// <param name="memberTypeSyntax">The syntax for the type of the field/property</param>
-		/// <param name="initializer">The initializer syntax for the field/property (possibly null)</param>
+		/// <param name="assignment">The assignment syntax for the field/property (possibly null)</param>
 		/// <returns>null if no checks are needed, otherwise a bunch of stuff.</returns>
 		private (ITypeSymbol, ImmutableTypeKind, Func<Location>)? GetStuffToCheckForMember(
 			ITypeSymbol memberType,
 			TypeSyntax memberTypeSyntax,
-			ExpressionSyntax initializer
+			ExpressionSyntax assignment
```
j3parker, line 347: This one actually is the initializer :)

Looking at this method in isolation, if I were to call it, I'd be wondering "which assignment"?
When it's called `initializer` I know exactly what it means and how to call it.

TimothyJCowen: It's actually all of them. This method gets called for each assignment, not just the initializer. See `CheckFieldOrProperty(...)`.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
omsmith: So that is to say, we need to handle `ExpressionBody`.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Yeah, so this is similar to the `NullReferenceException`. There are two ways to deal with the cast:

1. Assert that it always exists and suffer an `InvalidCastException` if you're wrong (which will trigger a bug report against the analyzers)
2. Handle the failure to cast somehow

If we do (2) and just return an empty array we are saying that there are no extra writes for non-`INamedTypeSymbol`... so I'm wondering if there is any case that can happen? Does returning `[]` make sense in all those cases? Is there a chance that we miss assignments by doing this, which would be a bug?

If we think `INamedTypeSymbol` is the only option here we should do option (1)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Are expression bodies supported for constructors?

j3parker: Oh damn they are.

Aside: we should have test cases for the destructuring assignment, that might be its own case I'm not sure. Probably bettter as a follow-up PR though.

j3parker: `foo.Body ?? foo.ExpressionBody`

omsmith: > Oh damn they are

For specific example: https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/statements-expressions-operators/expression-bodied-members#constructors

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Ohh yes good stuff.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +228,109 @@ ExpressionSyntax initializer
...
+					typeToCheck,
+					checkKind,
+					getLocation,
+					out var diagnostic
+				) ) {
+					continue;
+				}
 
-			if( !m_context.IsImmutable( typeToCheck, checkKind, getLocation, out var diagnostic ) ) {
 				diagnosticSink( diagnostic );
```
j3parker, line 256: So we'll output multiple diagnostics here potentially.... I like it 👍🏼

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +228,109 @@ ExpressionSyntax initializer
...
+		) {
+			List<ExpressionSyntax> assignments = new() {initializer};
+
+			// If the containing symbol isn't a named type (class) then there
+			// will be no constructors
+			if( symbol.ContainingSymbol is not INamedTypeSymbol namedTypeSymbol ) {
+				return assignments.ToImmutableArray();
+			}
+
+			foreach( var methodSymbol in namedTypeSymbol.Constructors ) {
```
j3parker, line 283: It got lost in the shuffle, but: https://github.com/Brightspace/D2L.CodeStyle/pull/698#discussion_r566178258

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +228,87 @@ ExpressionSyntax initializer
...
+			// If the containing symbol isn't a named type (class) then there
+			// will be no constructors
+			if( symbol.ContainingSymbol is not INamedTypeSymbol namedTypeSymbol ) {
+				return assignments.ToImmutableArray();
+			}
+
+			// Retrieve syntax of all constructors
+			foreach(
+				var constructorSyntax in
+				from syn in
```
j3parker, line 286: Ah, please don't use this syntax, we don't really use it at D2L.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -219,28 +228,87 @@ ExpressionSyntax initializer
...
+					from sym in namedTypeSymbol.Constructors
+					where sym.DeclaringSyntaxReferences.Length != 0
+					select sym.DeclaringSyntaxReferences
+					          .Single()
+					          .GetSyntax() as ConstructorDeclarationSyntax
+				where syn != null
+				select syn
```
j3parker, line 293: The DeclaringSyntaxReference for any non-implicit (I think `.IsImplicit` is on the symbol) constructor is always a `ConstructorDeclarationSyntax`, no need for the `where`

j3parker: http://sourceroslyn.io/#Microsoft.CodeAnalysis/Symbols/ISymbol.cs,5dd9c88c9828d99d,references

You can put this in a `.Where` to filter those out.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -274,38 +274,50 @@ ExpressionSyntax initializer
...
+			// Retrieve a list of assignment expressions from any constructors
+			// within the class
+			var assignmentExpressions = namedTypeSymbol
+			                           .Constructors
+			                           .Where( sym => !sym.IsImplicitlyDeclared )
+			                           .Select( sym => sym.DeclaringSyntaxReferences
+			                                              .Single()
+			                                              .GetSyntax() )
+			                           .SelectMany( syn => syn.DescendantNodes() )
+			                           .OfType<AssignmentExpressionSyntax>();
```
j3parker, line 303: Ok, this looks good, a few things:

* nit: the right alignment of the `.` here is wrong. Is this resharper, or did you configure your editor to do this? Your editor should [be doing this](https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs#L34-L41). It's not a big deal, it's just that if everyone disagrees then every PR is going to constantly fight between styles. Consistency is key.
* It's probably fine to combine the and `Select`+`Where` and `Select` from below on to the end of here, but I would combine the `Select`+`Where` into a single `Select` with a helper, e.g.

```cs
.Where( ae => IsAnAssignmentTo( ae, symbol ) )
.Select( ae => ae.Right );
```

The first `Select` that creates an anonymous type to coordinate stuff between the next `Where` and `Select` makes it hard to read.

* In `IsAnAssignmentTo`, instead of passing the semantic model down, grab it each time based on the passed in assignment expression. What we're doing here I think has a bug, because `partial` classes means that the semantic model for the field/property may not have include the constructor that writes to it (because it could be in a different file/document)
* nit: it's a bit weird to create a single element `List`, compose it into an `IEnumerable` with `AddRange` and then convert it to `ImmutableArray`. You can do `assignments.Append( initializer ).ToImmutableArray()` to achieve the same end result without the extra allocation.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -274,38 +274,50 @@ ExpressionSyntax initializer
...
+			var semanticModel = m_compilation.GetSemanticModel(
+				symbol.DeclaringSyntaxReferences.Single().GetSyntax().SyntaxTree
+			);
+
+			// Retrieve a list of assignment expressions from any constructors
+			// within the class
+			var assignmentExpressions = namedTypeSymbol
+			                           .Constructors
+			                           .Where( sym => !sym.IsImplicitlyDeclared )
+			                           .Select( sym => sym.DeclaringSyntaxReferences
```
j3parker, line 299: one more nit: we have `symbol` in this method that refers to a property, `namedTypeSymbol` which refers to its containing type, and `sym` here which refers to a constructor... we should probably choose better names for readability (member/containingType/constructor (or just c inside these lambdas))

tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutabilityAnalyzer.cs
```diff
@@ -54,6 +55,40 @@ public interface RegularInterface { }
...
+
+		[Immutable]
+		public sealed class SomeOtherOtherClassWithConstructor {
+			public readonly RegularInterface m_interface = new Good();
+
+			public SomeClassWithConstructor() {
+				m_interface = new Good();
+			}
+		}
+		#endregion
```
j3parker, line 90: Other test cases that would be nice:

* one with a more complicated constructor, like one with writes inside an if statement (to catch the previous bug that `DescendantNodes()` fixed
* expression-bodies constructors (`public Mytype() => ...`)
* destructuring assignment

The last two might not work yet, and if so then they can be implemented in a follow-up (this PR has had enough back and forth :) ) I got this example from a Microsoft blog that hits the last two cases:

```cs
public record Person { 
    public string FirstName { get; init; } 
    public string LastName { get; init; }

    public Person(string firstName, string lastName) 
      => (FirstName, LastName) = (firstName, lastName);
}

Comment (j3parker) [reply tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutabilityAnalyzer.cs:90]: Maybe throw this in as a test case, it should pass now:

```cs
public record Person(string FirstName, string LastName);
```

It's equivalent to the previous, but it's constructor is implicit. We shouldn't analyze it's constructor, but that's fine because the writes inside of it are always safe.

j3parker (dismissed):

j3parker (approved):

PR #709, opened 2021-02-03
Rename ConstantAnalyzer to ConstantAttributeAnalyzer and put it in the right folder

j3parker:


j3parker:
- Helpers looks like helper code; other analyzers like this go under ApiUsage
- Added "attribute" in the name to make it more obvious what this is

omsmith (approved):

PR #710, opened 2021-02-16
remove ImmutableGenericArgumentAnalyzer

omsmith:
This is largely covered via the work that ImmutabilityAnalyzer does.
IGAA would also check for implied immutabilty requirements through type
constraints, but only for constraints on the parameter itself. For
example, the following would raise a diagnostic, requiring T be marked
[Immutable]:

    interface IFoo<[Immutable] S> {}

    class Foo<T> where T : IFoo<T> {}

While this is interesting, it is only useful for recursive definitions,
which are very rare. A more useful version might cover the following,
requiring U be marked [Immutable]:

    interface IFoo<[Immutable] S> {}

    class Foo<T, U> where T : IFoo<U> {}

Given this is not captured, it's utility is minimal and can be ignored.
Such a diagnostic is a helper, but does not add safety.


omsmith:
Specifically, it's covered by the "types passed to `[Immutable]` type parameters must be immutable" checks. So, it's just covered in the broader sense. That does meant the diagnostic is a little less specific, but, :shrug:

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityScope.cs
```diff
@@ -1,10 +0,0 @@
-﻿namespace D2L.CodeStyle.Analyzers.Immutability {
-	/// <summary>
-	/// Determines what kind of immutable a type has.
-	/// </summary>
-	internal enum ImmutabilityScope {
```
j3parker, line 5: Woohoo

j3parker: (important milestone)

tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutabilityAnalyzer.cs
```diff
@@ -210,6 +218,15 @@ public sealed class AnalyzedClassMarkedImmutableExtendingImmutableBaseClass : Ty
 	public sealed class AnalyzedClassMarkedImmutableExtendingImmutableClass : Types.SomeImmutableClass { }
 
 	[Immutable]
+	public sealed class AnalyzedClassWithNonImmutableTypeParameterImplementingImmutableInterfaceParameter<T, U> : Types.SomeImmutableGenericInterfaceRestrictingT</* TypeParameterIsNotKnownToBeImmutable(T) */ T /**/, U> { }
+
+	public sealed class NonImmutableClassWithNonImmutableTypeParameterImplementingImmutableBaseClassParameter<T> : Types.SomeImmutableGenericBaseClassRestrictingT</* TypeParameterIsNotKnownToBeImmutable(T) */ T /**/> { }
+	public sealed class NonImmutableClassWithNonImmutableTypeParameterImplementingImmutableBaseClassParameter<T> : Types.SomeMutableClassWithImmutableTypeParameter</* TypeParameterIsNotKnownToBeImmutable(T) */ T /**/> { }
```
omsmith, line 224: I wasn't sure that we covered this case, but we do, and I suppose it's not a terrible pattern.

It's basically equivalent to using `[Immutable]` on a method parameter.

j3parker (approved):

PR #714, opened 2021-02-26
Allow method returns to be marked as [Constant].

TimothyJCowen:
- Will allow methods which determine which constant value to return between multiple.
- Added analyzer to ensure return statement of [Constant] method is a compile-time constant value.
- Added supporting tests.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/ConstantAttributeAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/ConstantAttributeAnalyzer.cs
@@ -12,7 +12,8 @@ namespace D2L.CodeStyle.Analyzers.ApiUsage {
 		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics =>
 			ImmutableArray.Create(
 				Diagnostics.NonConstantPassedToConstantParameter,
-				Diagnostics.InvalidConstantType
+				Diagnostics.InvalidConstantType,
+				Diagnostics.NonConstantReturnedFromConstantMethod
 			);
 
 		public override void Initialize( AnalysisContext context ) {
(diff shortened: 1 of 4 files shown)


omsmith:
So where would this be used - something like this?

```csharp
[Constant] private void GetMessage( object thing ) {
  return thing switch {
    object => "Bad Object",
    int => "Bad number",
    _ => throw new Exception();
  }; 
}

**Comment by TimothyJCowen** [comment]
> So where would this be used - something like this?
> 
> ```cs
> [Constant] private string GetMessage( object thing ) {
>   return thing switch {
>     object => "Bad Object",
>     int => "Bad number",
>     _ => throw new Exception();
>   }; 
> }
> ```

Exactly. Anywhere that you might want to choose between multiple messages, all of which are constant, but want its own method for whatever reason.

File: src/D2L.CodeStyle.Analyzers/ApiUsage/ConstantAttributeAnalyzer.cs [hunk 1]
```diff
@@ -52,6 +62,50 @@ CompilationStartAnalysisContext context
...
+			SyntaxNodeAnalysisContext context,
+			ReturnStatementSyntax returnSyntax,
+			ISymbol constantAttribute
+		) {
+			// Return statement has no expression, so do nothing
+			if( returnSyntax.Expression == null ) {
+				return;
+			}
+
+			var methodSyntax = returnSyntax.Parent?.Parent;
```
j3parker, line 75: This doesn't work... `return`s can be nested pretty deeply.

```cs
if ( ... ) {
  if ( ... ) {
    if ( ... ) {
      return ...;
```

You would need something like `.Ancestors().OfType<MethodDeclarationSyntax>()` instead.

PR #715, opened 2021-03-01
Don't treat Lazy<T> as conditionally immutable

j3parker:
Although this is arguably OK (given how we define immutability) it's working against us right now.

We are going to consolidate on `ContextFreeLazy<T>` in circumstances where this is necessary which integrates with `InstanceContextBlocker` to ensure initialization is (customer-) context-free.


omsmith (approved):
Do you already have lms-side changes ready?

j3parker:
> Do you already have lms-side changes ready?

WIP

TimothyJCowen (approved):

PR #718, opened 2021-03-07
add Method check to ImmutableAttributeConsistencyChecker

omsmith:


src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -562,5 +562,14 @@ public static class Diagnostics {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor InconsistentMethodAttributeApplication = new DiagnosticDescriptor(
+			id: "D2L0076",
+			title: "Attribute is applied inconsistently to method definitions.",
+			messageFormat: "The [{0}] attribute is applied inconsistently between \"{1}\" and its source definition \"{2}\". This wiil cause inconsistent behaviour.",
```
TimothyJCowen, line 569:
```suggestion
			messageFormat: "The [{0}] attribute is applied inconsistently between \"{1}\" and its source definition \"{2}\". This will cause inconsistent behaviour.",
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -45,7 +46,55 @@ INamedTypeSymbol typeSymbol
...
+		public void CheckMethodDeclaration(
+			IMethodSymbol methodSymbol
+		) {
+			if( methodSymbol.TypeParameters.Length == 0 ) {
+				return;
+			}
+
+			MethodDeclarationSyntax methodSyntax = null;
+			Location[] parameterLocations = null;
+			Location GetLocationOfNthParameter( int N ) {
```
j3parker, line 58: Can you just make this a private method?

These methods generate extra code behind the scene and do heap allocations to deal with the captures.

It's simpler, faster and more idiomatic to use a private helper method.

j3parker: The local function is nice if you need to return a `Func<...>` or something.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -45,7 +46,55 @@ INamedTypeSymbol typeSymbol
...
+
+				if( parameterLocations[ N ] != default ) {
+					return parameterLocations[ N ];
+				}
+
+				methodSyntax ??= methodSymbol
+					.DeclaringSyntaxReferences[ 0 ]
+					.GetSyntax() as MethodDeclarationSyntax;
+
+				Location loc = parameterLocations[ N ] = methodSyntax.TypeParameterList.Parameters[ N ].GetLocation();
```
j3parker, line 69: The multiple assignments, `??=`, capturing... try to make the control flow here easier to follow please.

j3parker (changes requested):
I think the control flow could be simpler here.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
omsmith: Had it there for caching, but I'll just remove it since perf doesn't really matter in error cases.

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
omsmith, line 569: Made the diagnostic itself general, since we should be applying the same checks to things like `[StatelessFunc]` and `[Constant]`

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -164,5 +164,32 @@ string name
 
 			return false;
 		}
+
+		public static ImmutableArray<IMethodSymbol> GetImplementedMethods(
```
j3parker, line 168: Possible other things that could use this helper: https://github.com/Brightspace/D2L.CodeStyle/search?q=FindImplementationForInterfaceMember

src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -164,5 +164,32 @@ string name
...
+			); ;
+
+			// Check OverriddenMethod != null instead of IsOverride
+			// because "override void Foo() {}" will have IsOverride == true
+			// even if there isn't a Foo to override
+			if( method.OverriddenMethod != null ) {
+				result.Add( method.OverriddenMethod );
+			}
+
+			return result.ToImmutable();
```
j3parker, line 192: Could avoid allocations with `yield`

j3parker: I guess that's not really true, it would avoid a "bigger" allocation. Meh.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -45,7 +46,38 @@ INamedTypeSymbol typeSymbol
...
+			IMethodSymbol methodSymbol
+		) {
+			if( methodSymbol.TypeParameters.Length == 0 ) {
+				return;
+			}
+
+			ImmutableArray<IMethodSymbol> implementedMethods = methodSymbol.GetImplementedMethods();
+			foreach( IMethodSymbol implementedMethod in implementedMethods ) {
+				Debug.Assert( methodSymbol.TypeParameters.Length == implementedMethod.TypeParameters.Length );
+
```
j3parker, line 59:
```suggestion
```

https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.debug.assert?view=net-5.0

This pops up a message box on failure. This would be very bad on a build server.
If you want the check then use an exception :)

j3parker (approved):

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
j3parker: Like, a win32 UI dialog :)

omsmith: Yeah - I thought it did something more reasonable depending on the situation.

j3parker: It's a shame.

j3parker (approved):

PR #729, opened 2021-03-10
Scrutinize anonymous methods more in ImmutableDefinitionChecker

### LINKED ISSUE #722: Assignments of Lambda expressions in constructors can introduce mutability
Previously we allowed this:

```cs
public class Foo {
  Func<IEnumerable<Guid>> m_x = () => new [] { Guid.NewGuid(); };
}
```

on the grounds that this lambda can't capture any mutable state "owned" by `Foo` that wouldn't otherwise cause an error.

However, now that we are considering assignments inside constructors as well, that changes:

```cs
public class Foo {
  Func<IEnumerable<Guid>> m_x;

  public Foo() {
    var evil = new List<Guid>();

    m_x = () => {
      evil.Add( Guid.NewGuid() );
      return evil;
    };
  }
}
```

We should be more careful and ensure that we only close over immutable things (or are a [`static` lambda](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/static-anonymous-functions)).

We should be more careful

j3parker:
Fixes #722

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
+++ b/src/D2L.CodeStyle.Analyzers/Diagnostics.cs
@@ -571,5 +571,14 @@ namespace D2L.CodeStyle.Analyzers {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor AnonymousFunctionsMayCaptureMutability = new DiagnosticDescriptor(
+			id: "D2L0077",
+			title: "Anonymous functions may capture mutable state and thus can't be assigned to members of an immutable type.",
+			messageFormat: "Anonymous functions may capture mutable state and thus can't be assigned to members of an immutable type. Try making the function static, or moving the assignment to an initializer.",
+			category: "Immutability",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
 	}
 }
--- a/src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.CSharp.Syntax.cs
+++ b/src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.CSharp.Syntax.cs
@@ -60,9 +60,7 @@ namespace D2L.CodeStyle.Analyzers.Extensions {
 			=> node.FirstAncestorOrSelf<DocumentationCommentTriviaSyntax>() != null;
 
 		public static bool IsStaticFunction( this ExpressionSyntax expr ) {
-			var afes = expr as AnonymousFunctionExpressionSyntax;
-
-			if ( afes == null ) {
+			if ( expr is not AnonymousFunctionExpressionSyntax afes ) {
 				return false;
 			}
 
--- a/src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.CSharp.Syntax.cs
+++ b/src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.CSharp.Syntax.cs
@@ -1,4 +1,5 @@
-﻿using Microsoft.CodeAnalysis;
+﻿using System.Linq;
+using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
 
@@ -55,8 +56,19 @@ namespace D2L.CodeStyle.Analyzers.Extensions {
 			return true;
 		}
 
-		public static bool IsFromDocComment(
-			this SyntaxNode node
-		) => node.FirstAncestorOrSelf<DocumentationCommentTriviaSyntax>() != null;
+		public static bool IsFromDocComment( this SyntaxNode node )
+			=> node.FirstAncestorOrSelf<DocumentationCommentTriviaSyntax>() != null;
+
+		public static bool IsStaticFunction( this ExpressionSyntax expr ) {
+			var afes = expr as AnonymousFunctionExpressionSyntax;
+
+			if ( afes == null ) {
+				return false;
+			}
+
+			return afes.Modifiers.Any(
+				m => m.IsKind( SyntaxKind.StaticKeyword )
+			);
+		}
 	}
 }
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
@@ -263,27 +263,33 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			var allAssignmentsAreOfImmutableValues = true;
 
 			foreach( var assignment in assignments ) {
-				var query = GetImmutabilityQueryForAssignment( assignment );
+				var kind = GetQueryForAssignment( assignment, out var query, out var diagnostic );
 
-				if( query == null ) {
-					// null is a signal that there is nothing further that needs to
-					// be checked for this assignment.
-					continue;
+				switch( kind ) {
+					case AssignmentQueryKind.NothingToCheck:
+						continue;
+
+					case AssignmentQueryKind.Hopeless:
+						diagnosticSink( diagnostic );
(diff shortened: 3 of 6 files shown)


src/D2L.CodeStyle.Analyzers/Extensions/Microsoft.CodeAnalysis.CSharp.Syntax.cs
```diff
@@ -55,8 +56,19 @@ INamedTypeSymbol attributeType
...
 		}
 
-		public static bool IsFromDocComment(
-			this SyntaxNode node
-		) => node.FirstAncestorOrSelf<DocumentationCommentTriviaSyntax>() != null;
+		public static bool IsFromDocComment( this SyntaxNode node )
+			=> node.FirstAncestorOrSelf<DocumentationCommentTriviaSyntax>() != null;
+
+		public static bool IsStaticFunction( this ExpressionSyntax expr ) {
+			var afes = expr as AnonymousFunctionExpressionSyntax;
```
j3parker, line 63: This is the base class of `AnonymousMethodExpressionSyntax` and `LambdaExpressionSyntax` (remember this for the StatelessFunc analyzer below)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -339,60 +345,93 @@ ISymbol memberSymbol
 			);
 		}
 
+		private enum AssignmentQueryKind {
+			NothingToCheck,
+			ImmutabilityQuery,
+			Hopeless
+		}
```
j3parker, line 352: What I really wanted was something like:

```cs
private enum AssignmentQuery {
  NothingToCheck,
  Hopeless(Diagnostic),
  SubQuery(ImmutabilityQuery)
}
```

[which some languages support](https://doc.rust-lang.org/rust-by-example/custom_types/enum.html) but not C#.

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
```diff
@@ -77,6 +77,11 @@ ImmutableHashSet<ISymbol> statelessFuncs
 				return;
 			}
 
+			// static functions can't capture state
+			if ( argument.IsStaticFunction() ) {
+				return;
+			}
```
j3parker, line 83: Pulled out common code to an extension and lifted this outside the switch

tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutabilityAnalyzer.cs
```diff
@@ -1421,4 +1421,34 @@ public class ClassExplicitlyImplementingMethodsInconsistent : IInterfaceWithMeth
...
+
+			SometimesBad = /* AnonymousFunctionsMayCaptureMutability */ () => x-- /**/;
+        }
+
+		ClassWithCapturedMutability() {
+			int x = 0;
+
+			SometimesBad2 = static () => 4;
+			SometimesBad2 = static () => AlwaysOk();
+			SometimesBad2 = /* AnonymousFunctionsMayCaptureMutability */ () => x++ /**/;
```
j3parker, line 1450: Not erroring prior to this patch.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -339,60 +345,93 @@ ISymbol memberSymbol
...
-
-			var model = m_compilation.GetSemanticModel( assignment.SyntaxTree );
-			var typeInfo = model.GetTypeInfo( assignment );
+					if ( assignment.Parent.Parent is VariableDeclaratorSyntax ) {
+						// Lambda initializers for readonly members are safe
+						// because they can only close over other members,
+						// which will be checked independently, or static
+						// members of another class, which are also analyzed.
+						return AssignmentQueryKind.NothingToCheck;
+					}
```
omsmith, line 390: I thiiiink this might also match?

```csharp
class C {
    private readonly Func<int> m_foo;
    
    public C() {
        Func<int> bar = m_foo = () => 0;
    }
    

```

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -571,5 +571,14 @@ public static class Diagnostics {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor AnonymousFunctionsMayCaptureMutability = new DiagnosticDescriptor(
+			id: "D2L0077",
+			title: "Anonymous functions may capture mutable state and thus can't be assigned to members of an immutable type.",
+			messageFormat: "Anonymous functions may capture mutable state and thus can't be assigned to members of an immutable type. Try making the function static, or moving the assignment to an initializer.",
```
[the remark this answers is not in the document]
j3parker: I don't think it's required for safety, because the set of stuff they can capture isn't mutable owing to the rest of our analysis (or if it is, there is another diagnostic... this matches the general philosophy of the analyzer)

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
omsmith: nit: collapse and change the pattern in the switch to `AnonymousFunctionExpressionSyntax` and move this there? Diagnostic message can change slightly to accommodate.

It used to be more collapsed, but I guess I missed the base class when switching to static funcs, https://github.com/Brightspace/D2L.CodeStyle/commit/a826fd3e49052dfa9314307454147ee238855a4d#diff-5b07894d106bc2ac1710cd24766f935abaa2a2076f85dc35925a2336c7e2cb96

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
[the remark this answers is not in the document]
omsmith: Yeah, I was just thinking towards not having to distinguish the two cases (yet) a la https://github.com/Brightspace/D2L.CodeStyle/pull/729#discussion_r591680018

src/D2L.CodeStyle.Analyzers/Immutability/StatelessFuncAnalyzer.cs
j3parker: Ah, I thought they were totally different diagnostics but you're right it just varies by the message.
I can do that but I'll do it as a follow-up (since it's a totally different analyzer)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -263,30 +263,49 @@ IEnumerable<ExpressionSyntax> assignments
...
+			}
 
-				diagnosticSink( diagnostic );
+			return isReadOnly && allAssignmentsAreOfImmutableValues;
+		}
 
-				// We're going to keep looking at all writes to surface as many
-				// relevant diagnostics as possible.
-				allAssignmentsAreOfImmutableValues = false;
+		private readonly struct AssignmentInfo {
```
j3parker, line 298: RE: https://github.com/Brightspace/D2L.CodeStyle/pull/729#discussion_r591675842
Now just keeping track of which one is the initializer.
This is a little lame since there is at most one but 🤷🏼‍♂️

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -339,60 +360,93 @@ ISymbol memberSymbol
...
+			out Diagnostic diagnostic
 		) {
 			// When we have an assignment we use it to narrow our check, e.g.
 			//
 			//   private readonly object m_lock = new object();
 			//
 			// is safe even though object (the type of the field) in general is not.
 
+			query = default;
+			diagnostic = null;
```
omsmith, line 388: early assignment makes me sad 😢

omsmith (approved):

TimothyJCowen (approved):

PR #732, opened 2021-03-11
Added check to allow Array/Enumerable.Empty calls as immutable.

TimothyJCowen:


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -425,32 +431,47 @@ private enum AssignmentQueryKind {
 
 					return AssignmentQueryKind.Hopeless;
 
-				default:
```
TimothyJCowen, line 428: Moved this out of the `switch` so that any `InvocationExpression`s that are not in the list still get caught here.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -370,6 +371,11 @@ private enum AssignmentQueryKind {
 			Hopeless
 		}
 
+		internal static readonly ImmutableArray<string> KnownImmutableReturningMethods = ImmutableArray.Create(
```
TimothyJCowen, line 374: Is there a better place for this? Is this even the right way to do it?

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -425,32 +431,47 @@ private enum AssignmentQueryKind {
...
+				case SyntaxKind.InvocationExpression:
+					var semanticModel = m_compilation.GetSemanticModel( assignment.Expression.SyntaxTree );
+					var symbol = semanticModel.GetSymbolInfo( assignment.Expression ).Symbol;
+
+					if( symbol is not IMethodSymbol methodSymbol ) {
+						break;
 					}
 
-					return AssignmentQueryKind.ImmutabilityQuery;
+					var symbolName = methodSymbol.ContainingSymbol.ToDisplayString() + "." + methodSymbol.Name;
```
TimothyJCowen, line 442: Not super hot on the string comparisons, though I'm struggling to find a way to do it otherwise.

j3parker, line 442: Can we map the list of magic things to `ISymbol` and not use string comparison here?

It probably makes sense to go into `ImmutabilityContext` because that is the "is this thing immutable" oracle: it says yes/no based on hard coded lists and the `[Immutable]` attribute... the `ImmutableDefinitionChecker` is the thing that validates the `[Immutable]` attribute was used correctly.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Huh? Just put that case above this?

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -425,32 +431,47 @@ private enum AssignmentQueryKind {
...
+
+					if( KnownImmutableReturningMethods.Contains( symbolName ) ) {
```
j3parker, line 444: Two options:

1) Make sure that we're passing no arguments so that we're always talking about the right overload (in the case of `Array.Empty` there are not going to be overloads, but we might as well make this feature general)
2) Use `ISymbol` so that we know we are talking about exactly the method we mean... when turning strings into `ISymbol` in `ImmutabilityContext` just always go for the zero-parameter method for now.

But... go with (2)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Ah yeah, see the note I left about `ImmutabilityContext`.

j3parker (commented):
Looks good. I left some notes.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
omsmith: Look at how DangerousMethodsAnalyzer loads its `IMethodSymbol`s as a basis. You can simplify it a bit by just having tuples of `(string TypeName, string MethodName)` instead of the dictionary of typename => MethodName[].

And as @j3parker pointed out elsewhere. Filter down to methods with 0 arguments.

https://github.com/Brightspace/D2L.CodeStyle/blob/7e9cde39e549dbc16e916b99b5e3b2e3f1a62f09/src/D2L.CodeStyle.Analyzers/ApiUsage/DangerousMemberUsages/DangerousMemberUsagesAnalyzer.cs#L223-L249

j3parker: Thanks @omsmith .

FYI`ImmutabilityContext` is split into two halves:

* `ImmutabilityContext.Factory.cs`: creates the context by poking around the compilation looking for symbols
* `ImmutabilityContext.cs`: the actual object (e.g. where your `.Contains` would go)

omsmith: Yeah, in slack the only reason I questioned whether ImmutabilityContext was the _right_ suggestion for me to make, is because ImmutabilityContext won't be using it. But not crazy for it to "hold" them and expose afunction to check against it.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.Factory.cs
```diff
@@ -129,5 +128,41 @@ internal partial class ImmutabilityContext {
 			return builder.ToImmutable();
 		}
 
+		public static ISymbol GetSymbol(
+			Compilation compilation,
+			string qualifiedAssembly,
+			string typeName,
+			string methodName = null
+		) {
+			ImmutableDictionary<string, IAssemblySymbol> compilationAssemblies = GetCompilationAssemblies( compilation );
```
omsmith, line 137: Did you intentionally move this expensive thing inside of the loop?

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.Factory.cs
```diff
@@ -77,24 +78,23 @@ internal partial class ImmutabilityContext {
...
 		);
 
+
+		internal static readonly ImmutableArray<(string TypeName, string MethodName, string AssemblyName)> KnownImmutableReturningMethods = ImmutableArray.Create(
+			( "System.Array", "Empty", default(string) ),
+			( "System.Linq.Enumerable", "Empty", default(string) )
+		);
+		
+
 		internal static ImmutabilityContext Create( Compilation compilation, AnnotationsContext annotationsContext ) {
```
omsmith, line 88: This `Create` function should be getting the `IMethodSymbol`s in `KnownImmutableReturningMethods` (`KnownImmutableReturningMethods` should be `private`) and then passing them in to `ImmutabilityContext`, so that getting the symbols is only done once.

omsmith: Explained differently, we're not looking to share ImmutabilityContext's loading code, we're instead looking to expand the scope of what ImmutabilityContext knows about.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
j3parker: Oh breaks, I get it

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.Factory.cs
```diff
@@ -77,24 +78,24 @@ internal partial class ImmutabilityContext {
...
-			ImmutableDictionary<string, IAssemblySymbol> compilationAssmeblies = GetCompilationAssemblies( compilation );
 
-			var builder = ImmutableDictionary.CreateBuilder<INamedTypeSymbol, ImmutableTypeInfo>();
+			var compilationAssemblies = GetCompilationAssemblies( compilation );
```
omsmith, line 90: just unnecessary diffage

```suggestion
			ImmutableDictionary<string, IAssemblySymbol> compilationAssemblies = GetCompilationAssemblies( compilation );
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.Factory.cs
```diff
@@ -103,12 +104,31 @@ internal partial class ImmutabilityContext {
...
+				if( type == null ) {
+					continue;
+				}
+
+				IMethodSymbol methodSymbol = type
+					.GetMembers( methodName )
+					.OfType<IMethodSymbol>()
+					.Single( m => m.Parameters.Length == 0 );
+
+				knownImmutableReturnsBuilder.Add( methodSymbol );
```
omsmith, line 125: Hmm, lets be more cautious here, just in case.

```suggestion
				IMethodSymbol[] methodSymbol = type
					.GetMembers( methodName )
					.OfType<IMethodSymbol>()
					.Where( m => m.Parameters.Length == 0 )
					.ToArray();

				knownImmutableReturnsBuilder.Add( methodSymbol[0] );
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -369,7 +370,7 @@ private enum AssignmentQueryKind {
 			ImmutabilityQuery,
 			Hopeless
 		}
-
+
```
omsmith, line 373:
```suggestion

```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -425,32 +426,50 @@ private enum AssignmentQueryKind {
...
-						);
+				case SyntaxKind.InvocationExpression:
+
+					// Some methods are known to have return values that are
+					// immutable (such as Enumerable.Empty()).
+					// These should be considered immutable by the Analyzer.
+					if( m_compilation
+						.GetSemanticModel( assignment.Expression.SyntaxTree )
+						.GetSymbolInfo( assignment.Expression )
+						.Symbol is not IMethodSymbol methodSymbol ) {
```
omsmith, line 437:
```suggestion
						.Symbol is not IMethodSymbol methodSymbol
					) {
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -425,32 +426,50 @@ private enum AssignmentQueryKind {
...
+						break;
+					}
+
+					if( m_context.IsReturnValueKnownToBeImmutable( methodSymbol ) ) {
+						return AssignmentQueryKind.NothingToCheck;
```
omsmith, line 442: Thinking out loud, we could probably approach this a little differently, such that this becomes `AssignmentQueryKind.MethodInvocationQuery` or some such, so that this `GetQueryForAssignment` is doing fewer checks, but I think this is fine & simple, and we can refactor as desired any time later.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -425,32 +426,50 @@ private enum AssignmentQueryKind {
...
-							ImmutableTypeKind.Total,
-							typeToCheck
-						);
+				case SyntaxKind.InvocationExpression:
+
+					// Some methods are known to have return values that are
+					// immutable (such as Enumerable.Empty()).
+					// These should be considered immutable by the Analyzer.
+					if( m_compilation
+						.GetSemanticModel( assignment.Expression.SyntaxTree )
```
omsmith, line 435: Getting the semantic model twice within this function is unfortunate. Lets do `SemanticModel model = null` above the switch, and do `model ??= m_compilation.GetSemanticModel( assignment.Expressions.SyntaxTree )` in the two usages.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityContext.Factory.cs
```diff
@@ -77,24 +78,24 @@ internal partial class ImmutabilityContext {
 			("System.IO.Abstractions.FileSystem", default)
 		);
 
+
+		internal static readonly ImmutableArray<(string TypeName, string MethodName, string AssemblyName)> KnownImmutableReturningMethods = ImmutableArray.Create(
+			( "System.Array", "Empty", default(string) ),
+			( "System.Linq.Enumerable", "Empty", default(string) )
```
omsmith, line 84:
```suggestion
			( "System.Array", "Empty", default ),
			( "System.Linq.Enumerable", "Empty", default )
```

TimothyJCowen: Can't do.
At least one of them has to have the `(string)` so that the rest can infer the type. The other `ImmutableArray` works because some of them are defined, which tells the rest that they are strings.

omsmith: Could probably do via the `Create<(string, string, string)>`, but, not important.

tests/D2L.CodeStyle.Analyzers.Test/Specs/ImmutabilityAnalyzer.cs
```diff
@@ -185,7 +187,12 @@ public interface SomeImmutableGenericInterfaceRestrictingTU<[Immutable] T, [Immu
 
 		/// <summary><see cref="SomeGenericMethodCref{T}()"/></summary>
 		public static void SomeGenericMethodCref<[Immutable] T>( int x ) { }
-
+		
+		[Immutable]
+		public sealed class Tester {
+			private readonly IEnumerable<int> m_enumerable = Enumerable.Empty<int>();
+			private readonly int[] m_array = Array.Empty<int>()
```
omsmith, line 194: Can you put these in `AnalyzedClassMarkedImmutable`. Also, missing a semicolon.

omsmith (approved):

Issue #738, opened 2021-04-08
Immutability analyzer doesn't see destructuring assignments


j3parker:
```cs
[Immutable]
public sealed class Whatever {
  private readonly object x;

  public Whatever() {
    x = new object(); // ok
    (x) = new ValueTuple<object>(new List<string>()); // needs to be caught
  }
}
```

while we're here, add a test case for the left hand side being parenthesized but not destructuring:

<img width="1325" alt="image" src="https://user-images.githubusercontent.com/597907/114043374-3b083300-9854-11eb-86fb-e8505bdedff2.png">

PR #740, opened 2021-04-09
prep AssignmentInfo to handle deconstructed-assignments

omsmith:
Ref: https://github.com/Brightspace/D2L.CodeStyle/issues/738

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
@@ -301,16 +301,38 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 		}
 
 		private readonly struct AssignmentInfo {
-			public AssignmentInfo(
-				bool isInitiailizer,
-				ExpressionSyntax expression
+			private AssignmentInfo(
+				bool isInitializer,
+				ExpressionSyntax expression,
+				ITypeSymbol assignedType
 			) {
-				IsInitializer = isInitiailizer;
+				IsInitializer = isInitializer;
 				Expression = expression;
+				AssignedType = assignedType;
 			}
 
 			public bool IsInitializer { get; }
 			public ExpressionSyntax Expression { get; }
+			public ITypeSymbol AssignedType { get; }
+
+			public static AssignmentInfo Create(
+				SemanticModel model,
+				bool isInitializer,
+				ExpressionSyntax expression
+			) {
+				TypeInfo typeInfo = model.GetTypeInfo( expression );
+
+				// Type can be null in the case of an implicit conversion where the
+				// expression alone doesn't have a type. For example:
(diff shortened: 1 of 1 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -352,9 +382,7 @@ ExpressionSyntax initializer
 			AssignmentExpressionSyntax assignmentSyntax,
 			ISymbol memberSymbol
 		) {
-			var semanticModel = m_compilation.GetSemanticModel(
-				memberSymbol.DeclaringSyntaxReferences.Single().GetSyntax().SyntaxTree
-			);
+			var semanticModel = m_compilation.GetSemanticModel( assignmentSyntax.SyntaxTree );
```
omsmith, line 385: Err, extra hunk, but correcter change

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -301,16 +301,38 @@ out diagnostic
...
 
 			public bool IsInitializer { get; }
 			public ExpressionSyntax Expression { get; }
+			public ITypeSymbol AssignedType { get; }
+
+			public static AssignmentInfo Create(
+				SemanticModel model,
+				bool isInitializer,
+				ExpressionSyntax expression
+			) {
```
omsmith, line 322: Might not end up with this, ~important change is it now has `AssignedType` exposed

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -432,7 +459,7 @@ private enum AssignmentQueryKind {
 					// Some methods are known to have return values that are
 					// immutable (such as Enumerable.Empty()).
 					// These should be considered immutable by the Analyzer.
-					semanticModel ??= m_compilation.GetSemanticModel( assignment.Expression.SyntaxTree );
+					SemanticModel semanticModel = m_compilation.GetSemanticModel( assignment.Expression.SyntaxTree );
```
j3parker, line 462: I wonder if the symbol info should be part of `AssignmentInfo`, as we are passing in the same semantic model to the `Create`

j3parker (approved):

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
omsmith: Yeah - somehow not calling `GetSemanticModel` everywhere would be good - they cache information for the syntax tree internally. At the same time, don't want to hold onto them longer than necessary, but that's probably the same lifetime as the syntax node itself anyway (as opposed to a syntax reference).

PR #741, opened 2021-04-09
Omsmith/destructured assignments

omsmith:
WIP - opening up to chat

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
@@ -357,13 +357,9 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				.Select( sr => sr.GetSyntax() )
 				.SelectMany( constructorSyntax => constructorSyntax.DescendantNodes() )
 				.OfType<AssignmentExpressionSyntax>()
-				.Where( assignmentSyntax => IsAnAssignmentTo( assignmentSyntax, memberSymbol ) )
-				.Select( assignmentSyntax => assignmentSyntax.Right )
-				.Select( expr => AssignmentInfo.Create(
-					model: m_compilation.GetSemanticModel( expr.SyntaxTree ),
-					isInitializer: false,
-					expression: expr
-				) );
+				.Select( assignmentSyntax => GetAssignmentInfoIfToSymbolOrNull( assignmentSyntax, memberSymbol ) )
+				.Where( info => info.HasValue )
+				.Select( info => info.Value );
 
 			if ( initializer != null ) {
 				assignmentExpressions = assignmentExpressions.Append(
@@ -378,7 +374,7 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 			return assignmentExpressions;
 		}
 
-		private bool IsAnAssignmentTo(
+		private AssignmentInfo? GetAssignmentInfoIfToSymbolOrNull(
 			AssignmentExpressionSyntax assignmentSyntax,
 			ISymbol memberSymbol
 		) {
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -378,19 +386,86 @@ ExpressionSyntax initializer
...
+					return AssignmentInfo.Create(
+						model: semanticModel,
+						isInitializer: false,
+						expression: assignmentSyntax.Right
+					);
+				}
+
+				if( semanticModel.GetTypeInfo( assignmentSyntax.Right ).Type is INamedTypeSymbol assignedType
+					&& assignedType.IsTupleType
+				) {
```
omsmith, line 423: Handle Value Tuples.

`GetDeconstructionInfo` does return stuff for them, but they don't have a `Method`, just some nested deconstructioninfos with `Identity` conversions - so need to extract the type info separately anyway.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -378,19 +386,86 @@ ExpressionSyntax initializer
...
+					if( assignedType.TypeArguments.Length != lhsExpressions.Length ) {
+						return AssignmentInfo.Create(
+							isInitializer: false,
+							expression: assignmentSyntax.Right,
+							assignedType: null
```
omsmith, line 428: > `assignedType: null`

This is an "error / unknown assignment" - there's probably a better way to signal this...

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -378,19 +386,86 @@ ExpressionSyntax initializer
...
+						expression: assignmentSyntax.Right,
+						assignedType: null
+					);
+				}
+
+				return AssignmentInfo.Create(
+					isInitializer: false,
+					expression: assignmentSyntax.Right,
+					assignedType: deconstruction.Method.Parameters[ i ].Type
+				);
```
omsmith, line 465: Only handles simple deconstructions, not nested ones like so:

```csharp
class A {
  public void Deconstruct( out string X, out B foo );
}

class B {
  public void Deconstruct( out string Y, out string Z );
}

public void Foo() {
  var (x, y, z) = new A();
}
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -378,19 +386,86 @@ ExpressionSyntax initializer
...
+				DeconstructionInfo deconstruction = semanticModel.GetDeconstructionInfo( assignmentSyntax );
+				if( deconstruction.Method == null ) {
+					return AssignmentInfo.Create(
+						isInitializer: false,
+						expression: assignmentSyntax.Right,
+						assignedType: null
+					);
+				}
+
+				if( deconstruction.Method.Parameters.Length != lhsExpressions.Length ) {
```
omsmith, line 453: Don't believe this can happen, just a sanity check.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -474,6 +549,17 @@ private enum AssignmentQueryKind {
...
 
+			if( assignment.AssignedType == null
+				|| assignment.AssignedType.TypeKind == TypeKind.Error
+			) {
+				diagnostic = Diagnostic.Create(
+					Diagnostics.NonImmutableTypeHeldByImmutable,
+					assignment.Expression.GetLocation(),
+					"blarg", "blarg", "blarg"
+				);
+				return AssignmentQueryKind.Hopeless;
```
omsmith, line 560: Need to add some sort of new "Unexpected" diagnostic. (We currently have UnexpectedTypeKind and UnexpectedMemberKind)

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableDefinitionChecker.cs
```diff
@@ -333,6 +333,18 @@ ExpressionSyntax expression
 					assignedType: assignedType
 				);
 			}
+
+			public static AssignmentInfo Create(
```
j3parker, line 337: Could the constructor just be public?

omsmith: Yeah for sho. Discovering the end solution organically.

PR #749, opened 2021-05-12
US127307 Add telemetry event logger code style

dustxd:
### Motivation/Context
* As part of the PBMM work we need to ensure we do not log telemetry events for Protected B clients so we added `ITelemetryEventLogger` to encapsulate all the telemetry endpoint usages for better control of what we are returning (we return the telemetry endpoint for non-PBMM clients and `null` for PBMM clients)

This PR has a dependency on another PR to be merged first: https://github.com/Brightspace/lms/pull/10201

### Summary of Changes
* Listed data telemetry endpoint as a banned config and added message to direct developers to use `ITelemetryEventLogger` instead

### Rally Story
https://rally1.rallydev.com/#/29180338367d/custom/486232622040?detail=%2Fuserstory%2F600523947568&fdp=true?fdp=true


omsmith (changes requested):
Marking just in case ;)

dustxd:
> Marking just in case ;)

Heh thank you 😁

omsmith (approved):

PR #750, opened 2021-05-21
allow programmatic usage of ImmutabilityAnlayzer with extra types

omsmith:
Could probably do this via AnalyzerOptions or AdditionalFiles or something, but, 🤷 

For @TimothyJCowen to configure in static-di


j3parker (approved):
We could allow it via additional data, of course we would want to monitor for that though....

For programmatic users this is less annoying, so I think it's good.

EDIT: whoops you already said that.

PR #753, opened 2021-07-28
Add WebpagesRpcParameterTypeValidator

rohitvinnakota:
- [x]  Investigate how to properly reference `D2L.Serialization.IDeserializer`


src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
```diff
@@ -102,6 +112,27 @@ INamedTypeSymbol dependencyAttributeType
...
+
+			CheckThatRpcParametersAreValid( context, method, deserializerType );
+		}
+
+		private static void CheckThatRpcParametersAreValid(
+			SyntaxNodeAnalysisContext context,
+			MethodDeclarationSyntax method,
+			INamedTypeSymbol deserializerType
+		) {
+			if( deserializerType.GetType() == null ) {
```
j3parker, line 124: Hm, so `serializerType.GetType()` is a call to `System.GetType()` which will always return `typeof( INamedType )` symbol here (never `null`).
```suggestion
			if( deserializerType.GetType() == null ) {
```suggestion
			if( deserializerType.GetType() == null ) {

```
```

There is a distinction between the reflection APIs like `GetType()` (which operate on runtime objects) and the Roslyn APIs like `INamedTypeSymbol`, which refer to things from the "semantic model" of the assembly being compiled.

Based on the context I'm guessing this should be

```suggestion
			if( deserializerType == null ) {
```

but `throw new Exception` seems wrong? Like this would crash in VS if they were missing this type. If you want to fail the build we should emit an error diagnostic rather than crash.

j3parker: Hm, but you already checked for `== null` in the initializer and if you couldn't find this type you bailed, so I think you can just exclude this check entirely.

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
```diff
@@ -102,6 +112,27 @@ INamedTypeSymbol dependencyAttributeType
...
+				throw new Exception( "Failed to resolve type: 'D2L.Serialization.IDeserializer'" );
+			}
+			var parameters = method.ParameterList.Parameters;
+			foreach( var parameterType in parameters.Select( p => p.Type ).Where( parameterType => parameterType != null ) ) {
+				bool isValidParameterType = WebpagesRpcParameterTypeValidator.IsValidParameterType( type: parameterType.GetType(), deserializerType: deserializerType.GetType() );
```
j3parker, line 129: `GetType()` won't work here (see other thing about reflection types vs. semantic model types.)
If you want to chat about Roslyn APIs feel free to DM me on Slack.

j3parker:
FYI -- you'll probably want to run this against the LMS as you're developing it.

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
omsmith: I don't think `IDeserializer` is a _necessary_ type for these general checks. I expect the null check should be removed from the initializer, and then logic surrounding `IDeserializer` should only be considered if the type available (similar to the dependency attribute: https://github.com/Brightspace/D2L.CodeStyle/pull/753/files#diff-459cb71828994b9c631d770281ec937c431c9691462e7dbfa168c7321aa445f3R105)

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
```diff
@@ -102,6 +129,54 @@ INamedTypeSymbol dependencyAttributeType
...
+				// Skip injected parameters
+				if( dependencyAttributeType != null ) {
+					if( IsMarkedDependency(
+						dependencyAttributeType: dependencyAttributeType,
+						parameter: parameter,
+						model: context.SemanticModel
+					) ) {
+						continue;
+					}
+				}
```
omsmith, line 163: Could do the same "splitter" approach activation diagnostics was doing (separating parameters into a tuple of (context, dependencies, parameters), instead of duplicating this enumeration and check.

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
```diff
@@ -102,6 +129,54 @@ INamedTypeSymbol dependencyAttributeType
...
+
+				var parameterTypeSymbol = context.SemanticModel.GetSymbolInfo( parameter.Type ).Symbol as ITypeSymbol;
```
omsmith, line 165: Not necessarily `INamedTypeSymbol` (as was written) due to arrays.

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
```diff
@@ -102,6 +129,54 @@ INamedTypeSymbol dependencyAttributeType
 				// other things to check:
 				// - appropriate use of [Dependency]
 			}
+
+			CheckThatRpcParametersAreValid(
```
j3parker, line 133: So this should probably move up into the `if` above (and remove the TODO-esque comment).
The same `if` is replicated inside this method in the loop, so that could be removed if it goes into the `if`.

tests/D2L.CodeStyle.Analyzers.Test/Specs/RpcAnalyzer.cs
```diff
@@ -76,9 +93,32 @@ string b
...
+			DeserializableClass[] l,
+			ClassWithDeserializerConstructor m,
+			IDictionary<ClassWithDeserializerConstructor, int> n,
+			IDictionary<ClassWithDeserializerConstructor, int> o
+		) { }
 	}
 
-	public sealed class BadRpcs {
+	public sealed class BadRpcs
+	{
```
omsmith, line 121:
```suggestion
	public sealed class BadRpcs {
```

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
j3parker: Er, not quite... maybe lifting the loop out to here would make more sense

omsmith: We need to check the other parameters regardless of whether `[Dependency]` is available though, yeah?

j3parker: and maybe moving the null check into `IsMarkedDependency`.... I dno, consider it.

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
j3parker: Ah yeah, same thing I was mentioning up above. You could split or just do a big loop that checks both things.

j3parker (approved):
Consider trying to re-arrange the parameter loop a bit to make it clearer.

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
```diff
@@ -26,171 +28,205 @@ internal sealed class RpcAnalyzer : DiagnosticAnalyzer {
...
+				return (default, ImmutableArray<(IParameterSymbol, ParameterSyntax)>.Empty, ImmutableArray<(IParameterSymbol, ParameterSyntax)>.Empty);
+			}
+
+			SemanticModel model = context.SemanticModel;
+
+			int index = 0;
+
+			ParameterSyntax rpcContextSyntax = parameters[ index++ ];
+			IParameterSymbol rpcContext = model.GetDeclaredSymbol( rpcContextSyntax );
+			var dependenciesBuilder = ImmutableArray.CreateBuilder<(IParameterSymbol, ParameterSyntax)>();
```
omsmith, line 182: Don't actually use the dependencies. Could not collect them.

I'll change that to shouldn't as to not waste analyzer cycles. Activation Diagnostics would make sure the type was actually registered. We do that elsewhere now.

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.IsValidParameterType.cs
```diff
@@ -0,0 +1,90 @@
...
+
+		private static bool HasConstructorDeserializer( INamedTypeSymbol type, RpcTypes rpcTypes ) {
+			if( rpcTypes.IDeserializer == null || rpcTypes.IDeserializer.Kind == SymbolKind.ErrorType ) {
+				return false;
+			}
+
+			foreach( var constructor in type.Constructors ) {
+				if( constructor.DeclaredAccessibility != Accessibility.Public ) {
+					continue;
+				}
```
omsmith, line 65: Force public, so it's not inconsistent between local usage and cross-assembly usage

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.cs
```diff
@@ -26,171 +28,202 @@ internal sealed class RpcAnalyzer : DiagnosticAnalyzer {
...
+			ImmutableHashSet<INamedTypeSymbol> knownRpcParameterTypes = ImmutableHashSet.Create<INamedTypeSymbol>(
+				SymbolEqualityComparer.Default,
+				context.Compilation.GetSpecialType( SpecialType.System_Boolean ),
+				context.Compilation.GetSpecialType( SpecialType.System_Decimal ),
+				context.Compilation.GetSpecialType( SpecialType.System_Double ),
+				context.Compilation.GetSpecialType( SpecialType.System_Single ),
+				context.Compilation.GetSpecialType( SpecialType.System_Int32 ),
+				context.Compilation.GetSpecialType( SpecialType.System_Int64 ),
+				context.Compilation.GetSpecialType( SpecialType.System_String )
+			);
```
omsmith, line 60:
```csharp
		private static readonly ImmutableHashSet<Type> KnownRpcParameterTypes = ImmutableHashSet.Create<Type>(
				typeof( bool ),
				typeof( decimal ),
				typeof( double ),
				typeof( float ),
				typeof( int ),
				typeof( long ),
				typeof( string ),
				typeof( ITreeNode ),
				typeof( IDictionary<string, string> )
			);
```

src/D2L.CodeStyle.Analyzers/ApiUsage/RpcAnalyzer.IsValidParameterType.cs
j3parker: What do the reflecting BindingFlags look like in the LMS now? If they allow private constructors could we stop that too?

j3parker (approved):

PR #773, opened 2022-01-12
Add analyzer to identify and isolate blocking code

j3parker:


src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,234 @@
...
+
+			ctx.ReportDiagnostic(
+				Diagnostic.Create(
+					Diagnostics.AsyncMethodCannotBeBlocking,
+					attrData.ApplicationSyntaxReference.GetSyntax().GetLocation(),
+					methodSymbol.Name
+				)
+			);
+
+			// TODO: code fix to remove [Blocking]
```
j3parker, line 109: I left all the code fixes for the next PR.

Having the fixes will be handy e.g. for switching things to `async` manually/lazily. Mostly there are only a couple of things that can get reused a few times.

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,234 @@
...
+
+			var invokedMethodSymbol = (IMethodSymbol)invokedSymbolInfo.Symbol;
+
+			// async methods are never [Blocking] (see other parts of the analyzer)
+			if( invokedMethodSymbol.IsAsync ) {
+				return;
+			}
+
+			// If the thing we're calling isn't [Blocking] then OK.
+			// TODO: support other means of inferring blockingness, for 3rd party libraries.
```
j3parker, line 137: I'm thinking we could pass some in as a config, and/or do something automatic/fancy, e.g. if there is a compatible looking `FooAsync`, assume that `Foo` is synchronous.

It could work well in the AWS SDK where (in .NET Framework) things are uniform like that. Or maybe we only want to do the automatic thing for some cases, e.g. for the AWS clients...

dno, for later.

JeffAshton:
One thing your PR made me thinking about is interface methods.
Maybe an implementation cannot call blocking methods unless the interface method also has the attribute.
You'd have to match for implicit/explicit interface methods.

JeffAshton (dismissed):

j3parker:
Ah good call. I made the attribute inherited but the other direction is more important: not-blocking should be inherited to avoid insanity. 

I’ll add a check for that, it’ll be easy. Also should test both directions, whoops.

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,234 @@
...
+			if( !HasAttribute( invokedMethodSymbol, blockingAttr ) ) {
+				return;
+			}
+
+			var myMethodDeclaration = invocation.Ancestors()
+				.FirstOrDefault( s => s.IsKind( SyntaxKind.MethodDeclaration ) )
+				as MethodDeclarationSyntax;
```
omsmith, line 144:
```suggestion
			var myMethodDeclaration = invocation.FirstAncestorOrSelf<MethodDeclarationSyntax>();
```

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,234 @@
...
+			}
+
+			ctx.ReportDiagnostic(
+				Diagnostic.Create(
+					Diagnostics.BlockingCallersMustBeBlocking,
+					myMethodDeclaration.Identifier.GetLocation(),
+					invokedMethodSymbol.Name,
+					myMethodSymbol.Name
+				)
+			);
```
omsmith, line 192: Would this be more appropriately placed on the invocation (same as the async-cannot-call)?

tests/D2L.CodeStyle.Analyzers.Test/Specs/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,61 @@
+﻿// analyzer: D2L.CodeStyle.Analyzers.Async.BlockingAnalyzer
```
JeffAshton, line 1: Adding test cases for methods within methods.

```
public static void Foo() {
	static void Bar() { }
	Bar();
}
```

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
omsmith: Also, might need to think about lamdas/anonymous functions

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
j3parker: Hm, seems reasonable. I guess I was thinking that is (approximately) where the suggested code fix goes, but the relationship doesn't need to be so tight.

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
j3parker: Made it more robust

tests/D2L.CodeStyle.Analyzers.Test/Specs/BlockingAnalyzer.cs
j3parker: Done

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
j3parker: Done

JeffAshton (dismissed):

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,359 @@
...
+				ctx => InspectMethodDeclaration(
+					ctx,
+					attr,
+					noteMethodWithPossiblyUnusedBlockingAttribute: m
+						=> havePossiblyUnusedBlockingAttribute[m] = true
+				),
+				SymbolKind.Method
+			);
+
+			var callsSomethingBlocking = new ConcurrentDictionary<IMethodSymbol, bool>(
```
omsmith, line 58: nit: names `havePossiblyUnusedBlockingAttribute` of `callsSomethingBlocking` are referring to plural and singular items. Switch either to `has` or `call` to align them.

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,359 @@
...
+		private static void CompilationStart( CompilationStartAnalysisContext ctx ) {
+			var attr = ctx.Compilation.GetTypeByMetadataName(
+				"D2L.CodeStyle.Annotations.BlockingAttribute"
+			);
+
+			if( attr == null ) {
+				return;
+			}
+
+			var havePossiblyUnusedBlockingAttribute = new ConcurrentDictionary<IMethodSymbol, bool>(
```
omsmith, line 44: I've been a fan of target-type new. Just sharing, not asking you to change it if you prefer the var.

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,359 @@
...
+
+				ctx.ReportDiagnostic(
+					Diagnostic.Create(
+						Diagnostics.AsyncMethodCannotBeBlocking,
+						attrData.ApplicationSyntaxReference.GetSyntax().GetLocation(),
+						methodSymbol.Name
+					)
+				);
+				return;
+			}
```
omsmith, line 153: Should this more generically be "returns awaitable" rather than relying on the async modifier itself?

omsmith: Similar logic would be needed if we wanted to make this an error:

```csharp
public IFoo {
  [Blocking]
  Task BarAsync();
}
```

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,359 @@
...
+
+			// Don't let us use [Blocking] if we're implementing something non-blocking,
+			// callers need to know that they may block.
+			if( implementedNonBlockingThings.Count > 0 ) {
```
omsmith, line 157: I don't think there's explicit TODO comments about it anymore, so checking:

We'll have to be able to configure, say `IDisposable.Dipose`, as `[Blocking]` and not just be relying on `HasAttribute`, right?

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,359 @@
...
+			}
+
+			var attrData = GetAttribute( methodSymbol, blockingAttr );
+
+			if( attrData == null ) {
+				if( implementedBlockingThings.Count == 0 ) {
+					// We don't have [Blocking] and none of the things we implement are
+					// blocking, so there is nothing to do here.
+					return;
+				} else {
```
omsmith, line 111: nit: unnecessary else

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
omsmith: Ah, found the TODO I was thinking of in the method call one

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,359 @@
...
+			if( invokedSymbolInfo.Symbol?.Kind != SymbolKind.Method ) {
+				return;
+			}
+
+			var invokedMethodSymbol = (IMethodSymbol)invokedSymbolInfo.Symbol;
+
+			// async methods are never [Blocking] (see other parts of the analyzer)
+			if( invokedMethodSymbol.IsAsync ) {
+				return;
+			}
```
omsmith, line 202: Is this visible cross-assembly? Just an optimization to skip the attribute check, regardless, yes?

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,359 @@
...
+					Diagnostic.Create(
+						Diagnostics.AsyncMethodCannotCallBlockingMethod,
+						invocation.GetLocation(),
+						invokedMethodSymbol.Name
+					)
+				);
+
+				// TODO: code fix that looks for an async equivalent and suggests calling it instead.
+				return;
+			}
```
omsmith, line 238: Same question as above, re: returning awaitables/tasklikes

```csharp
public Task FooAsync() {
  BlockingCall();
  BlockingCall2();

  return BarAsync();
}
```

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
```diff
@@ -0,0 +1,359 @@
...
+				case SyntaxKind.ConstructorDeclaration:
+					place = "constructors";
+					break;
+				case SyntaxKind.ParenthesizedLambdaExpression:
+				case SyntaxKind.SimpleLambdaExpression:
+					place = "lambdas";
+					break;
+				case SyntaxKind.AnonymousMethodExpression:
+					place = "delegates";
+					break;
```
omsmith, line 327: Wouldn't this be incompatible with say. `T ICache.Add<T>( Func<T> createAndReturn, ... )`, where `createAndReturn` is guaranteed to call something blocking? I this case, could alternatively call the blocking thing ahead of time and just capture the result in the lambda, but that doesn't work for say `GetSetAndCacheLocaly` where the fetch parameters are determined by the cache (filtering out keys that are already cached).

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
j3parker: I had thought about it and decided not to. If we did it would probably be better to do `IAsyncResult` instead, but admittedly the difference isn't going to matter for D2L. I decided not to because it's not immediately trouble to do blocking calls and return a `Task`. It's very likely the case that if you were doing this you'd end up `await`ing anyway (and get the error) but if not, I mean, maybe you're fine? You can certainly imagine scenarios where its fine, but I don't think they are relevant for D2L either way.

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
j3parker: I was thinking (but didn't communicate) that we'd ultimately need something like `StatelessFunc` for blocking funcs :(

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
j3parker: It's not. Ya when I originally wrote it it was an optimization but like the code has changed a bunch since then and I never gave it much thought -- I'll remove it because there is already enough going on in here.

j3parker: Oh, also early attempts to avoid spurious diagnostics, like if you were making a function be async then don't bug all of the callers about how they need to be blocking/start bugging them that they can remove blocking, but switching from sync -> async requires a bunch of changes at the call site anyway so I don't think this is legit.

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
j3parker: ditto -- I intentionally allowed blocking in here because it isn't itself trouble, but it is debatable.

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
j3parker: (which would still of course require changes here)

src/D2L.CodeStyle.Analyzers/Async/BlockingAnalyzer.cs
omsmith: So if we never awaited something in the whole stack not encountering the error, is it problematic to return a "blocking Task" in an async controller?

Either way, you're right it should get "caught eventually", happy to defer.

j3parker: Possibly not as bad, but it really depends. I switched it to `IAsyncResult` return vals.

omsmith (approved):

PR #780, opened 2022-01-25
Update LangVersion and enable implicit usings

j3parker:


Directory.build.props
```diff
@@ -1,6 +1,7 @@
 <Project>
 	<PropertyGroup>
 		<Deterministic>true</Deterministic>
-		<LangVersion>9.0</LangVersion>
+		<LangVersion>10.0</LangVersion>
+		<ImplicitUsings>true</ImplicitUsings>
```
j3parker, line 5: This is the default for new projects in their template. It maybe feels weird but I think its pretty nice when you get used to it. The list depends on your project's SDK, for normal .NET things it is:

```
System
System.Collections.Generic
System.IO
System.Linq
System.Net.Http
System.Threading
System.Threading.Tasks
```

src/D2L.CodeStyle.Annotations/D2L.CodeStyle.Annotations.csproj
```diff
@@ -2,7 +2,8 @@
 
   <PropertyGroup>
     <TargetFrameworks>net20;netstandard2.0</TargetFrameworks>
-    <LangVersion>7.3</LangVersion>
+    <!-- We build this for .NET 2.0 and the implicit usings bring in namespaces that didn't exist back then -->
+    <ImplicitUsings>disable</ImplicitUsings>
```
j3parker, line 6: .NET 2.0 was for the COM stuff, I'm not sure if we still need to keep doing this but whatever, its just the annotations DLL.

Directory.build.props
j3parker: `<nullable>` is also enabled by default. I haven't done that yet that in this PR because there will be some clean-up.

omsmith (approved):

Directory.build.props
JeffAshton: The `ImplicitUsings` scares me, but not enough to get use to the new world

JeffAshton (approved):

Directory.build.props
j3parker: It's pretty questionable. Like they better be really careful about adding new things in those namespaces... but 99% of the time you end up having all of those anyway so it's really not much of a change.

j3parker: Mostly I just don't want to fight it, but it does make writing a new file a little less annoying (which is a niche concern)

PR #782, opened 2022-01-28
Release D2L.CodeStyle.Analyzers@0.180.0

mthjones:
This version fixes the name of the StaticDILocator type being banned.


omsmith (approved):

mthjones:
@j3parker Looks like this will also include some of the blocking analyzer codefixes. Are those good to release here as well?

j3parker (approved):

j3parker:
Ah yeah, thanks!

j3parker:
No one should use it yet so I'm not too worried. Also they should work fine.

PR #784, opened 2022-01-28
Add SourceFilePathHintNameGenerator

j3parker:


src/D2L.CodeStyle.Analyzers/Async/Generator/SourceFilePathHintNameGenerator.cs
```diff
@@ -0,0 +1,110 @@
...
+
+		private static readonly Regex DisallowedPathChars = new(
+			@"[^a-zA-Z0-9-._]*",
+			RegexOptions.Compiled
+		);
+
+		public IEnumerable<string> GetHintNames() {
+			// Hint names need to be unique, but the sanitization we could do
+			// may introduce ambiguities.
+			var used = new HashSet<string>();
```
j3parker, line 79: Alternatives we talked about:

* When there is an ambiguity suffix with the sha256 of the path or something: a pro here is that it would be more stable as people add different paths (the user of this code will pass in sorted input, but if you add/remove files then it could cause unrelated files to have different hintnames) which would be interesting if we were checking the output into git (so that it shows up in PRs with diffs)
* We talked about an unambiguous encoding, like how percent encoding does ☃ -> `%E2%98%83` and `%` -> `%25` or other related schemes... but then I'd need to not do the `.Replace( x, "" )` below that deletes bad chars.

Honestly I don't really care, I'll do whatever. I don't think this is a very likely scenario (maybe if you had `D2L.Foo.Bar\x.cs` and `D2L.FooBar\x.cs`? but like realy?) but I didn't want to crash (really annoying in VS) or bother plumbing diagnostics out of here.

src/D2L.CodeStyle.Analyzers/Async/Generator/SourceFilePathHintNameGenerator.cs
```diff
@@ -0,0 +1,110 @@
+﻿#nullable enable
```
j3parker, line 1: Oh, I should note I'm manually enabling this in all my files.

I did a quick look at enabling this for the whole project and it was a little daunting. It was one of the contributor to my other questions on Slack about deleting particular analyzers :P

Not impossible to migrate the whole thing, I just haven't gotten around to it.

omsmith (approved):

omsmith:
Not doing the ".g" or ".generated" suggestion?

j3parker:
Oh, I can. It's suggested somewhere?

j3parker:
image
Ok -- will do. Interestingly they add `.cs` if you don't.

j3parker:
Their code -- https://sourceroslyn.io/#Microsoft.CodeAnalysis/SourceGeneration/AdditionalSourcesCollection.cs,32
I could make that regex a bit more tolerant too.

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/SourceFilePathHintNameGeneratorTests.cs
```diff
@@ -0,0 +1,94 @@
+﻿using System.Collections.Immutable;
+using NUnit.Framework;
+
+namespace D2L.CodeStyle.Analyzers.Async.Generator;
+
+[TestFixture]
+internal class SourceFilePathHintNameGeneratorTests {
```
mthjones, line 7: Consider adding some cases with forward slashes as well, to make sure we're good for non-Windows systems.

Also, totally silly scenario, but consider adding a case for files that are on different drives as well. Could happen if, for example, source files were pulled from a NuGet package cached on C: while building from a checkout on D:.

src/D2L.CodeStyle.Analyzers/Async/Generator/SourceFilePathHintNameGenerator.cs
```diff
@@ -0,0 +1,110 @@
...
+				while( idx < maxPrefixLength ) {
+					if( sourcePath[idx] != commonPrefix[idx] ) {
+						break;
+					}
+
+					idx++;
+				}
+
+				if( idx != commonPrefix.Length ) {
+					commonPrefix = sourcePath.Substring( 0, idx );
```
JeffAshton, line 46: This could lead to a possible noisy diff if we check these files in.  Like all of a sudden someone checks in on new file, and 50 others will show as changes because a new common path is established

src/D2L.CodeStyle.Analyzers/Async/Generator/SourceFilePathHintNameGenerator.cs
JeffAshton: Can do do the opposite.  Disable all existing files and enabled by default?

JeffAshton (dismissed):
See comments

src/D2L.CodeStyle.Analyzers/Async/Generator/SourceFilePathHintNameGenerator.cs
j3parker: I'm fine with that in general, just wasn't bothering with in this PR anyway.

src/D2L.CodeStyle.Analyzers/Async/Generator/SourceFilePathHintNameGenerator.cs
j3parker: Yep.

src/D2L.CodeStyle.Analyzers/Async/Generator/SourceFilePathHintNameGenerator.cs
JeffAshton: Good to remove now

j3parker: Done

JeffAshton (approved):

PR #785, opened 2022-01-28
Add SyntaxTransformer base class

j3parker:


src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+
+namespace D2L.CodeStyle.Analyzers.Async.Generator;
+
+/// <summary>
+/// Base class for things that transform a SyntaxNode into another and collect
+/// diagnostics. This is useful for writing generators that copy+paste code
+/// with only light modifications.
+/// </summary>
+public abstract class SyntaxTransformer {
```
j3parker, line 14: I don't actually anticipate a second one of these (I hope) but I was keeping the common stuff in a separate class I could drop the `private`/`protected` stuff (which I'm using to hide like, `m_diagnostics` behind a dinky API) and use a `partial` if you dislike it.

j3parker: To use this class you `new` up a transformer and hand it a node, and it will either give you a node back or a bunch of diagnostics.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
...
+	/// Reports an error in our generator to the user.
+	/// </summary>
+	/// <param name="location"></param>
+	/// <param name="what"></param>
+	protected void GeneratorError( Location location, string what ) =>
+		ReportDiagnostic( Diagnostics.GenericGeneratorError, location, Name, what );
+
+	/// <summary>
+	/// Transform every element of a SyntaxList and filter out nulls.
+	/// </summary>
```
j3parker, line 81: These give a hint at what the async half of this will look like, but transforing elements in a list felt like a common case so I pushed that down to here.

src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -735,5 +735,14 @@ public static class Diagnostics {
...
 		);
+
+		public static readonly DiagnosticDescriptor GenericGeneratorError = new DiagnosticDescriptor(
+			id: "D2L0095",
+			title: "A source generator encountered an unexpected situation. This is a bug in D2L.CodeStyle.Analyzers.",
+			messageFormat: "The source generator {0} encountered an unexpected situation: {1}",
+			category: "Generators",
+			defaultSeverity: DiagnosticSeverity.Error,
+			isEnabledByDefault: true
+		);
```
j3parker, line 746: For places where I'm making assumptions about C# trees that I believe are true, even accounting for like partially written code in VS (but it is easy to be wrong about that)

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
...
+	protected abstract string Name { get; }
+
+	public SyntaxTransformer( SemanticModel model ) {
+		m_model = model;
+	}
+
+	/// <summary>
+	/// Suggested return type for public transformation methods.
+	/// </summary>
+	public sealed record TransformResult<T>(
```
JeffAshton, line 30: `record struct` might make sense here

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
...
+/// with only light modifications.
+/// </summary>
+public abstract class SyntaxTransformer {
+	protected readonly SemanticModel m_model;
+	private readonly ImmutableArray<Diagnostic>.Builder m_diagnostics = ImmutableArray.CreateBuilder<Diagnostic>();
+
+	/// <summary>
+	/// A name for the generator that will be shown to the user if we crash
+	/// </summary>
+	protected abstract string Name { get; }
```
JeffAshton, line 21: Not used in this class FYI

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	/// <summary>
+	/// Suggested return type for public transformation methods.
+	/// </summary>
+	public sealed record TransformResult<T>(
+		T? Value,
+		ImmutableArray<Diagnostic> Diagnostics
+	) {
+		public bool Success => Diagnostics.Length == 0;
```
JeffAshton, line 34: `.IsEmpty`

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
...
+			Value: value,
+			Diagnostics: ImmutableArray<Diagnostic>.Empty
+		) { }
+	}
+
+	protected TransformResult<T> GetResult<T>( T value ) {
+		var result = m_diagnostics.ToImmutable();
+		m_diagnostics.Clear();
+
+		if( result.Length == 0 ) {
```
JeffAshton, line 51: consider `.IsEmpty`

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
...
+	/// <summary>
+	/// Transform every element of a SyntaxList and filter out nulls.
+	/// </summary>
+	protected SyntaxList<U> TransformAll<T, U>(
+		SyntaxList<T> input,
+		Func<T, U?> transformer
+	) where T : SyntaxNode
+	  where U : SyntaxNode
+	=> SyntaxFactory.List(
+		(IEnumerable<U>)input // Cast away the ? because of the Where filter
```
JeffAshton, line 88: Be good to measure the LINQ overhead later on.

JeffAshton (approved):

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
omsmith: Used in `GeneratorError`

https://github.com/Brightspace/D2L.CodeStyle/pull/785/files#diff-c358c76ce31b7970f2ed441e3a685108b80e12e0f0cb42e468139b51e1cc550dR77

JeffAshton: My bad

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
...
+	protected SyntaxTokenList TransformAll(
+		SyntaxTokenList input,
+		Func<SyntaxToken, SyntaxToken?> transformer
+	) => SyntaxFactory.TokenList(
+		input.Select( transformer )
+			.Where( NotNull )
+			.Select( static t => t!.Value )
+	);
+
+	private static bool NotNull<T>( T? t ) => t != null;
```
omsmith, line 119: Take it or leave it. Could theoretically be good for us to use the pattern matching for null checks more generally (no operator overloading), but they're not a tool I'm used to reaching for yet.

```suggestion
	private static bool NotNull<T>( T? t ) => t is not null;
```

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
...
+	/// <summary>
+	/// Transform every element of a SyntaxTokenList and filter out nulls.
+	/// </summary>
+	protected SyntaxTokenList TransformAll(
+		SyntaxTokenList input,
+		Func<SyntaxToken, SyntaxToken?> transformer
+	) => SyntaxFactory.TokenList(
+		input.Select( transformer )
+			.Where( NotNull )
+			.Select( static t => t!.Value )
```
omsmith, line 116: I don't think I understand the `.Value` access here.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
j3parker: I might change this anyway

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
j3parker: Ah, `SyntaxToken` is a struct so `SyntaxToken?` is `Nullable<SyntaxToken>`

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -0,0 +1,120 @@
+﻿#nullable enable
```
JeffAshton, line 1: Can remove now

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
j3parker: Done

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
j3parker: So these lists are pretty small, but this would get used frequently in things with `[GenerateSync]`.
`IEnumerable` avoids a copy but has its own book-keeping.

I switched it to a generator which still has some book-keeping. The `SyntaxToken?` stuff made the types awkward for sharing code, but `SyntaxToken` has the "quirk" that `default` is a well-known "none token", so I changed the signature on that one to not use `?` (really, `null` just means there are two kinds of "none" and we weren't filtering out `default` before anyway....)

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
j3parker: I don't think I have a preference, this helpers is gone now tho.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
j3parker: I ended up switching `SyntaxToken?` to `SyntaxToken` and relying on `default` which works out to be a "none token".
It made stuff a little simpler.

JeffAshton (approved):

PR #787, opened 2022-01-28
Remove test analyzers covered by nunit.analyzers (2nd pass)

mthjones:
This removes the D2LTESTS006 diagnostic and its associated analyzers. These are covered by the `nunit.analyzers`[^0] package we are bringing in instead.

More possibility of debate on this one. It's only an "Info" diagnostic by default, which we'll likely want to increase. However, if we increase it we'll need to switch all the exclusions in the allowlist right now to use "SuppressMessage" instead.

[^0]: https://github.com/nunit/nunit.analyzers


j3parker (approved):

src/D2L.CodeStyle.TestAnalyzers/Common/Diagnostics.cs
```diff
@@ -33,16 +33,6 @@ public static class Diagnostics {
 			description: "Custom service locators are expensive to instantiate and slow down tests significantly. Use the default locator instead."
 		);
 
-        public static readonly DiagnosticDescriptor TestAttributeMissed = new DiagnosticDescriptor(
-            id: "D2LTESTS006",
-            title: "Method not labelled as [Test], [Theory], [TestCase], or [TestCaseSource]",
```
mthjones, line 38: https://github.com/nunit/nunit.analyzers/blob/master/documentation/NUnit1026.md

j3parker (approved):

omsmith (approved):

PR #791, opened 2022-01-29
Fixing the diagnostics warnings about punctuation

JeffAshton:


src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -13,7 +13,7 @@ public static class Diagnostics {
 			category: "Correctness",
```
JeffAshton, line 13: Fixed all the straight forward ones, but this is the sort of thing it warns about:

image

j3parker (approved):

PR #814, opened 2022-01-31
AttributeAliasesAnalyzer

JeffAshton:
…hing different


src/D2L.CodeStyle.Analyzers/Language/AttributeAliasesAnalyzer.cs
```diff
@@ -0,0 +1,115 @@
...
+				return;
+			}
+		}
+
+		private static bool IsEquivalentToUsingAlias(
+				IdentifierNameSyntax attributeIdentifier,
+				IdentifierNameSyntax usingAliasIdentifier
+			) {
+
+			string attributeName = attributeIdentifier.ToString();
```
omsmith, line 63: nit: `ToString` in every iteration

omsmith (dismissed):
Likely doesn't catch any global usings

src/D2L.CodeStyle.Analyzers/Language/AttributeAliasesAnalyzer.cs
JeffAshton: I think the following is substantially true,

    `number of files with attributes` > `number of files with attributes & more than 1 alias`

So much so that formatting the string ahead of time is probably extra work in the grand scheme of things.

JeffAshton:
> Likely doesn't catch any global usings

I say we block global usings that are aliases if that's a thing.

src/D2L.CodeStyle.Analyzers/Language/AttributeAliasesAnalyzer.cs
```diff
@@ -0,0 +1,115 @@
...
+			}
+
+			bool equivalent = string.CompareOrdinal( attributeName, 0, usingAlias, 0, shortUsingAliasLength ) == 0;
+			return equivalent;
+		}
+
+		private IEnumerable<IdentifierNameSyntax> GetUsingAliases( AttributeSyntax attribute ) {
+
+			IEnumerable<UsingDirectiveSyntax> rootUsingDirectives = attribute.Ancestors()
+				.OfType<CompilationUnitSyntax>()
```
j3parker, line 87: `attribute.SyntaxTree.GetCompilationUnitRoot()` should usually be faster I think

tests/D2L.CodeStyle.Analyzers.Test/Specs/AttributeAliasesAnalyzer.cs
```diff
@@ -0,0 +1,59 @@
+﻿// analyzer: D2L.CodeStyle.Analyzers.Language.AttributeAliasesAnalyzer
+
+using System;
+
+using RootAliasAttribute = Test.Attributes.TestAttribute;
```
j3parker, line 5: Probably a bit silly but a test case for `using static` to pull an attribute out of a static class? Ahem `Objects.Immutable` 😒

src/D2L.CodeStyle.Analyzers/Language/AttributeAliasesAnalyzer.cs
j3parker: (Mostly just field accesses)

j3parker (approved):
see comments but 👍

src/D2L.CodeStyle.Analyzers/Language/AttributeAliasesAnalyzer.cs
JeffAshton: Switched

j3parker (approved):

tests/D2L.CodeStyle.Analyzers.Test/Specs/AttributeAliasesAnalyzer.cs
JeffAshton: Added

omsmith (approved):

j3parker (approved):

src/D2L.CodeStyle.Analyzers/Language/AttributeAliasesAnalyzer.cs
```diff
@@ -0,0 +1,119 @@
...
+				NameEqualsSyntax? alias = usingDirective.Alias;
+				if( alias == null ) {
+					continue;
+				}
+
+				// ignore aliases that just import the class from the namespace
+				SimpleNameSyntax unqualifiedUsingName = usingDirective.Name.GetUnqualifiedName();
+				if( StringComparer.Ordinal.Equals( alias.Name.ToString(), unqualifiedUsingName.ToString() ) ) {
+					continue;
+				}
```
j3parker, line 113: Extension method on `NameSyntax`? It seems like this would need to get re-used a bunch.

JeffAshton: Will need to play to feel things out.  Will depend on the next PR too.  Whether we just want to check both, or keep it simple with one.

PR #815, opened 2022-01-31
AttributeUnqualifiedNameAnalyzer

JeffAshton:


src/D2L.CodeStyle.Analyzers/Language/AttributeUnqualifiedNameAnalyzer.cs
```diff
@@ -0,0 +1,52 @@
...
+				AttributeSyntax attribute
+			) {
+
+			string unqualifiedName = attribute.Name
+				.GetUnqualifiedName()
+				.ToString();
+
+			if( !unqualifiedName.EndsWith( "Attribute" ) ) {
+				return;
+			}
```
j3parker, line 42: So `class JsonAttributeAttribute` would be unusable?
I'm not super worried about that or anything.
(silly example because json doesn't have "attributes" but like, I don't know maybe there could be a case like that for something.)

j3parker: basically I'm asking: is it worth the effort to try harder here?

j3parker (approved):

src/D2L.CodeStyle.Analyzers/Language/AttributeUnqualifiedNameAnalyzer.cs
JeffAshton: So Visual Studio already complains with a Style warning:

image

https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0001

I see no value in being able to specify the attributes 2 different ways.

omsmith: Right - Jacob's explaining a case where "Attribute" is part of the domain model, which this is insensitive to.

```csharp

public class XmlSerializerAttribute : Attribute {}
public class XmlAttributeAttribute : Attribute {}
public class XmlElementAttribute : Attribute {}

[XmlSerializer]
public class Person {

  [XmlAttribute]
  public string FirstName { get; }
  
  [XmlElement]
  public string LastName { get; }

}
```

```xml
<person firstName="John">
  <lastName>Smith</lastName>
</person>
```

This would complain about the correct `[XmlAttribute]` usage.

JeffAshton: No I get that, but that doesn't exist in the dotnet framework or our framework, so not super concerned.

JeffAshton: Oh it does exist:

https://source.dot.net/#System.Private.Xml/System/Xml/Serialization/XmlAttributeAttribute.cs,64ee1e30fec3074f,references

JeffAshton: I guess it doesn't search suffixes:

image

PR #816, opened 2022-01-31
rework allowlist prevent unnecessary entries

omsmith:


omsmith:
Seems a little too variable in general to meaningfully conclude though.

image

PR #828, opened 2022-02-01
Explicitly setting TreatWarningsAsErrors to the false

JeffAshton:
Explicitly setting TreatWarningsAsErrors to the false default so that it doesn't pick up the lms settings


src/D2L.CodeStyle.Analyzers/D2L.CodeStyle.Analyzers.csproj
```diff
@@ -19,6 +19,7 @@
     <IncludeSymbols>True</IncludeSymbols>
     <Nullable>enable</Nullable>
     <WarningsAsErrors>Nullable</WarningsAsErrors>
+    <TreatWarningsAsErrors>false</TreatWarningsAsErrors>
```
JeffAshton, line 22: Explicitly setting so it doesn't pick up the lms default.  We do want to burn this down and set to true though

omsmith (approved):

PR #832, opened 2022-02-02
Add FileCollector (for the async generator)

j3parker:


src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,273 @@
...
+
+namespace D2L.CodeStyle.Analyzers.Async.Generator;
+
+internal partial class SyncGenerator {
+	/// <summary>
+	/// Creates a source file containing generated methods and the necessary
+	/// namespaces, using directives, and type declarations for them to live in
+	/// (but only the necessary ones).
+	/// </summary>
+	internal sealed class FileCollector {
```
j3parker, line 15: This piece isn't used yet, but there are some tests below that show what it's for.
The `Create()` API is a bit awkward to call, but it works well in the generator pipeline.

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,273 @@
...
+
+		// This is a mutable dictionary because we will remove from it as we go
+		// and check that it is empty at the end, just to be safe.
+		private readonly Dictionary<SyntaxNode, IEnumerable<string>> m_methods;
+
+		// When we recurse into a namespace or type, we're not sure if the
+		// preamble ("namespace foo {", "\t\tpartial class Bar<T> {", etc.) is
+		// necessary, because maybe there are no generated methods.
+		// Using a LinkedList<string> as a double-ended queue.
+		private readonly LinkedList<string> m_preambles = new();
```
j3parker, line 32: A real double-ended-queue would be more efficient, but its not like this list will be big (one entry for each nested namespace/type)

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,273 @@
...
+
+		public string CollectSource() {
+			// File-scoped usings:
+			m_out.Append( m_root.Usings.ToFullString() );
+
+			WriteChildren( m_root );
+
+			if( m_preambles.Count != 0 ) {
+				throw new Bug( "left over preambles" );
+			}
```
j3parker, line 66: These caught some bugs while I was writing it, so I left it in place. It's cheap and it makes it easier to debug tests too.

j3parker: These `FileCollector.Bug` exceptions will get caught and turned into diagnostics to not anger VS.

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,273 @@
...
+
+		private bool WriteNamespace( BaseNamespaceDeclarationSyntax @namespace ) {
+			SyntaxToken? closeBrace;
+
+			switch( @namespace ) {
+				case FileScopedNamespaceDeclarationSyntax ns:
+					m_preambles.AddLast(
+						ns.NamespaceKeyword.ToFullString()
+						+ ns.Name.ToFullString()
+						+ ns.SemicolonToken.ToFullString()
```
j3parker, line 120: Something to note in here is that I'm religiously copy+pasting tokens/syntax from the source file into the destination. This avoids introducing inconsistent whitespace, e.g. the wrong kind of newlines etc.

The only new whitespace tokens I ever use are either an empty token list or a single space character (these are cached in statics above). I do create `partial` tokens and insert them obviously. I could copy the source partial token but by always putting `partial` into the generated code the user gets a diagnostic in their code saying "yo, add a partial here" which is ideal.

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: Like at the point we use this we will have (possibly cached) generated source strings to replace select methods within a single file, and this things responsibility is to concatenate them with all the required usings/hierarchy (partial types/namespaces)

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,273 @@
...
+			if( wroteSomething && m_preambles.Count != 0 ) {
+				throw new Bug( "Left over preambles" );
+			}
+
+			if( !wroteSomething && m_preambles.Count != countBefore ) {
+				throw new Bug( "Incorrect number of preambles" );
+			}
+		}
+
+		public sealed class Bug : Exception {
```
JeffAshton, line 263: nit: `BugException`

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,273 @@
...
+
+			return wroteSomething;
+		}
+
+		private bool WriteNamespace( BaseNamespaceDeclarationSyntax @namespace ) {
+			SyntaxToken? closeBrace;
+
+			switch( @namespace ) {
+				case FileScopedNamespaceDeclarationSyntax ns:
+					m_preambles.AddLast(
```
JeffAshton, line 117: Hmm.  Should m_preambles be a `LinkedList<SyntaxToken[]>` to punt the string concatenation down to StringBuilder?

JeffAshton (approved):

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: Good call. There are two other types beyond `SyntaxToken` that we put into the preamble (covered by `SyntaxNode` and, annoyingly, `SyntaxList<UsingDeclarationSyntax>`). I added a little wrapper struct for this to defer the calls to `ToFullString()`.

There is the win, as you say, from sending them through `StringBuilder`. We also will win a bit for discarded preambles although I suspect that will not be the common case. We pay a tax of allocating that (small) array in gen0 and a little bit of reflection (casts) to implement `ToString()` in the wrapper. This seems reasonable to me.

Something I didn't mention but once we have some serious usage of this it would be good to spy on what the typical size of the `StringBuilder` in practice so that we can tune the initial capacity.

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,304 @@
...
+
+			public static implicit operator PieceOfSyntax( SyntaxList<UsingDirectiveSyntax> list )
+				=> new() { Data = list };
+
+			public override string ToString() => Data switch {
+				SyntaxNode n => n.ToFullString(),
+				SyntaxToken t => t.ToFullString(),
+				SyntaxList<UsingDirectiveSyntax> nl => nl.ToFullString(),
+				_ => throw new BugException( "invalid piece of syntax " + Data.GetType().Name )
+			};
```
j3parker, line 49: See the other thread for context. `SyntaxList<UsingDirectiveSyntax>` may seem really specific, but I don't expect this list of types to grow over time -- this functionality should be relatively "done".

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,304 @@
...
+		private readonly Dictionary<SyntaxNode, IEnumerable<string>> m_methods;
+
+		/// <summary>
+		/// Wraps the sort of syntax that we will append as part of a preamble, to
+		/// defer calls to ToFullString()
+		/// </summary>
+		private readonly struct PieceOfSyntax {
+			public object Data { get; init; }
+
+			public static implicit operator PieceOfSyntax( SyntaxNode node)
```
omsmith, line 35:
```suggestion
			public static implicit operator PieceOfSyntax( SyntaxNode node )
```

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,304 @@
...
+			CompilationUnitSyntax root,
+			Dictionary<SyntaxNode, IEnumerable<string>> methods
+		) {
+			m_root = root;
+			m_methods = methods;
+		}
+
+		public static FileCollector Create(
+			CompilationUnitSyntax root,
+			ImmutableArray<(SyntaxNode Parent, string Source)> methods
```
mthjones, line 70: Why take `string` for the method source here instead of something like `SyntaxNode`/`SyntaxTree`?

mthjones (approved):
Seems reasonable to me as a self-contained piece. Will need to see how it fits into the broader picture though.

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
omsmith, line 70: Parent is _usually_ a type?

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,304 @@
...
+			}
+
+			return m_out.ToString();
+		}
+
+		private bool WriteChildren( SyntaxNode node ) {
+			// Early-out if we have no methods left to place.
+			if( m_methods.Count == 0 ) {
+				return false;
+			}
```
omsmith, line 103: Would this early out ever by appropriate first thing in `CollectSource` as well, or should FileCollector never be instantiated for nothing? Could that be a new BugException in create?

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: In the full pipeline (not evident in this PR unfortunately) the methods are generated in parallel (with some caching) and collected into these files. That collection step causes some synchronization, and it seems to work better to split it up this way. Ultimately we need to get a `string` per file out (as that's what the generator APIs take).

Imagine you're in a file, making edits. Every edit will cause this code to run again, concatenating possibly cached generated sync methods. We need to do this concatenation on every edit, unfortunately, because edits outside of the method can impact the generated file (e.g. you change the usings, or nest/unnest namespaces types etc., basically the stuff we're I'm calling "preamble" isn't cachable within a file easily.)

j3parker: Or namespace or compilation unit

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: I don't believe we'll ever actually get called with 0 methods, but I added a test to assert we output the empty string for that case (we might also output usings.)

This check would cause us to not consider Baz/Quux (not bother with their preambles) in this example:

```cs
partial class Foo {
  partial class Bar {
    [GenerateSync]
    async Task HelloAsync() => await Task.Delay( 3 );

    partial class Baz {}
  }

  partial class Quux {}
}
```

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,304 @@
...
+
+			return true;
+		}
+
+		// Inline assertions to catch some obvious bugs quickly
+		private void CheckPremableInvariant( bool wroteSomething, int countBefore ) {
+			if( wroteSomething && m_preambles.Count != 0 ) {
+				throw new BugException( "Left over preambles" );
+			}
```
omsmith, line 287: Will this stuff work for sibling classes?

```
namespace Foo {

   class X {}

   class Y {}
   
```

j3parker: Yep. So when you are in Foo then countBefore will be 1, and the idea is that after recursing into Foo the only valid values should be 1 (i.e. we did nothing inside X, so we pushed but then discarded `X`s preamble) or 0 (because we flushed them all to write out a method). Loop invariant basically.

j3parker: This should be covered by tests I think, although i was a bit unstructured with it.

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
JeffAshton: Yeah.  Seeing the pooling stuff like this in the Roslyn repo is eye opening

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
omsmith: K, I'm not sure if I'm following it through the logic properly.

If I'm recursed in such that my base preamble count is `1`, then how is it `0` between children?

```csharp
			bool wroteSomething = false;

			// First recurse into child types
			foreach( var child in node.ChildNodes() ) {
				bool wroteSomethingHere = child switch {
					ClassDeclarationSyntax @class => WriteType( @class ),
					StructDeclarationSyntax @struct => WriteType( @struct ),
					InterfaceDeclarationSyntax @interface => WriteType( @interface ),

					// Covers record + record struct:
					RecordDeclarationSyntax @record => WriteType( @record, @record.ClassOrStructKeyword ),

					// Covers file-scoped and block namespaces
					BaseNamespaceDeclarationSyntax @namespace => WriteNamespace( @namespace ),

					// Ignore any other child node
					_ => false
				};

				if( wroteSomethingHere ) {
					wroteSomething = true;
				}

				CheckPremableInvariant( wroteSomething, preambleCountBefore );
			}
```

You're right that the "Complicated" test appears to cover it though.

omsmith: Oh okay right, because preambles goes to 0 when we wrote all of the namespace information, we're not rewriting each time.

K

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
omsmith: > but I added a test to assert we output the empty string for that case (we might also output usings.)

Yeah, I was mostly thinking about that we'd be outputting a file of just usings, which isn't very helpful. But then we shouldn't be doing that anyway.

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
omsmith: Parent of methods? Nah. WriteMethods implies it's just TypeDeclarationSyntax

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,304 @@
...
+			AddPartialTypeToPreamble(
+				decl,
+				keyword1: decl.Keyword,
+				// for "record struct" etc.
+				keyword2: secondaryKeyword,
+				identifier: decl.Identifier,
+				typeParameterList: decl.TypeParameterList,
+				constraintClauses: decl.ConstraintClauses,
+				openBrace: decl.OpenBraceToken
+			);
```
omsmith, line 196: Not desirable to only do this if I or my children have methods?

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: It looks like the bare ones aren't directly under `CompilationUnitSyntax` but instead `GlobalStatementSyntax`: https://sharplab.io/#v2:C4LghgzsA+CwBQABALAAgGIHtMAoCUqA3ggL5A==

j3parker: (not saying this is something that's going to work or that we want to support it (probably not? weird with partials))

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: Yeah. Like e2e I think that won't happen but I'm neutral on blocking it in here (imo intentionally crash or just do something "reasonable" like output an empty file/with usings)

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: Sorry -- I don't follow

omsmith: Okay sorry, I missed that CloseBraceIfNeccesary is throwing this away if it was unused. I didn't read the method because the name seemed simple. Rename to `CloseTypeOrDiscardPreamble` or similar?

omsmith: CloseBlockOr..., rather, I suppose.

j3parker: Went the other direction with `FinishBlock` (vaguer -- it clearly "does something" and you should read it to understand, etc.)

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: I did what you were getting at here and changed the interface to assume that parents are `TypeDeclarationSyntax`. It will mean some more casting/error handling in the upper layer but that is fine. It was going to crash somewhere ultimately, moving it earlier will mean that we separate "you have a method in an unsupported kind of type" from the error that will happen for bare methods/whatever crazy new syntax they come up with next... seems good to me.

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -0,0 +1,305 @@
...
+
+		private void FinishBlock( bool wroteSomething, SyntaxToken? closeBrace ) {
+			if( wroteSomething ) {
+				// Add the closing } if it exists
+				if( closeBrace.HasValue ) {
+					m_out.Append( closeBrace.Value.ToFullString() );
+				}
+			} else {
+				// Throw away the start of the decl that we never wrote
+				m_preambles.RemoveLast();
```
omsmith, line 258: Could consider holding `PieceOfSyntax[]` to pass it into `FinishBlock` to assert that `RemoveLast` is removing what we're expecting.

omsmith (approved):

JeffAshton (approved):

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: Hm, interesting.

omsmith: I mean, as long as it's working and the tests show it is, that's also pretty okay. Just we're asserting some things already, so have it on mind etc etc

j3parker: Yeah. I like the idea/goal of making the invariants more obviously correct. I fiddled a bit now and I couldn't get it to a place where I was totally happy, but I'm also multi-tasking on a Zoom call. I'm going to keep fiddling with it and possibly follow-up.

PR #833, opened 2022-02-03
copy and use AllowedTypeList helper in CustomTestServiceLocatorAnalyzer

omsmith:


j3parker (approved):
You could maybe reference the same .cs file? Weird but 🤷

omsmith:
Oh huh yeah, neat idea.

PR #836, opened 2022-02-03
load NUnitTypes once

omsmith:


src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -58,13 +57,13 @@ internal sealed class CategoryAnalyzer : DiagnosticAnalyzer {
...
 			);
-		}
 
-		private static void OnCompilation( CompilationAnalysisContext context ) {
-			if( !TryLoadNUnitTypes( context.Compilation, out NUnitTypes types ) ) {
-				return;
-			}
+			context.RegisterCompilationEndAction(
+				ctx => OnCompilationEnd( ctx, types )
+			);
```
j3parker, line 63: This definitely isn't a race condition?

j3parker: Doc says end happens after all the non-end things have been invoked, I'm not sure if that includes them running to completion or not (it would make sense, just wondering): https://github.com/dotnet/roslyn/blob/main/docs/analyzers/Analyzer%20Actions%20Semantics.md

omsmith: Sorry, where would the race condition be?

j3parker: if it doesn't wait for the completion of the CompilationStart actions before running CompilationEnd ones

j3parker: I guess we probably do this elsewhere already?

j3parker (approved):

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
omsmith: Yeah we do it elsewhere, this also isn't doing any "collect and then process things", the "OnCompilation" is just to raise diagnostics on the assembly. It could probably actually be `RegisterSymbolAction( SymbolKind.Assembly )`, but I'm not positive.

Either way, I think `RegisterCompilationEnd` from within `CompilationStart` is functionally equivalent to `RegisterCompilation` from within Initialize?

PR #838, opened 2022-02-03
Add AsyncToSyncMethodTransformer

j3parker:


src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -0,0 +1,135 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace D2L.CodeStyle.Analyzers.Async.Generator;
+
+internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
```
j3parker, line 7: This file is where the fun is going to happen. It will be a lot bigger ultimately (as it will need to handle every kind of syntax) but from my prototyping I think it can be really dumb/stateless.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -0,0 +1,135 @@
+﻿using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace D2L.CodeStyle.Analyzers.Async.Generator;
+
+internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
+	public AsyncToSyncMethodTransformer( SemanticModel model ) : base( model ) { }
```
j3parker, line 8: Right now this is based on `SemanticModel`. It uses that to do stuff like look up types to check if they are `Task`. There's only a bit of that in here right now though.

If this becomes a performance issue I think it will be possible to revisit it, but I'm interested in other people's feelings.

This will run once per method concurrently in the incremental generator pipeline and there will be caching inside VS, but of course not on command line builds.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -0,0 +1,135 @@
...
+		TransformAll( attributeList.Attributes, ReplaceGenerateSyncAttribute )
+	);
+
+	private AttributeSyntax ReplaceGenerateSyncAttribute(
+		AttributeSyntax attribute
+	) {
+		// Attributes other than [GenerateSync] pass through
+		if( !SyncGenerator.IsGenerateSyncAttribute( m_model, attribute ) ) {
+			return attribute;
+		}
```
j3parker, line 38: Usage of the SemanticModel. This one can maybe go away but it isn't the only usage.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -0,0 +1,135 @@
...
+
+		const int AsyncSuffixLength = 5;
+
+		return SyntaxFactory.Identifier(
+			ident.ValueText.Substring( 0, ident.ValueText.Length - AsyncSuffixLength )
+		).WithTriviaFrom( ident );
+	}
+
+	private TypeSyntax TransformReturnType( TypeSyntax returnType ) {
+		var returnTypeInfo = m_model.GetTypeInfo( returnType );
```
j3parker, line 72: Another usage of `SemanticModel`. This one feels more important? If we only do it by string then we won't handle all the cases, but maybe that's fine. If you're using `[GenerateSync]` you should just be reasonable?

I'm on the fence about this. WDYT?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -0,0 +1,135 @@
...
+			SyntaxFactory.ThrowStatement(
+				SyntaxFactory.Token( SyntaxKind.ThrowKeyword )
+					.WithTrailingTrivia( SyntaxFactory.Space ),
+				SyntaxFactory.LiteralExpression(
+					SyntaxKind.NullLiteralExpression,
+					SyntaxFactory.Token( SyntaxKind.NullKeyword )
+				),
+				SyntaxFactory.Token( SyntaxKind.SemicolonToken )
+			)
+		).WithTriviaFrom( block );
```
j3parker, line 133: Just temporary to demonstrate stuff with tests.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
omsmith: SyncGenerator hasn't been introduced yet.

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+[TestFixture]
+internal sealed class AsyncToSyncMethodTransformerTests {
+	[Test]
+	public void BasicTask() {
+		var actual = Transform( @"[GenerateSync] async Task HelloAsync( int abc, string def ) { await Task.Delay( 3 ); }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Hello( int abc, string def ) {throw null;}", actual.Value.ToFullString() );
```
j3parker, line 20: There is a way to compare syntax trees, but `Assert.AreEqual` on strings has way better error messages when the test fails (shows you the first difference). Using the `SyntaxFactory.AreEquivalent` or `.IsEquivalentTo` results in opaque true/false style errors.

j3parker: The downside is that these assertions are sensitive to whitespace, e.g. `{throw null;}` because I didn't bother formatting that nicely. But I'm trying to be somewhat careful about mirror the source whitespace so that's maybe not so bad.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
j3parker: Yeah that's why this is draft--just working out the kinks before adding reviewers. 😄

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -97,15 +97,28 @@ ImmutableArray<Diagnostic> Diagnostics
 	/// </summary>
 	protected SyntaxTokenList TransformAll(
 		SyntaxTokenList input,
-		Func<SyntaxToken, SyntaxToken> transformer
```
j3parker, line 100: Eh I screwed up with this stuff. Went back to having a separate impl for `SyntaxToken?` and using `null` to single filtering out the token. Copy and paste but whatever.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
omsmith: I would personally run with it as-is. It will be hard to actually profile until we have usage too.

Given we're planning to use this a lot and for a while, it will eventually be important to figure out the most performant but functional thing though.

j3parker: Yeah. The cost will be zero to start because this (and the thing that will call it) only runs if you have something to transform, so it's at least an incremental cost (vs. our analyzers where you often pay a lot up-front)

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
omsmith: I don't think it would be that crazy to require "Task" => "void" and "Task<...>".Match( /Task<(.+)>/)[1]. Maybe eventually add ValueTask as well. Any custom awaitable support would be insane for this to handle probably. Maybe just make the diagnostic a little more specific: "Expected return type to be exactly 'Task' or 'Task<...>' (no qualified names)."?

omsmith: Said simpler, I think I would okay with these being stringy as long as the diagnostic messaging is clear.

I'm also okay with it using the model right now, echoing the previous sentiment.

j3parker: Yep, I already have that diagnostic below and definitely wasn't intended to support anything more general than the main task types.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -0,0 +1,146 @@
...
+		ArrowExpressionClauseSyntax? body
+	) {
+		if( body == null ) {
+			return null;
+		}
+
+		GeneratorError(
+			body.GetLocation(),
+			"ArrowExpressionClauseSyntax is not supported"
+		);
```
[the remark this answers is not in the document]
j3parker: Yeah for sure.

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
omsmith: I believe we do AreEquivalent in StaticDI and get the trees output in the message?

omsmith: I'm find with them being stringy though, given you do have exact goals.

omsmith (approved):

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
j3parker: you just get true/false :'(

omsmith: K, in StaticDI we do: `Assert.Equal( expected, actual, SyntaxNodeComparer.Instance )` where SyntaxNodeComparer uses ` SyntaxFactory.AreEquivalent( x, y )` and get outputs.

j3parker: Oh right, what I disliked about that one is still problematic in here: it outputs the diff as a rendered string.

I used to have a version that asserted each node on the SyntaxNode structure recursively, accounting for whitespace.

The reason it mattered in static DI was that we hand SyntaxTrees to the compiler, and you can construct syntax trees that look good when rendered to a string but will cause the compiler to crash in odd ways.

Here, we will ultimately be handing strings to the compiler anyway (weird API). So I don't _really_ care about that stuff. Assert.AreEqual would probably work fine here, you're right.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -0,0 +1,146 @@
...
+
+		const int AsyncSuffixLength = 5;
+
+		return SyntaxFactory.Identifier(
+			ident.ValueText.Substring( 0, ident.ValueText.Length - AsyncSuffixLength )
+		).WithTriviaFrom( ident );
+	}
+
+	private TypeSyntax TransformReturnType( TypeSyntax returnType ) {
+		var returnTypeInfo = m_model.GetTypeInfo( returnType );
```
JeffAshton, line 72: Pass CancelationToken to GetTypeInfo

JeffAshton: It's this running within  the callback to RegisterSourceOutput?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -0,0 +1,146 @@
...
+					SyntaxKind.NullLiteralExpression,
+					SyntaxFactory.Token( SyntaxKind.NullKeyword )
+				),
+				SyntaxFactory.Token( SyntaxKind.SemicolonToken )
+			)
+		).WithTriviaFrom( block );
+	}
+
+	private bool IsGenerateSyncAttribute( AttributeSyntax attribute ) {
+		var attributeConstructorSymbol = m_model.GetSymbolInfo( attribute ).Symbol as IMethodSymbol;
```
JeffAshton, line 137: CancellationToken

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
j3parker: No, before that

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -0,0 +1,149 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace D2L.CodeStyle.Analyzers.Async.Generator;
+
+internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
+	public AsyncToSyncMethodTransformer(
+		SemanticModel model,
+		CancellationToken token
+	) : base( model, token ) { }
```
j3parker, line 11: I take the cancellation token in the constructor so that it won't have to be based down recursively (that's mostly the purpose of this class -- there is going to be _a lot_ of recursion here.)

See note below RE: tests and cancellation tokens.

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -0,0 +1,104 @@
...
+
+	public static TransformResult<MethodDeclarationSyntax> Transform(
+		string methodSource,
+		CancellationToken? cancellationToken = null
+	) {
+		var (compilation, methodDecl) = ParseMethod( methodSource );
+
+		var transformer = new AsyncToSyncMethodTransformer(
+			compilation.GetSemanticModel( methodDecl.SyntaxTree ),
+			cancellationToken ?? CancellationToken.None
```
j3parker, line 62: So I went to go write tests and unfortunately starting a transformation with a pre-cancelled token didn't do anything. I think because a lot of code paths lead to [this method](https://sourceroslyn.io/#Microsoft.CodeAnalysis.CSharp/Compilation/MemberSemanticModel.cs,9471331dd9be3500,references) which ignores its cancellation token.

I still think it's worth threading through for good form at least, and when the AsyncToSyncMethodTransformer gets bigger there will probably be reasonable places to check manually for cancellation (maybe every block or something?) so I left it in (including this unused support in tests.)

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
j3parker: This buddy helps turn an `IncrementalValuesProvider<(MethodDeclarationSyntax, Compilation)>` into an `IncrementalValuesProvider<MethodGenerationResult>` (that type isn't defined yet, but it's the `string` source for one method or diagnostics). From there we `Collect()` them into one big batch and sort them by source file, and then `FileCollector` (previous PR) concatenates them into one file creating an `IncrementalValuesProvider<FileGenerationResult>` (another undefined type -- the `string` for the source of an entire output file + all the associated diagnostics). That part is fed into `RegisterSourceOutput` and the callback there emits the file source (which is a string) and all of the associated diagnostics.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
```diff
@@ -11,13 +11,16 @@ namespace D2L.CodeStyle.Analyzers.Async.Generator;
...
+
 	private readonly ImmutableArray<Diagnostic>.Builder m_diagnostics = ImmutableArray.CreateBuilder<Diagnostic>();
 
 	/// <summary>
 	/// A name for the generator that will be shown to the user if we crash
 	/// </summary>
-	public SyntaxTransformer( SemanticModel model ) {
+	public SyntaxTransformer( SemanticModel model, CancellationToken token ) {
 		m_model = model;
+		m_token = token;
```
JeffAshton, line 23: Taken but not used?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -51,11 +51,15 @@ internal sealed class AsyncToSyncMethodTransformerTests {
 		);
 	}
 
-	public static TransformResult<MethodDeclarationSyntax> Transform( string methodSource ) {
+	public static TransformResult<MethodDeclarationSyntax> Transform(
+		string methodSource,
+		CancellationToken? cancellationToken = null
```
JeffAshton, line 56: Just omit for now

JeffAshton (dismissed):

src/D2L.CodeStyle.Analyzers/Async/Generator/SyntaxTransformer.cs
j3parker: Yeah, it's used below (although to no effect currently, at least in NUnit. Maybe in VS 🤷)

omsmith (approved):

PR #839, opened 2022-02-04
Jashton/report diagnostic extensions

JeffAshton:
Replacing the `context.ReportDiagnostic( Diagnostic.Create( ... ) )` pattern with a single method per context type.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Extensions/ReportDiagnosticExtensions.cs
+++ b/src/D2L.CodeStyle.Analyzers/Extensions/ReportDiagnosticExtensions.cs
@@ -1,0 +1,188 @@
+﻿using System.Collections.Immutable;
+
+namespace Microsoft.CodeAnalysis.Diagnostics {
+
+	internal static class ReportDiagnosticExtensions {
+
+		#region CompilationAnalysisContext
+
+		public static void ReportDiagnostic(
+				this CompilationAnalysisContext context,
+				DiagnosticDescriptor descriptor,
+				Location? location,
+				params object?[]? messageArgs
+			) {
+
+			ReportDiagnostic(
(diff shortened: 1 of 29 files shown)


src/D2L.CodeStyle.Analyzers/Extensions/ReportDiagnosticExtensions.cs
```diff
@@ -0,0 +1,113 @@
...
+
+	internal static class ReportDiagnosticExtensions {
+
+		public static void ReportDiagnostic(
+				this SourceProductionContext context,
+				DiagnosticDescriptor descriptor,
+				Location? location = null,
+				IEnumerable<Location>? additionalLocations = null,
+				ImmutableDictionary<string, string?>? properties = null,
+				object?[]? messageArgs = null
```
JeffAshton, line 13: By not having `params` on `messageArgs`, you only need one method.

omsmith (approved):
I don't mind the existing API, but don't feel strongly either way.

@j3parker ?

JeffAshton:
> I don't mind the existing API, but don't feel strongly either way.
> 
> @j3parker ?

The two call boiler plate has always annoyed me.  I snapped last night adding the additional DTO diagnostics.  Took me a bit to realize that the thing that made their API so shitty was the `params` on `messageArgs`.  That pattern works ok for something like String.Format where there is only one parameter before the `params`, but it gets dangerous as that set grows.  `params` is the original default parameter of `null`.

Roughly the same amount of characters on the screen when you are using the `messageArgs`, but of high value then `Diagnostic.Create` because the named parameter increases readability by grouping them together.

image

j3parker (approved):

PR #840, opened 2022-02-04
Jashton/explicit cancellation token argument analyzer

JeffAshton:


src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -705,5 +705,14 @@ public static class Diagnostics {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor ExplicitCancellationTokenArgumentRequired = new DiagnosticDescriptor(
+			id: "D2LXXXX",
+			title: "CancellationToken arguments must be explicitly passed",
+			messageFormat: "CancellationToken arguments must be explicitly passed. CancellationToken.None is acceptable in some situations, but in most case should be passed through an entire async chain to ensure our applications are responsive.",
+			category: "Correctness",
+			defaultSeverity: DiagnosticSeverity.Error,
```
omsmith, line 714: Viable to introduce as Error today?

JeffAshton: Fixing them before I merge

src/D2L.CodeStyle.Analyzers/Async/ExplicitCancellationTokenArgumentAnalyzer.cs
```diff
@@ -0,0 +1,72 @@
...
+				return;
+			}
+
+			context.RegisterOperationAction(
+					context => {
+						IArgumentOperation argument = (IArgumentOperation)context.Operation;
+						AnalyzeArgument( context, argument, cancellationTokenType );
+					},
+					OperationKind.Argument
+				);
```
j3parker, line 40: Try to conform to the code formatting in the rest of this repo

j3parker (approved):
Talking on Slack about just using the one from roslyn-analyzers instead, so that we don't need to maintain our own (and get the freebies like code fixes.) It's the "opposite" approach of forcing you to thread the ct when possible vs requiring that you pass something (maybe CancellationToken.None).

It would mean its harder to find the places where we don't do cancellation, but on the other hand the UX would be better and the code fix encourages you to do the correct thing.

If we had APIs that required you to take a cancellation token at the top (e.g. handling web requests) and required them in key places at the bottom (the DB) I think the roslyn-analyzers approach would work well.

JeffAshton:
> Talking on Slack about just using the one from roslyn-analyzers instead, so that we don't need to maintain our own (and get the freebies like code fixes.) It's the "opposite" approach of forcing you to thread the ct when possible vs requiring that you pass something (maybe CancellationToken.None).
> 
> It would mean its harder to find the places where we don't do cancellation, but on the other hand the UX would be better and the code fix encourages you to do the correct thing.
> 
> If we had APIs that required you to take a cancellation token at the top (e.g. handling web requests) and required them in key places at the bottom (the DB) I think the roslyn-analyzers approach would work well.

Tried using the following:

[CA2016: Forward the CancellationToken parameter to methods that take one](https://docs.microsoft.com/en-ca/dotnet/fundamentals/code-analysis/quality-rules/ca2016)

However, as the much states, "to methods that take one", is not the same thing as this analyzer.  If you accidentally forget to pass a `CancellationToken` to a method in the chain, it allows it.

PR #845, opened 2022-02-07
Add SyncGenerator source generator

j3parker:


src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+		var methodsByFile = methodResults
+			.GroupBy( static mr => mr.Original.SyntaxTree.GetCompilationUnitRoot() )
+			.OrderBy( static mr => mr.First().Original.SyntaxTree.FilePath )
+			.ToImmutableArray();
+
+		var hintNames = SourceFilePathHintNameGenerator.Create(
+			// We're assuming every root has a tree here -- when is that false?
+			// Generated code?
+			methodsByFile.Select( static grp => grp.Key.SyntaxTree.FilePath ).ToImmutableArray()
+		).GetHintNames();
```
j3parker, line 134: We talked about the better way to do relative path stuff on a chat today (getting the projects base path).
I can fix that up separately in another PR (inside `SourceFilePathHintNameGenerator`)

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -71,21 +71,24 @@ class TestType{{{methodSource}}}" );
...
+		assemblyName: "test",
+		references: References.Value,
+		syntaxTrees: trees,
+		options: new(
+			// Not actually going to write the output
+			outputKind: OutputKind.DynamicallyLinkedLibrary,
+			warningLevel: 0,
+			optimizationLevel: OptimizationLevel.Debug
+		)
+	);
```
j3parker, line 90: Hacking out a helper method here 🤷

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/SyncGeneratorTests.cs
```diff
@@ -0,0 +1,180 @@
...
+using Microsoft.CodeAnalysis.CSharp;
+using NUnit.Framework;
+
+namespace D2L.CodeStyle.Analyzers.Tests.Async.Generator;
+
+/// <summary>
+/// Integration tests for the generator as a whole.
+/// Prefer unit tests of AsyncToSyncTransformer and FileCollector as they are
+/// simpler. We just want a bit of smoke-testing here.
+/// </summary>
```
j3parker, line 12: I would change my tune on this if we had something as good as spec tests for generators, but I'm not feeling the urge to rush into that right now personally.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.Results.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+
+namespace D2L.CodeStyle.Analyzers.Async.Generator;
+
+internal sealed partial class SyncGenerator {
+	// All of the intermediate results generated in the incremental pipeline.
+	// All of these should implement value-based equality. So use records and
+	// make sure the things they hold types we hold implement equality.
```
omsmith, line 10:
```suggestion
	// make sure the things they hold implement equality.
```

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+		CancellationToken _
+	) {
+		if( node is not MethodDeclarationSyntax method ) {
+			return false;
+		}
+
+		// We only care about async methods
+		if( !method.Modifiers.Any( t => t.IsKind( SyntaxKind.AsyncKeyword ) ) ) {
+			return false;
+		}
```
omsmith, line 52: I could imagine non-`async` Async methods being easy as well, but, good start regardless.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+namespace D2L.CodeStyle.Analyzers.Async.Generator;
+
+[Generator]
+internal sealed partial class SyncGenerator : IIncrementalGenerator {
+	public void Initialize( IncrementalGeneratorInitializationContext context ) {
+		// Collect all the methods we want to generate sync versions for
+		// individually (for better incremental builds)
+		IncrementalValuesProvider<(MethodDeclarationSyntax, Compilation)> methodsToGenerate =
+			context.SyntaxProvider
+			   .CreateSyntaxProvider( IsInterestingLookingMethod, ExtractSyntax )
```
omsmith, line 15: For my reading benefit, not knowing the APIs uber well

```suggestion
			   .CreateSyntaxProvider( predicate: IsInterestingLookingMethod, transform: ExtractSyntax )
```

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
j3parker: Yeah.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+
+		// We only care about async methods
+		if( !method.Modifiers.Any( t => t.IsKind( SyntaxKind.AsyncKeyword ) ) ) {
+			return false;
+		}
+
+		// The ones we care about have to have a particular attribute -- so
+		// perk up when we see _any_ attributes (we don't have a SemanticModel
+		// in this method.)
+		return method.AttributeLists.Any();
```
omsmith, line 57: Could consider doing a quick check for `attribute.Name.ToString()` being equal to one of `"GenerateSync"` or `"GenerateSyncAttribute"` before going to the model to verify in the transform. Believe that's what Jeff has done the the DtoGenerator.

It's more restrictive than only-model (requiring specific strings in the syntax), but at least still correct by backing up the string check with the symbol lookup later.

j3parker: I mostly based this off the Microsoft ones. The predicate is just supposed to be a cheap filter.
Doing it with strings is less correct because of aliases (`using X = D2L.CodeStyle.Annotations.GenerateSync`).
I don't care that much but I figure that's why it's typically done this way.

omsmith: Actually no, Jeff just does `unqualifiedNames.Contains( unqualifiedNameOfAttribute )`, so not less restrictive. That could be an EndsWith. https://github.com/Brightspace/D2L.CodeStyle/blob/master/src/D2L.CodeStyle.Analyzers/Language/AttributeAliasesAnalyzer.cs being added makes that ultimately good and correct.

j3parker: Example predicate: https://github.com/dotnet/runtime/blob/80ba698a41cafa092a12a18119ae4a1c57347d64/src/libraries/System.Text.RegularExpressions/gen/RegexGenerator.Parser.cs#L24-L26

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+		var model = data.Compilation.GetSemanticModel( data.MethodDeclaration.SyntaxTree );
+		var methodSymbol = model.GetDeclaredSymbol( data.MethodDeclaration, cancellationToken );
+
+		if( methodSymbol == null || methodSymbol.Kind == SymbolKind.ErrorType ) {
+			return null;
+		}
+
+		var transformer = new AsyncToSyncMethodTransformer( model, cancellationToken );
+
+		var transformResult = transformer.Transform( data.MethodDeclaration );
```
omsmith, line 112: Still just `throw null` right?

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+
+		foreach( var file in files ) {
+			if( cancellationToken.IsCancellationRequested ) {
+				yield break;
+			}
+
+			var generatedMethods = ImmutableArray.CreateBuilder<(TypeDeclarationSyntax, string)>();
+			var diagnostics = ImmutableArray.CreateBuilder<Diagnostic>();
+
+			foreach( var method in file.Data ) {
```
omsmith, line 149: For deterministic ordering within the file?

```suggestion
			foreach( var method in file.Data.OrderBy( m => m.Original.GetLocation().SourceSpan ) ) {
```

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
j3parker: It's only correct with that analyzer, but I don't know that I'd call it _good_.
Anyway, I'm just doing what others do here.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
j3parker: Yes, see tests

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+				diagnostics.AddRange( method.Diagnostics );
+				generatedMethods.Add( ((method.Original.Parent as TypeDeclarationSyntax)!, method.GeneratedSyntax ) );
```
omsmith, line 151: Should this just be a cast?

```suggestion
				generatedMethods.Add( ((TypeDeclarationSyntax)method.Original.Parent, method.GeneratedSyntax ) );
```

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
omsmith: Would mostly like to see _us_ do it one way. I'm happy to defer that to follow-ups in either direction though.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+
+		return null;
+	}
+
+	internal static bool IsGenerateSyncAttribute(
+		SemanticModel model,
+		AttributeSyntax attribute,
+		CancellationToken cancellationToken
+	) {
+		var attributeConstructorSymbol = model.GetSymbolInfo( attribute, cancellationToken ).Symbol as IMethodSymbol;
```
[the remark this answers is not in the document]
JeffAshton: Not sure where they use it

https://github.com/dotnet/roslyn/blob/2b7f137ebbfdf33e9eebffe87d036be392815d2b/src/Compilers/CSharp/Portable/Symbols/Source/QuickAttributeChecker.cs#L99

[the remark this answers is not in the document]
j3parker: I disagree, unless we measure a performance problem here.
The semantic model in this method comes "for free", it's not created on first use.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
j3parker: I like the idea but I don't want to add this until I can reproduce the problem/see Microsoft doing similar things/etc.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+	private static IEnumerable<FileGenerationResult> GenerateFiles(
+		ImmutableArray<MethodGenerationResult> methodResults,
+		CancellationToken cancellationToken
+	) {
+		var methodsByFile = methodResults
+			.GroupBy( static mr => mr.Original.SyntaxTree.GetCompilationUnitRoot() )
+			.OrderBy( static mr => mr.First().Original.SyntaxTree.FilePath )
+			.ToImmutableArray();
+
+		var hintNames = SourceFilePathHintNameGenerator.Create(
```
omsmith, line 130: nit: Had to look it up when reading. Also needs a using if accepted.

```suggestion
		IEnumerable<string> hintNames = SourceFilePathHintNameGenerator.Create(
```

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+
+		var files = methodsByFile.Zip(
+			hintNames,
+			static ( fileData, name ) => ( HintName: name, Data: fileData )
+		);
+
+		foreach( var file in files ) {
+			if( cancellationToken.IsCancellationRequested ) {
+				yield break;
+			}
```
omsmith, line 144: I believe it was determined msbuild expects and handles this?

```suggestion
			cancellationToken.ThrowIfCancellationRequested();
```

j3parker: It does, but I don't need to throw here.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
j3parker: Not with nullable on apparently. What I put is more explicit about discarding the `?` I guess.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
omsmith: Under the assumption that roslyn will stop processing the enumerable produced by the SelectMany? Doesn't that same logic mean we don't need to check this at all, given we're a generator?

omsmith: Err, nvm for the second part that, I suppose generators aren't a state machine in that way in .NET. Had my implementations mixed up.

j3parker: `yield break` always stops enumeration. Generators are state machines, [here's a decompiled example that shows this](https://sharplab.io/#v2:D4AQDABCCMAsDcBYAUCkBmCBDARgZwBcAnLAYwKgCYIBhCAbxQmak10JPIhwHseAbCAGUAFjwCu/ACZCCPAA4AKCFIk5+AUwgAPCAEokyFhCYsMUaOgA8q8eo0A+CAFlFehqePNb9iAEsIAF4IMEMvFgB3ET9NZWJxLXdGI3CvGGgoAHZ/MNTjAIBqYPQAOjhcvOY/ADNlUQlpWQVlAPckz0rjdO4iDSwAawrKgF8O41GU5gnhoA)

It could be the case that the thing processing our resulting `IEnumerable<T>` (which is being transformed into an `IncrementalValuesProvider<T>`) also checks for cancellation, but there is "no way" for me to know that for sure.

It could be the case that their code doesn't check for cancellation at all, and that further parts of the pipeline continue to go on whatever comes out of our `IEnumerable<T>`, but that would be Microsoft's bug I believe.

The exception would be handy in places where we can't easily stop, e.g. inside `AsyncToSyncMethodTransformer` its going to be infeasible to stop the recursion, but in general control flow via exceptions kinda stinks.

j3parker: When you do absolutely want to throw is when you want to communicate in the standard way that the task was cancelled to the caller, but that's not really our responsibility here as far as I know, it's the frameworks.

omsmith: Sorry, I didn't communicate well since I was thinking through it in my head at the same time.

```csharp
		IncrementalValuesProvider<FileGenerationResult> generatedFiles =
			generatedMethods.Collect()
				.SelectMany( GenerateFiles );
```

What I was thinking through here is that while yes, `yield break` stops one invocation, it does not further invocations for other elements which would still do work.

However, SelectMany isn't Linq in this case, and it's actually a _1 => many (or none) transform_ (as per the comment in https://sourceroslyn.io/#Microsoft.CodeAnalysis/SourceGeneration/Nodes/ValueSourceExtensions.cs,5fdbcdb2b8308e69). Since it's a source of 1, there's no other invocations to make, so yes I agree it should be about equivalent.

Now, that was thinking I had to do though.

As far as I can tell, `ThrowIfCancellationRequested` is the idiom for cancellationTokens. `IsCancellationRequested` is there to check should you need to do cleanup / revert work before proceeding to ultimately call `ThrowIfCancellationRequested`.

See the example here, for instance: https://docs.microsoft.com/en-us/dotnet/standard/parallel-programming/task-cancellation

I would suggest just using `ThrowIfCancellationRequested` unless we've an explicit reason to do otherwise.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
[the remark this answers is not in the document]
JeffAshton: Ran the profiler.  Looks like because of this condition:

```
!method.Modifiers.Any( t => t.IsKind( SyntaxKind.AsyncKeyword ) )
```

And there are almost no `async` methods in the LMS, we won't have an impact.  When I remove this condition and profile, it joins the list of our worst offenders:

image

So it won't be a problem until async gets used more.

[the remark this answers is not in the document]
JeffAshton: Using the UnqualifiedName as a comparison, the average time is 612 ticks.  

image

Where as before with the `GetSymbolInfo` it was an average of 2910 ticks.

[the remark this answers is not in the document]
JeffAshton: Scanned the lms.  Only 3 existing async methods with attributes;

* https://search.d2l.dev/xref/lms/lp/framework/core/D2L.LP/LP/Aws/IoT/AwsIotWebSocket.cs?r=53ed0b24#15
* https://search.d2l.dev/xref/lms/platformTools/diagnostics/D2L.LP.Diagnostics.Web/Application/AsyncTestController.cs?r=12a8200d#26
* https://search.d2l.dev/xref/lms/platformTools/diagnostics/D2L.LP.Diagnostics.Web/Application/AsyncTestController.cs?r=12a8200d#37

So all concerns gone.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	internal static bool IsGenerateSyncAttribute(
+		SemanticModel model,
+		AttributeSyntax attribute,
+		CancellationToken cancellationToken
+	) {
+		var attributeConstructorSymbol = model.GetSymbolInfo( attribute, cancellationToken ).Symbol as IMethodSymbol;
+
+		if( attributeConstructorSymbol == null ) {
```
JeffAshton, line 87: Visual studio's recommendation isn't horrible.  The `is not` definitely reads nice:

image

JeffAshton (dismissed):

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -0,0 +1,182 @@
...
+		if( !data.Compilation.ContainsSyntaxTree( data.MethodDeclaration.SyntaxTree ) ) {
+			return null;
+		}
+
+		var model = data.Compilation.GetSemanticModel( data.MethodDeclaration.SyntaxTree );
+		var methodSymbol = model.GetDeclaredSymbol( data.MethodDeclaration, cancellationToken );
+
+		if( methodSymbol == null || methodSymbol.Kind == SymbolKind.ErrorType ) {
+			return null;
+		}
```
mthjones, line 108: Are these all guards for "the compilation is otherwise broken, so ignore"? Or are there other scenarios where we might expect these to happen?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/SyncGeneratorTests.cs
```diff
@@ -0,0 +1,180 @@
...
+	public void OneFileWithOneMethod() {
+		var result = RunGenerator( @"
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using D2L.CodeStyle.Annotations;
+
+namespace Foo;
+
+sealed class Bar {
```
mthjones, line 50: nit: `partial`? Is this (or will this be) caught by a separate analyzer that isn't running in these tests?

```suggestion
sealed partial class Bar {
```

mthjones (dismissed):

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
j3parker: I think only compilation broken, and honestly I'm not even sure in what scenarios it would happen there. `null` is possible by the contract, `ErrorType` honestly I sprinkle around when I remember to (that one definitely means "compile is broken but we're still running things anyway" and its safe to do whatever basically.)

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/SyncGeneratorTests.cs
j3parker: Ah good eye. So it won't be -- you'll get the standard error about "this needs to be partial" in the _source_ file which is actually kinda nice.

It would be reasonable to use `partial` in the test case but its not the worst to accidentally/indirectly test this I guess. 🤷

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
j3parker: Ya I'm cool with the english, I just haven't gotten it in my muscle memory yet

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
j3parker: We talked more on Zoom -- I'm still not 100% sold that this isn't allowed ala the first bullet point, and that it's the generators responsibility to not cache things mid cancellation, but it's at least not totally clear. I did notice that they check the tokens before they advance our iterator, but I don't want to trace around to see what happens _after_, and even so it's not clear thats a good idea (relying on implementation details? I'm not sure.)

Will switch to throw.

omsmith (approved):

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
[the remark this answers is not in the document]
j3parker: Ok, will fix in a follow-up.

PR #857, opened 2022-02-08
Adding test case for calling via interface

JeffAshton:


tests/D2L.CodeStyle.Analyzers.Test/Specs/OnlyVisibleToAnalyzer.cs
```diff
@@ -234,6 +234,28 @@ public static class AlwaysVisibleWithinSameContainer {
...
+		public static void Caller() {
+
+			// Ok when called on this
+			this.RestrictedMethod();
+
+			// Violation when called via interface, since not in set
+			IRestrictedInterface @interface = this;
+			/* MemberNotVisibleToCaller(RestrictedMethod) */ @interface.RestrictedMethod() /**/;
+		}
+	}
```
omsmith, line 258:
```suggestion
	public class RestrictedMembersOnSelf : IRestrictedInterface {

		[OnlyVisibleToType( typeof( string ) )]
		public void RestrictedMethod() {}

		public void Caller() {

			// Ok when called on this
			this.RestrictedMethod();

			// Violation when called via interface, since not in set
			IRestrictedInterface @interface = this;
			/* MemberNotVisibleToCaller(RestrictedMethod) */ @interface.RestrictedMethod() /**/;
		}
	}
```

omsmith (approved):

PR #863, opened 2022-02-09
Add minimal support for sync transformation of expressions and statements

j3parker:
This is the first "fun" PR.

It still isn't anything useful, but future PRs I envision adding syntax slowly/partly as needed, and those changes will typically involve additions to the expression/statement list


src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -105,37 +105,43 @@ AttributeSyntax attribute
...
 
+	private StatementSyntax Transform( StatementSyntax stmt )
+		=> stmt switch {
+			ExpressionStatementSyntax exprStmt =>
+				exprStmt.WithExpression( Transform( exprStmt.Expression) ),
+
+			ReturnStatementSyntax returnStmt =>
+				returnStmt.Expression is null
+				? returnStmt
+				: returnStmt.WithExpression( Transform( returnStmt.Expression ) ),
```
j3parker, line 133: Most things will look like this: handle a case by recursing and adding the result back in with WithX.

In practice I found very few things would need more intelligent handling. There is already support for `Task<T>` -> `T` in this file, the real big thing is around await expressions and some interesting stuff with async Lambdas (if we choose to eventually support that, even).

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -13,20 +13,29 @@ namespace D2L.CodeStyle.Analyzers.Tests.Async.Generator;
 internal sealed class AsyncToSyncMethodTransformerTests {
 	[Test]
 	public void BasicTask() {
-		var actual = Transform( @"[GenerateSync] async Task HelloAsync( int abc, string def ) { await Task.Delay( 3 ); }" );
+		var actual = Transform( @"[GenerateSync] async Task HelloAsync( int abc, string def ) { return; }" );
```
j3parker, line 16: I would like the tests to have decent looking examples, but now that we aren't transforming things into `throw null` we can only use what we support.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -146,4 +152,14 @@ AttributeSyntax attribute
...
 			== "D2L.CodeStyle.Annotations.GenerateSyncAttribute";
 	}
+
+	public T UnhandledSyntax<T>( T node ) where T : SyntaxNode {
+		GeneratorError(
+			node.GetLocation(),
+			$"unhandled syntax kind: {node.Kind()}"
+		);
+
+		return node;
```
j3parker, line 162: We're forced by the nature of the recursion to return _something_, so we will just return the source text unmodified and stop recursing.

It will look a little weird in the generated output, but doing something else in all the cases would be a lot more work I think. It would possibly need to be different for every overload at least, e.g. `throw null` works for ExpressionSyntax, and if you wrap it in ExpressionStatement it could work for StatementSyntax, but etc. etc.

j3parker: We could also throw, but then that would stop adjacent recursions leading to a poor UX (you only get one diagnostic, which you fix and then get another diagnostic, repeat.... vs. just giving you all the diagnostics we reasonably can up-front.)

JeffAshton (approved):

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -105,37 +105,43 @@ AttributeSyntax attribute
...
-		).WithTriviaFrom( block );
+		var transformed = TransformAll( block.Statements, Transform );
+
+		return block.WithStatements( transformed );
 	}
 
+	private StatementSyntax Transform( StatementSyntax stmt )
+		=> stmt switch {
+			ExpressionStatementSyntax exprStmt =>
+				exprStmt.WithExpression( Transform( exprStmt.Expression) ),
```
omsmith, line 128:
```suggestion
				exprStmt.WithExpression( Transform( exprStmt.Expression ) ),
```

omsmith (approved):

omsmith (approved):

JeffAshton (approved):

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
j3parker:
```cs
ReturnStatementSyntax returnStmt => returnStmt
    .WithExpression( MaybeTransform( returnStmt.Expression, Transform ) ),

PR #864, opened 2022-02-09
adjust for 0-based line numbers in PrettyDiagnostic output

omsmith:


j3parker (approved):

JeffAshton (approved):

src/D2L.CodeStyle.SpecTests/SpecTestFixtureBase.cs
```diff
@@ -336,7 +336,7 @@ string Message
 
 				sb.AppendLine( "{" );
 				WriteProperty( nameof( Id ), Id );
-				WriteProperty( nameof( LinePosition ), LinePosition );
+				WriteProperty( nameof( LinePosition ), $"({LinePosition.Start.Line + 1},{LinePosition.Start.Character})-({LinePosition.End.Line + 1},{LinePosition.End.Character})" );
```
JeffAshton, line 339: Does the end line character need a bump to match the leading space of ` /**/`

omsmith: Both lines and characters are accurate to where the comment delimiters are. Lines are 0 based though so it's often just confusing to readers, when the editor starts at 1.

There's a TODO in the parser itself for trying to trim to the nearest actual token when computing the character positions though, which is where I'd try to do that (and not just do +/-1):

https://github.com/Brightspace/D2L.CodeStyle/blob/9f8d3d2287ee4806faad5d74a6c9fc07f5ac6a95/src/D2L.CodeStyle.SpecTests/SpecTestFixtureBase.cs#L272-L273

PR #865, opened 2022-02-09
Support interface/abstract method signatures in SyncGenerator

j3parker:


src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -46,8 +46,17 @@ CancellationToken _
...
+		// TODO: add an analyzer to alert about uses of [GenerateSync] that this would filter out
+		if( !method.Identifier.ValueText.EndsWith( "Async", StringComparison.Ordinal ) ) {
+			return false;
+		}
+
+		// Require the async keyword if applicable.
+		// Interface and abstract methods won't have bodies and thus won't have the
+		// keyword but we still want to translate their signatures.
+		// TODO: we could probably relax this, or we should have an error.
+		if( method.Body is not null && !method.Modifiers.Any( t => t.IsKind( SyntaxKind.AsyncKeyword ) ) ) {
```
j3parker, line 59: Two things:

* I noticed that this actually wasn't supported. Our unit tests wrap the method in a class, so I added an integration test
* Looking at this again, filtering out stuff here aggressively means that invalid uses of `[GenerateSync]` are just ignored. We should have an analyzer to go along with this.

omsmith (approved):

PR #866, opened 2022-02-10
some immutability tweaks

omsmith:
Didn't measure anything terribly noticeable. There's variance and I'm not about to run it 5 times each :P

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
@@ -136,22 +136,22 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				SyntaxKind.LocalFunctionStatement
 			);
 
-			context.RegisterSyntaxNodeAction(
+			context.RegisterSymbolAction(
 				ctx => AnalyzeConflictingImmutabilityOnTypeParameters(
 					ctx,
+					(INamedTypeSymbol)ctx.Symbol,
 					annotationsContext
 				),
-				SyntaxKind.TypeParameter
+				SymbolKind.NamedType
 			);
 
-			context.RegisterSyntaxNodeAction(
+			context.RegisterSymbolAction(
 				ctx => AnalyzeConflictingImmutabilityOnMember(
 					ctx,
+					(INamedTypeSymbol)ctx.Symbol,
 					annotationsContext
 				),
-				SyntaxKind.ClassDeclaration,
-				SyntaxKind.InterfaceDeclaration,
-				SyntaxKind.StructDeclaration
+				SymbolKind.NamedType
 			);
 		}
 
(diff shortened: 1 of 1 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -136,22 +136,22 @@ CompilationStartAnalysisContext context
...
 				ctx => AnalyzeConflictingImmutabilityOnMember(
 					ctx,
+					(INamedTypeSymbol)ctx.Symbol,
 					annotationsContext
 				),
-				SyntaxKind.ClassDeclaration,
-				SyntaxKind.InterfaceDeclaration,
-				SyntaxKind.StructDeclaration
+				SymbolKind.NamedType
 			);
```
omsmith, line 155: There's a few NamedType symbol registrations, but leaving refactoring or combining for a later exercise

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -136,22 +136,22 @@ CompilationStartAnalysisContext context
...
+					(INamedTypeSymbol)ctx.Symbol,
 					annotationsContext
 				),
-				SyntaxKind.TypeParameter
+				SymbolKind.NamedType
 			);
 
-			context.RegisterSyntaxNodeAction(
+			context.RegisterSymbolAction(
 				ctx => AnalyzeConflictingImmutabilityOnMember(
```
omsmith, line 149: Still named wrong, but also later

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -136,22 +136,22 @@ CompilationStartAnalysisContext context
...
-			context.RegisterSyntaxNodeAction(
+			context.RegisterSymbolAction(
 				ctx => AnalyzeConflictingImmutabilityOnTypeParameters(
 					ctx,
+					(INamedTypeSymbol)ctx.Symbol,
 					annotationsContext
 				),
-				SyntaxKind.TypeParameter
+				SymbolKind.NamedType
 			);
```
omsmith, line 146: This was checking all type parameters before, whereas now it will only check type parameters on types.

We already have another check saying `[OnlyIf]` is not allowed a method type parameters, so we don't need to check method type parameters for `[Immutable]` + `[OnlyIf]`, since we don't care about that.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -325,60 +325,57 @@ CancellationToken cancellationToken
 		}
 
 		private static void AnalyzeConflictingImmutabilityOnTypeParameters(
```
omsmith, line 327: I think this one reads better now, and it being on only namedtype type parameters is good.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -325,60 +325,57 @@ CancellationToken cancellationToken
...
-
 			// Get information about immutability
 			bool hasImmutable = annotationsContext.Objects.Immutable.IsDefined( symbol );
 			bool hasConditionallyImmutable = annotationsContext.Objects.ConditionallyImmutable.IsDefined( symbol );
 			bool hasImmutableBase = annotationsContext.Objects.ImmutableBaseClass.IsDefined( symbol );
 
+			TypeDeclarationSyntax syntax = null;
+			TypeDeclarationSyntax GetSyntax() {
+				return syntax ??= (TypeDeclarationSyntax)symbol.DeclaringSyntaxReferences[ 0 ].GetSyntax( ctx.CancellationToken );
+			}
```
j3parker, line 366: nice

j3parker (dismissed):

src/D2L.CodeStyle.Analyzers/Immutability/ImmutabilityAnalyzer.cs
```diff
@@ -325,60 +325,57 @@ CancellationToken cancellationToken
...
+
 			// Check if there are conflicting immutability attributes
 			if( hasImmutable && hasConditionallyImmutable ) {
 				// [Immutable] and [ConditionallyImmutable] both exist,
 				// so create a diagnostic
 				ctx.ReportDiagnostic(
 					Diagnostics.ConflictingImmutability,
-					syntax.Identifier.GetLocation(),
+					GetSyntax().Identifier.GetLocation(),
```
omsmith, line 374: I checked afterward and `symbol.Locations[ 0 ]` is actually the identifier location, so grabbing the syntax isn't actually necessary.

`Locations[0]` and `DeclaringSyntaxReferences[0]` of course both have the problem of "it might not be the _right_ partial" with the attributes on it.

omsmith: Err actually, existing setup has that problem anyway and would create multiple diagnostics for partials.

PR #872, opened 2022-02-10
Add awaits and identifiers to SyncGenerator

j3parker:


src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -170,6 +179,9 @@ private ExpressionSyntax Transform( ExpressionSyntax expr )
 				.WithExpression( Transform( eaExpr.Expression ) )
 				.WithArgumentList( TransformAll( eaExpr.ArgumentList, Transform ) ),
 
+			IdentifierNameSyntax identExpr => identExpr
+				.WithIdentifier( RemoveAsyncSuffix( identExpr.Identifier, optional: true ) ),
```
j3parker, line 183: This is _really_ stupid, like we're making a big assumption here:

* If you ever talk about `FooAsync` you want to talk about `Foo` in sync
* Every async method will use the suffix.

Here's an example of what I'm going for:

```cs
var x = FooAsync();
var y = await BarAsync();
await x;
```

-->

```cs
var x = Foo();
var y = Bar();
```

Two things to note:

1. This changes the control flow. This example is intentionally bad, but it's something to keep in mind.
2.  Something that we don't do yet is detect that `await x;` -> `x;` isn't valid as an ExpressionStatementSyntax (its a valid Expression, but only some can become statements):

> CS0201: Only assignment, call, increment, decrement, and new object expressions can be used as a statement

In a future PR I'll delete these by replacing them with an EmptyStatement (just a `;` on its own)

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -154,6 +159,10 @@ private ExpressionSyntax Transform( ExpressionSyntax expr )
 				.WithLeft( Transform( asgnExpr.Left ) )
 				.WithRight( Transform( asgnExpr.Right ) ),
 
+			AwaitExpressionSyntax awaitExpr =>
+				// Extract the inner expression, removing the "await"
+				Transform( awaitExpr.Expression ).ConcatLeadingTriviaFrom( awaitExpr ),
```
j3parker, line 164: Here's an example of what I'm going for:

```cs
var x = FooAsync();
var y = await BarAsync();
await x;
```

-->

```cs
var x = Foo();
var y = Bar();
```

Two things to note:

1. This changes the control flow. This example is intentionally bad, but it's something to keep in mind.
2.  Something that we don't do yet is detect that `await x;` -> `x;` isn't valid as an ExpressionStatementSyntax (its a valid Expression, but only some can become statements):

> CS0201: Only assignment, call, increment, decrement, and new object expressions can be used as a statement

In a future PR I'll delete these by replacing them with an EmptyStatement (just a `;` on its own)

JeffAshton (approved):

mthjones (approved):

omsmith (approved):

PR #887, opened 2023-06-02
gha

omsmith:


.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
...
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+
+jobs:
+  all:
+    runs-on: windows-latest
```
omsmith, line 11: Errors about a Sqlite binary on linux. Assumedly fixed in newer versions, but will look after getting things going.

omsmith: It's also insanely slow to do the `dotnet build` on windows runners for whatever reason (2m26s vs 19s), so would be nice to switch

.github/workflows/build.yml
```diff
@@ -0,0 +1,32 @@
...
+        run: dotnet build -c Release
+
+      - name: Test
+        run: dotnet test -c Release --no-build
+
+      - name: Archive
+        uses: Brightspace/third-party-actions@actions/upload-artifact
+        with:
+          name: D2L.CodeStyle.${{ github.sha }}
+          path: src/**/*.symbols.nupkg
```
omsmith, line 32: Downloadable from the workflow summary https://github.com/Brightspace/D2L.CodeStyle/actions/runs/5158993147

Just maintaining the "upload to nuget manually" thing

Issue #736, opened 2021-03-29
Require D2LPage implementations to be partial


j3parker:
D2LPage (the "legacy framework") can use dependency injection, but only if the class is `partial`.
There is no harm to requiring `partial` every time, and this will remove a frequent source of confusion.

Put this one under the "ApiUsage" folder.

PR #888, opened 2023-06-12
Require making D2LPage derived classes partial

cesar-d2l:
Closes: https://github.com/Brightspace/D2L.CodeStyle/issues/736

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/D2LPageAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/D2LPageAnalyzer.cs
@@ -1,0 +1,56 @@
+﻿using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.Analyzers.ApiUsage {
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal class D2LPageAnalyzer : DiagnosticAnalyzer {
+		public override ImmutableArray<DiagnosticDescriptor> SupportedDiagnostics => ImmutableArray.Create(
+			Diagnostics.D2LPageDerivedMustBePartial
+		);
+
+		public override void Initialize( AnalysisContext context ) {
+			context.EnableConcurrentExecution();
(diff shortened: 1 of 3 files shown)


src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -748,5 +748,14 @@ public static class Diagnostics {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor D2LPageDerivedMustBePartial = new DiagnosticDescriptor(
+			id: "D2L0101",
+			title: "Classes derived from D2LPage must be partial",
+			messageFormat: "Partial page objects can support dependency injection rather than having to rely on locators",
+			category: "Correctness",
+			defaultSeverity: DiagnosticSeverity.Info,
```
omsmith, line 757: Given you can land the version bump at the same time as the necessary edits to fix existing diagnostics, I think this would be fine as error?

omsmith: Back with VS2017(I think?) when I could manage to open Lms.sln, for diagnostics that we made codefixes for I'd apply it to the whole sln with VS. But otherwise yeah, I do rewrites separately. https://github.com/Brightspace/omsmith-scratch/tree/roslyn-refactoring is the tooling I use, but I think y'all have also put stuff together for yourselves these days?

cesar-d2l: Think I have something working, so bringing this back to error severity

src/D2L.CodeStyle.Analyzers/ApiUsage/D2LPageAnalyzer.cs
```diff
@@ -0,0 +1,56 @@
...
+				if( baseType.Equals( d2lPage, SymbolEqualityComparer.Default ) ) {
+					inheritsD2LPage = true;
+					break;
+				}
+				baseType = baseType.BaseType;
+			}
+
+			if( inheritsD2LPage && !cds.Modifiers.Any( m => m.IsKind( SyntaxKind.PartialKeyword ) ) ) {
+				context.ReportDiagnostic( Diagnostics.D2LPageDerivedMustBePartial, symbol.Locations.First() );
+			}
```
omsmith, line 53: `if( cds.Modifiers.Any( m => m.IsKind( SyntaxKind.PartialKeyword ) ) { return; }` is probably a faster check that could be done before going to the semantic model.

omsmith (approved):

PR #891, opened 2023-07-05
Add some sync generators

AnaCoda:
Generators and tests for Variable Declaration, Member Access Expression, and Predefined Type

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -1,5 +1,4 @@
-﻿using System.Diagnostics;
-using D2L.CodeStyle.Analyzers.Extensions;
+﻿using D2L.CodeStyle.Analyzers.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
@@ -317,7 +316,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 	public T UnhandledSyntax<T>( T node ) where T : SyntaxNode {
 		GeneratorError(
 			node.GetLocation(),
-			$"unhandled syntax kind: {node.Kind()} and {node.GetType()}"
+			$"unhandled syntax kind: {node.Kind()}"
 		);
 
 		return node;
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -259,7 +259,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 		if (memberAccessExpr.IsKind(SyntaxKind.SimpleMemberAccessExpression )) {
 			return memberAccessExpr
 				.WithExpression( Transform( memberAccessExpr.Expression ) )
-				.WithExpression( Transform( memberAccessExpr.Name ) );
+				.WithName(  memberAccessExpr.Name );
 		}
 
 		return UnhandledSyntax( memberAccessExpr );
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -1,4 +1,5 @@
-﻿using D2L.CodeStyle.Analyzers.Extensions;
+﻿using System.Diagnostics;
+using D2L.CodeStyle.Analyzers.Extensions;
 using Microsoft.CodeAnalysis;
 using Microsoft.CodeAnalysis.CSharp;
 using Microsoft.CodeAnalysis.CSharp.Syntax;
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -107,12 +106,45 @@ AttributeSyntax attribute
...
+					GeneratorError(
+						typeSynt.GetLocation(),
+						$"Unexpected Task type: {returnTypeInfo.Type.MetadataName}"
+					);
+					return typeSynt;
+			}
+		}
+
+		return typeSynt;
+	}
```
j3parker, line 137: At a glance this looks the same as `TransformReturnType`?
If they are the same, is there a reason for having both? Could we just rename `TransformReturnType` to `TransformType` if it reads better?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -149,6 +181,19 @@ private StatementSyntax Transform( StatementSyntax stmt )
...
+			LocalDeclarationStatementSyntax localDeclStmt => localDeclStmt
+				.WithDeclaration( Transform( localDeclStmt.Declaration ) ),
+
+			TryStatementSyntax tryStmt => tryStmt
+				.WithBlock( Transform( tryStmt.Block ) )
+				.WithCatches( TransformAll( tryStmt.Catches, Transform ) ),
+
+			ForEachStatementSyntax forEachStmt => forEachStmt
+				.WithType( TransformType( forEachStmt.Type ) )
+				.WithIdentifier( RemoveAsyncSuffix( forEachStmt.Identifier, optional: true ) )
```
j3parker, line 193: Hm, is this like:

```cs
foreach( var fooAsync = ... ) ...
```

into

```cs
foreach( var foo = ... ) ...
```

If so I don't think we need to modify the identifier -- its really unlikely to have an async suffix, but also we're not really un-asyncing it per-se. And like consider that inside the body of the `foreach` unless we modify every identifier to remove the suffixes this would potentially introduce errors, like:

```cs
foreach( var fooAsync = ... ) {
  Bar( fooAsync );
}
```

-->

```cs
foreach( var foo = ... ) {
  Bar( fooAsync ); // error: fooAsync is undefined
}
```

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -65,6 +65,87 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+		var actual = Transform( @"[GenerateSync] async Task BarAsync() { var baz = await ( Task<int> )fooHandler.ReadFooAsync( ""foo"" ); }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Bar() { var baz = ( int )fooHandler.ReadFoo( \"foo\" ); }", actual.Value.ToFullString() );
+	}
+
+	[Test]
+	public void ForEach() {
+		var actual = Transform( @"[GenerateSync] async Task BarAsync() { foreach( Task<int> fooAsync in foosAsync ) { bar += await fooAsync; } }" );
```
j3parker, line 106: Mmmm I see where your transformation of the identifier is coming from (my comments above)... interesting.
Maybe we should consistently transform all identifiers. It maybe has downsides (more likely to have name collisions etc.) but it would be very predictable.

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -65,6 +65,87 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+			}", actual.Value.ToFullString() );
+	}
+
+	[Test]
+	public void DeclarationExpression() {
+		var actual = Transform( @"[GenerateSync] async Task BarAsync() { await TryBazAsync( input, out Task<int> result ); }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Bar() { TryBaz( input,out int result ); }", actual.Value.ToFullString() );
```
j3parker, line 146: Fantastic

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
j3parker: I left a comment below in the test case

j3parker (approved):

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
AnaCoda: I made TransformType so that ReportDiagnostic( Diagnostics.NonTaskReturnType, returnType.GetLocation() ); would not be reported for non-return types of a function. Should I combine them and add an optional param?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
AnaCoda: That's what I was thinking. It would be a bit confusing to remove some of the suffixes and not all of them

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
j3parker: Ah ya I see. That would make sense to me!

j3parker (approved):

PR #895, opened 2023-07-10
Delete CompletedTask, FromResult, and Fix Generator Cast Error

AnaCoda:


tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -120,6 +120,33 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+	}
+
+	[Test]
+	public void FromResult() {
+		var actual = Transform( @"[GenerateSync] async Task BarAsync() { return Task.FromResult( baz ); }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Bar() { return ( baz ); }", actual.Value.ToFullString() );
+	}
```
omsmith, line 139: I think the realistic input here would be `[GenerateSync] Task BarAsync() { return Task.FromResult( baz ); }`. As noted in your other PR, having the `async` method without the `await` is itself an error.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -264,20 +264,28 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
...
 			}
 		}
 		return invocationExpr
 			.WithExpression( Transform( invocationExpr.Expression ) )
 			.WithArgumentList( TransformAll( invocationExpr.ArgumentList, Transform ) );
 	}
 
-	private MemberAccessExpressionSyntax Transform( MemberAccessExpressionSyntax memberAccessExpr ) {
+	private ExpressionSyntax Transform( MemberAccessExpressionSyntax memberAccessExpr ) {
+		string[] identifiersToRemove = { "FromResult", "CompletedTask" };
```
j3parker, line 283: This does an allocation each call to `Transform`... another approach is to make a helper function that returns `bool` and does a `switch`, e.g.

```cs
bool MemberAccessesToRemove( MemeberAccessExpressionSyntax memberAccessExpr )
  => memberAccessExpr.Name.Identifier.ValueText switch {
    "FromResult" => true,
    "CompletedTask" => true,
    _ => false
  };
  ```
This general approach may also be more efficient than `.Contains` for a larger number of cases (currently the compiler turns a 7+-way switch into a code-gen'd hash table iirc). Not a signficant concern here of course.

File: tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs [hunk 3]
```diff
@@ -120,6 +120,33 @@ internal sealed class AsyncToSyncMethodTransformerTests {
 		Assert.AreEqual( "[Blocking] void Bar() { Foo fooFirst = m_foo.Bar( barrer ); }", actual.Value.ToFullString() );
 	}
 
+	[Test]
+	public void CompletedTaskConfigureAwait() {
+		var actual = Transform( @"[GenerateSync] async Task BarAsync() { Task.CompletedTask.ConfigureAwait(false); }" );
```
j3parker, line 125: I think this would have to be `await TaskCompletedTask...` to be valid input?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -120,6 +120,33 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Bar() { ; }", actual.Value.ToFullString() );
+	}
+
+	[Test]
+	public void FromResult() {
+		var actual = Transform( @"[GenerateSync] async Task BarAsync() { return Task.FromResult( baz ); }" );
```
j3parker, line 134: Hm, this one is `async Task` but is returning a `Task<T>`... Not sure exactly what you wanted to test here, but I think you wanted the return type to be `Task<T>` and not to have this be `async`?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
j3parker: I wonder, however, if we should be more conservative here -- rather than making this call inside a member access expression, should we be doing it in a `ReturnStatementSyntax` (or whatever its called)? Like `return Task.FromResult( ... )` --> `return ...`.

I wonder because there is code like this:

```cs
var result = Task.CompletedTask;
return result;
```

which is not likely but I think we'd get maybe a bit further trying to syncify this than we'd like?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -16,7 +16,7 @@ CancellationToken token
 		decl = decl.WithAttributeLists( ReplaceGenerateSyncAttribute( decl.AttributeLists ) )
 			.WithModifiers( RemoveAsyncModifier( decl.Modifiers ) )
 			.WithIdentifier( RemoveAsyncSuffix( decl.Identifier ) )
-			.WithReturnType( TransformType( decl.ReturnType, optional: false ) )
+			.WithReturnType( TransformType( decl.ReturnType, isNotReturnType: false ) )
```
j3parker, line 19: nit: it might be easier to read if you remove the negation, like `isReturnType` instead. Just consider it heh.

j3parker: Avoids stuff like `!isNotReturnType` and cases where you have an if/else can just be swapped easily etc.

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -120,6 +120,43 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+	}
+
+	// Not the best thing to be doing but unlikely this will happen and it should be easily caught
+	[Test]
+	public void DontRemoveUnreturnedFromResult() {
+		var actual = Transform( @"[GenerateSync] Task BarAsync() { var Baz = Task.CompletedTask; return Baz; }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Bar() { var Baz = Task.CompletedTask; return Baz; }", actual.Value.ToFullString() );
```
j3parker, line 157: This will be a compiler error but that's OK. It would be a bit confusing to diagnose as a user (why is the generator trying to return something here?), so an explicit error would be nicer... even if its opaque, like if we detect we are returning something from a void method then we can just bail and say "uh, can't generate this one sorry".

I'm just thinking out loud. Having a test to demonstrate the current behaviour + a comment is good 👍

j3parker (dismissed):

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
omsmith: I would follow this suggestion before merging.

AnaCoda: I was going to do this in the next PR (https://github.com/Brightspace/D2L.CodeStyle/pull/896), is that alright since I already made the changes there? Whoops

omsmith: I think it should be easy enough to bring the changes forward and resolve the differences in your follow-up. That's at least partly what git is for :smile:.

j3parker (dismissed):

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -264,20 +263,36 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
...
+	bool ReturnedMemberAccessesToRemove( MemberAccessExpressionSyntax memberAccessExpr )
+	  => memberAccessExpr.Name.Identifier.ValueText switch {
+		  "FromResult" => true,
+		  "CompletedTask" => true,
+		  _ => false
+	  };
+
+	private ExpressionSyntax Transform( MemberAccessExpressionSyntax memberAccessExpr ) {
 		if( memberAccessExpr.IsKind( SyntaxKind.SimpleMemberAccessExpression ) ) {
+			if( ReturnedMemberAccessesToRemove( memberAccessExpr ) &&
```
omsmith, line 290:
```suggestion
	bool ShouldRemoveReturnedMemberAccess( MemberAccessExpressionSyntax memberAccessExpr )
	  => memberAccessExpr.Name.Identifier.ValueText switch {
		  "FromResult" => true,
		  "CompletedTask" => true,
		  _ => false
	  };

	private ExpressionSyntax Transform( MemberAccessExpressionSyntax memberAccessExpr ) {
		if( memberAccessExpr.IsKind( SyntaxKind.SimpleMemberAccessExpression ) ) {
			if( ShouldRemoveReturnedMemberAccess( memberAccessExpr ) &&
```

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -264,20 +263,36 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
...
 			.WithArgumentList( TransformAll( invocationExpr.ArgumentList, Transform ) );
 	}
 
-	private MemberAccessExpressionSyntax Transform( MemberAccessExpressionSyntax memberAccessExpr ) {
+	bool ReturnedMemberAccessesToRemove( MemberAccessExpressionSyntax memberAccessExpr )
+	  => memberAccessExpr.Name.Identifier.ValueText switch {
+		  "FromResult" => true,
+		  "CompletedTask" => true,
+		  _ => false
+	  };
```
omsmith, line 286: I think this would ideally check that `memberAccessExpr.Expression` is actually `Task`

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -264,20 +263,36 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
...
+
+	private ExpressionSyntax Transform( MemberAccessExpressionSyntax memberAccessExpr ) {
 		if( memberAccessExpr.IsKind( SyntaxKind.SimpleMemberAccessExpression ) ) {
+			if( ShouldRemoveReturnedMemberAccess( memberAccessExpr ) &&
+				( memberAccessExpr.Parent.IsKind( SyntaxKind.ReturnStatement ) ||
+				( memberAccessExpr.Parent?.Parent?.IsKind( SyntaxKind.ReturnStatement ) ?? false ) ) ) {
```
omsmith, line 292: Fine for now, but I think we could probably pass around that we're in a return statement on the stack into MaybeTransform/Transform from the `ReturnStatementSyntax` arm in `private StatementSyntax Transform( StatementSyntax stmt )` rather than inspecting upwards.

omsmith (approved):

PR #898, opened 2023-07-13
Modify returns if generated sync method is void

AnaCoda:
[US154650](https://rally1.rallydev.com/#/?detail=/userstory/706189919605&fdp=true): Async to Sync Generator: Generate RequestAuthenticator

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -192,7 +192,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 	private StatementSyntax Transform( ReturnStatementSyntax returnStmt ) {
 		var expr = returnStmt.Expression;
 		if( m_generatedFunctionReturnsVoid && expr is not null ) {
-			if( expr is InvocationExpressionSyntax || expr is AssignmentExpressionSyntax || expr is IIncrementOrDecrementOperation || expr is DeclarationExpressionSyntax || expr is MemberAccessExpressionSyntax ) {
+			if( IsStatementCompatibleExpression( expr ) ) {
 				return SyntaxFactory.ParseStatement( $"{Transform(expr).WithTriviaFrom(returnStmt)};\nreturn;" );
 			} else {
 				return SyntaxFactory.ParseStatement( "return;" );
@@ -200,6 +200,16 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 		}
 		return returnStmt.WithExpression( MaybeTransform( returnStmt.Expression, Transform ) );
 	}
+
+	private static bool IsStatementCompatibleExpression( ExpressionSyntax expr )
+		=> expr switch {
+			InvocationExpressionSyntax e => true,
+			AssignmentExpressionSyntax e => true,
+			IIncrementOrDecrementOperation e => true,
+			DeclarationExpressionSyntax e => true,
+			MemberAccessExpressionSyntax e => true,
+			_ => false
+		};
 
 	private ExpressionSyntax Transform( ExpressionSyntax expr )
 		=> expr switch {
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -198,7 +198,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 						SyntaxFactory.ReturnStatement().WithLeadingTrivia( SyntaxFactory.Space ).WithTrailingTrivia( SyntaxFactory.Space ) )
 					.WithTriviaFrom( returnStmt );
 			} else {
-				return SyntaxFactory.ParseStatement( "return;" );
+				return SyntaxFactory.ReturnStatement().WithTriviaFrom( returnStmt );
 			}
 		}
 		return returnStmt.WithExpression( MaybeTransform( returnStmt.Expression, Transform ) );
@@ -208,9 +208,9 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 		=> expr switch {
 			InvocationExpressionSyntax e => true,
 			AssignmentExpressionSyntax e => true,
-			IIncrementOrDecrementOperation e => true,
-			DeclarationExpressionSyntax e => true,
-			MemberAccessExpressionSyntax e => true,
+			PostfixUnaryExpressionSyntax e => true,
+			PrefixUnaryExpressionSyntax e => true,
+			ObjectCreationExpressionSyntax => true,
 			_ => false
 		};
 
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -287,7 +287,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 				newExpr = memberAccess.Expression;
 				return Transform( newExpr );
 			}
-		} else if( memberAccess is not null && ShouldWrapMemberAccessesInTaskRun( memberAccess ) ) {
+		} else if( memberAccess is not null && ShouldWrapMemberAccessInTaskRun( memberAccess ) ) {
 			m_disableTaskRunWarningFlag = true;
 			return SyntaxFactory.ParseExpression( $"Task.Run(() => {invocationExpr}).Result" );
 		}
@@ -297,7 +297,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 			.WithArgumentList( TransformAll( invocationExpr.ArgumentList, Transform ) );
 	}
 
-	bool ShouldRemoveReturnedMemberAccesses( MemberAccessExpressionSyntax memberAccessExpr ) {
+	bool ShouldRemoveReturnedMemberAccess( MemberAccessExpressionSyntax memberAccessExpr ) {
 		return ( memberAccessExpr.Expression.ToString(), memberAccessExpr.Name.Identifier.ValueText ) switch
 		{
 			( "Task", "FromResult" ) => true,
@@ -306,7 +306,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 		};
 	}
 
-	bool ShouldWrapMemberAccessesInTaskRun( MemberAccessExpressionSyntax memberAccessExpr ) {
+	bool ShouldWrapMemberAccessInTaskRun( MemberAccessExpressionSyntax memberAccessExpr ) {
 		return (memberAccessExpr.Expression.ToString(), memberAccessExpr.Name.Identifier.ValueText) switch {
 			(_, "ReadAsStringAsync" ) => true,
 			_ => false
@@ -315,7 +315,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 
 	private ExpressionSyntax Transform( MemberAccessExpressionSyntax memberAccessExpr ) {
 		if( memberAccessExpr.IsKind( SyntaxKind.SimpleMemberAccessExpression ) ) {
-			if( ShouldRemoveReturnedMemberAccesses( memberAccessExpr ) &&
+			if( ShouldRemoveReturnedMemberAccess( memberAccessExpr ) &&
 				( memberAccessExpr.Parent.IsKind( SyntaxKind.ReturnStatement ) ||
 				( memberAccessExpr.Parent?.Parent?.IsKind( SyntaxKind.ReturnStatement ) ?? false ) ) ) {
 				return SyntaxFactory.ParseExpression( "" );
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -195,7 +195,8 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 			if( IsStatementCompatibleExpression( expr ) ) {
 				return SyntaxFactory.Block(
 						SyntaxFactory.ExpressionStatement( Transform( expr ) ).WithLeadingTrivia( SyntaxFactory.Space ),
-						SyntaxFactory.ReturnStatement().WithLeadingTrivia( SyntaxFactory.Space ).WithTrailingTrivia( SyntaxFactory.Space ) )
+						SyntaxFactory.ReturnStatement().WithLeadingTrivia( SyntaxFactory.Space ).WithTrailingTrivia( SyntaxFactory.Space )
+					)
 					.WithTriviaFrom( returnStmt );
 			} else {
 				return SyntaxFactory.ReturnStatement().WithTriviaFrom( returnStmt );
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -11,6 +11,9 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 		CancellationToken token
 	) : base( model, token ) { }
 
+	// Need to disable D2L0018 in the method if we add Task.Run() to syncify something
+	private bool disableTaskRunWarningFlag;
+
 	public TransformResult<MethodDeclarationSyntax> Transform( MethodDeclarationSyntax decl ) {
 		// TODO: remove CancellationToken parameters
 		decl = decl.WithAttributeLists( ReplaceGenerateSyncAttribute( decl.AttributeLists ) )
@@ -19,6 +22,17 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 			.WithReturnType( TransformType( decl.ReturnType, isNotReturnType: false ) )
 			.WithExpressionBody( MaybeTransform( decl.ExpressionBody, Transform ) )
 			.WithBody( MaybeTransform( decl.Body, Transform ) );
+
+		if ( disableTaskRunWarningFlag ) {
+			PragmaWarningDirectiveTriviaSyntax restorePragma = SyntaxFactory.PragmaWarningDirectiveTrivia( SyntaxFactory.Token( SyntaxKind.RestoreKeyword ), true )
+				.AddErrorCodes( SyntaxFactory.IdentifierName( "D2L0018" ) ).NormalizeWhitespace().WithLeadingTrivia( SyntaxFactory.SyntaxTrivia( SyntaxKind.EndOfLineTrivia, "\n" ) ); ;
+			PragmaWarningDirectiveTriviaSyntax disablePragma = SyntaxFactory.PragmaWarningDirectiveTrivia( SyntaxFactory.Token( SyntaxKind.DisableKeyword ), true )
+				.AddErrorCodes( SyntaxFactory.IdentifierName( "D2L0018" ) ).NormalizeWhitespace();
+			decl = decl
+				.WithLeadingTrivia( decl.GetLeadingTrivia().Add( SyntaxFactory.Trivia( disablePragma ) ) )
+				.WithTrailingTrivia( decl.GetTrailingTrivia().Insert( 0, SyntaxFactory.Trivia( restorePragma ) ) );
+		}
+
 		return GetResult( decl );
 	}
 
@@ -273,6 +287,11 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 				return Transform( newExpr );
 			}
 		}
+		else if( memberAccess is not null && MemberAccessesToWrapInTaskRun( memberAccess ) ) {
+			disableTaskRunWarningFlag = true;
+			return SyntaxFactory.ParseExpression( $"Task.Run(() => {invocationExpr}).Result" );
+		}
+
 		return invocationExpr
 			.WithExpression( Transform( invocationExpr.Expression ) )
 			.WithArgumentList( TransformAll( invocationExpr.ArgumentList, Transform ) );
@@ -282,6 +301,12 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 	  => memberAccessExpr.Name.Identifier.ValueText switch {
 		  "FromResult" => true,
 		  "CompletedTask" => true,
+		  _ => false
+	  };
+
+	bool MemberAccessesToWrapInTaskRun( MemberAccessExpressionSyntax memberAccessExpr )
+	  => memberAccessExpr.Name.Identifier.ValueText switch {
+		  "ReadAsStringAsync" => true,
 		  _ => false
 	  };
 
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -31,6 +31,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 			decl = decl
 				.WithLeadingTrivia( decl.GetLeadingTrivia().Add( SyntaxFactory.Trivia( disablePragma ) ) )
 				.WithTrailingTrivia( decl.GetTrailingTrivia().Insert( 0, SyntaxFactory.Trivia( restorePragma ) ) );
+			disableTaskRunWarningFlag = false;
 		}
 
 		return GetResult( decl );
@@ -297,6 +298,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 			.WithArgumentList( TransformAll( invocationExpr.ArgumentList, Transform ) );
 	}
 
+	// TODO: These two methods may need future modification for more specificity (make sure it's Task.FromResult or Content.ReadAsStringAsync)
 	bool ReturnedMemberAccessesToRemove( MemberAccessExpressionSyntax memberAccessExpr )
 	  => memberAccessExpr.Name.Identifier.ValueText switch {
 		  "FromResult" => true,
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -198,14 +198,14 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 		if( !m_generatedFunctionReturnsVoid || expr is null ) {
 			return returnStmt.WithExpression( MaybeTransform( returnStmt.Expression, Transform ) );
 		}
-		
+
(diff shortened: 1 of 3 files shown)


src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -185,6 +189,18 @@ private StatementSyntax Transform( StatementSyntax stmt )
 			_ => UnhandledSyntax( stmt )
 		};
 
+	private StatementSyntax Transform( ReturnStatementSyntax returnStmt ) {
+		var expr = returnStmt.Expression;
+		if( m_generatedFunctionReturnsVoid && expr is not null ) {
+			if( expr is InvocationExpressionSyntax || expr is AssignmentExpressionSyntax || expr is IIncrementOrDecrementOperation || expr is DeclarationExpressionSyntax || expr is MemberAccessExpressionSyntax ) {
+				return SyntaxFactory.ParseStatement( $"{Transform(expr).WithTriviaFrom(returnStmt)};\nreturn;" );
```
j3parker, line 196:
```suggestion
				return SyntaxFactory.ParseStatement( $"{Transform(expr).WithTriviaFrom(returnStmt)};" );

```

Do we need the `return;`? I wouldn't go out of my way to remove redundant things, but I also wouldn't go out of my way to add them.

j3parker: I think this could be `SyntaxFactory.ExpressionStatement( Transform( expr ) )` instead of calling into `ParseStatement` in that case?

ExpressionStatement takes an expression as input and "lifts" it up to be a statement. E.g. `Foo()` is an InvocationExpression, and `Foo();` is an ExpressionStatement contain that invocation (note also the semicolon)

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -185,6 +189,18 @@ private StatementSyntax Transform( StatementSyntax stmt )
 			_ => UnhandledSyntax( stmt )
 		};
 
+	private StatementSyntax Transform( ReturnStatementSyntax returnStmt ) {
+		var expr = returnStmt.Expression;
+		if( m_generatedFunctionReturnsVoid && expr is not null ) {
+			if( expr is InvocationExpressionSyntax || expr is AssignmentExpressionSyntax || expr is IIncrementOrDecrementOperation || expr is DeclarationExpressionSyntax || expr is MemberAccessExpressionSyntax ) {
```
j3parker, line 195: Nice -- this could maybe be a helper, and you could use `.RawKind` to do a switch (better perf than reflection/the `||`s (maybe on that second bit)):

```cs
private static bool IsStatementCompatibleExpression( ExpressionSyntax expr ) {
  switch( expr.RawKind ) {
    case SyntaxKind.InvocationExpression:
    case SyntaxKind.AssignmentExpression:
    // ...
      return true;
    default:
      return false;
  }
}
```

something like that

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
AnaCoda: I believe we need we keep the return afterwards for cases like this:
```
Task FooAsync() {
	if( condition ) {
		return BarAsync(2);
	}
	return BarAsync(4);
}
```
We don't want to transform this into 
```
Task FooAsync() {
	if( condition ) {
		Bar(2);
	}
	Bar(4);
}
```

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
AnaCoda: I would rather not use Kind because there is no SyntaxKind.AssignmentExpression, I would have to add cases for all of these:
image
Could I do something like this?
```
private static bool IsStatementCompatibleExpression( ExpressionSyntax expr ) {
  switch( expr) {
    case InvocationExpressionSyntax e:
    case AssignmentExpressionSyntax e:
    // ...
      return true;
    default:
      return false;
  }
}
```

j3parker: Ah ok. A helper would still be good IMO just to keep things tidy but you don't have to do it in this PR.

j3parker (dismissed):

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
omsmith, line 196: Given we're just inserting a newline here and not applying other trivia, won't this end up like

```csharp
    return BarAsync();
  }
```
-->
```csharp
    Bar();
return;}
```

j3parker: I'm also kind of wondering what the output of ParseStatement is for this, like the type? Because I see two statements here?

We could do `SyntaxFactory.Block( SyntaxFactory.ExpressionStatement( Transform( expr ) ), SyntaxFactory.Return() );` (ignoring whitespace anyway). I worry that ParseStatement is coming up with a crazy answer that happens to just serialize back to string correctly here.

j3parker: When I put `return; return;` into roslyn-quoter and ask it to parse it as a statement I get back out:

```
ReturnStatement()
.WithSemicolonToken(
    Token(
        TriviaList(),
        SyntaxKind.SemicolonToken,
        TriviaList(
            new []{
                Trivia(
                    SkippedTokensTrivia()
                    .WithTokens(
                        TokenList(
                            Token(SyntaxKind.ReturnKeyword)))),
                Trivia(
                    SkippedTokensTrivia()
                    .WithTokens(
                        TokenList(
                            Token(SyntaxKind.SemicolonToken))))})))
.NormalizeWhitespace()
```

Like the extra return gets jammed into the trivia of the first one as a skipped token. This wouldn't compile but if you `.ToString()` it (like we do in this generator) it will round trip...

AnaCoda: I see... sorry I didn't notice this!
```
Block(ExpressionStatement(
                        InvocationExpression(
                            IdentifierName("Baz"))),
         ReturnStatement())))))
```
Definitely seems to be the correct way. I'll try to do that and maybe in the process fix the whitespace

AnaCoda: Looks like SyntaxFactory.Block() is kind of the only way. Normally we might worry about it causing scope issues because of the extra {} but since we're returning it shouldn't usually matter. It doesn't look the greatest but it's the only way with our generation model and apparently common in code generation

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -193,7 +193,7 @@ private StatementSyntax Transform( StatementSyntax stmt )
 		var expr = returnStmt.Expression;
 		if( m_generatedFunctionReturnsVoid && expr is not null ) {
 			if( IsStatementCompatibleExpression( expr ) ) {
-				return SyntaxFactory.ParseStatement( $"{Transform(expr).WithTriviaFrom(returnStmt)};\nreturn;" );
+				return SyntaxFactory.Block( SyntaxFactory.ExpressionStatement( Transform( expr ) ), SyntaxFactory.ReturnStatement() ).NormalizeWhitespace();
```
omsmith, line 196: `NormalizeWhitespace` iiis, not performant. Probably fine on small statements, but from previous experience (in static-di) it can really tank. I wouldn't want to use it here. It also won't end up being consistent with the surrounding code anyway.

I don't think we necessarily need to get the whitespace / style _perfect_, but just want to be thoughtful / intentional.

AnaCoda: Removed NormalizeWhitespace and replace with new formatting that looks like this:
```
[Blocking]
void IBar.Delete( Guid id ) {
	{ m_baz.Delete( id ); return; }
}
```

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -154,16 +154,16 @@ internal sealed class AsyncToSyncMethodTransformerTests {
 
 		Assert.IsTrue( actual.Success );
 		Assert.IsEmpty( actual.Diagnostics );
-		Assert.AreEqual( "[Blocking] void Bar() { var Baz = Task.CompletedTask; return Baz; }", actual.Value.ToFullString() );
+		Assert.AreEqual( "[Blocking] void Bar() { var Baz = Task.CompletedTask; return;}", actual.Value.ToFullString() );
```
j3parker, line 157: Haha this worked out nice!

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -220,6 +220,33 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+	}
+
+	[Test]
+	public void TaskToVoidReturnTypeDoubleInvocation() {
+		var actual = Transform( @"[GenerateSync] Task FooAsync() { if( QuuxAsync() ) { return BarAsync(2); } return BarAsync(4); }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Foo() { if( Quux() ) { { Bar(2); return; } } { Bar(4); return; } }", actual.Value.ToFullString() );
+	}
```
j3parker, line 248: Could I bug you to add cases for the other expressions we allow that make sense?
`return foo++;`, `return foo.bar;`, `return x++` etc.?

j3parker (dismissed):

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
j3parker: You can use sharplab to double check if things like `return foo++;` are valid syntax

j3parker (dismissed):

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -185,6 +189,31 @@ private StatementSyntax Transform( StatementSyntax stmt )
 			_ => UnhandledSyntax( stmt )
 		};
 
+	private StatementSyntax Transform( ReturnStatementSyntax returnStmt ) {
+		var expr = returnStmt.Expression;
+		if( m_generatedFunctionReturnsVoid && expr is not null ) {
+			if( IsStatementCompatibleExpression( expr ) ) {
+				return SyntaxFactory.Block(
+						SyntaxFactory.ExpressionStatement( Transform( expr ) ).WithLeadingTrivia( SyntaxFactory.Space ),
+						SyntaxFactory.ReturnStatement().WithLeadingTrivia( SyntaxFactory.Space ).WithTrailingTrivia( SyntaxFactory.Space ) )
```
omsmith, line 198:
```suggestion
						SyntaxFactory.ReturnStatement().WithLeadingTrivia( SyntaxFactory.Space ).WithTrailingTrivia( SyntaxFactory.Space )
					)
```

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -101,6 +105,7 @@ AttributeSyntax attribute
 		if( returnTypeInfo.Type.ContainingNamespace.ToString() == "System.Threading.Tasks" ) {
 			switch( returnTypeInfo.Type.MetadataName ) {
 				case "Task":
+					if( isReturnType ) { m_generatedFunctionReturnsVoid = true; }
 					return isReturnType ? SyntaxFactory.ParseTypeName( "void" ).WithTriviaFrom( typeSynt ) : typeSynt;
```
j3parker, line 109:
```suggestion
					if( isReturnType ) {
						m_generatedFunctionReturnsVoid = true;
						return SyntaxFactory.ParseTypeName( "void" ).WithTriviaFrom( typeSynt )
					} else {
						return typeSynt;
					}
```

this would be simpler?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -185,6 +189,32 @@ private StatementSyntax Transform( StatementSyntax stmt )
...
+				return SyntaxFactory.Block(
+						SyntaxFactory.ExpressionStatement( Transform( expr ) ).WithLeadingTrivia( SyntaxFactory.Space ),
+						SyntaxFactory.ReturnStatement().WithLeadingTrivia( SyntaxFactory.Space ).WithTrailingTrivia( SyntaxFactory.Space )
+					)
+					.WithTriviaFrom( returnStmt );
+			} else {
+				return SyntaxFactory.ReturnStatement().WithTriviaFrom( returnStmt );
+			}
+		}
+		return returnStmt.WithExpression( MaybeTransform( returnStmt.Expression, Transform ) );
```
j3parker, line 209: An alternative way to write this is:

```suggestion
		if( !m_generatedFunctionReturnsVoid || expr is null ) {
			return returnStmt.WithExpression( MaybeTransform( returnStmt.Expression, Transform ) );
		}
		
		if( IsStatementCompatibleExpression( expr ) ) {
			return SyntaxFactory.Block(
					SyntaxFactory.ExpressionStatement( Transform( expr ) ).WithLeadingTrivia( SyntaxFactory.Space ),
					SyntaxFactory.ReturnStatement().WithLeadingTrivia( SyntaxFactory.Space ).WithTrailingTrivia( SyntaxFactory.Space )
			).WithTriviaFrom( returnStmt );
		}
		
		return SyntaxFactory.ReturnStatement().WithTriviaFrom( returnStmt );
```

i.e. by flipping the condition in the first `if` and returning we can take the other case and rather than having an else just flatten one level of indentation. And then that other final else also becomes optional because the inside of the `if` always returns.

The benefit of this style is (1) less indentation overall, which can make it easier to read (2) smaller stack trace in your head that you have to unwind, like as you nest ifs you need to push/pop assumptions about the variables you're checking in the condition, whereas in the flattened one you are only ever accumulating facts (not popping) as you go along, like the context becomes more and more specific the longer the function goes which is in a way simpler (imagine playing "20 questions", after a bunch of questions as long as you can look at the previous questions you've asked (look at the preceeding if statements), the fewer possible worlds you are in as you've asked more questions.)

Just a style thing though/matter of taste

j3parker: If you google "return early" you can get an infinite amount of opinions on this

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -220,6 +220,78 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+		Assert.AreEqual( "[Blocking] void Foo() { { --bar; return; } }", actual.Value.ToFullString() );
+	}
+
+	[Test]
+	public void TaskToVoidReturnTypeMemberAccess() {
+		var actual = Transform( @"[GenerateSync] Task FooAsync() { return Baz.bar; }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Foo() { return; }", actual.Value.ToFullString() );
```
j3parker, line 274: 👍 I mentioned in DM that technically property getters could do evil here, and we'd be changing the semantics in the generated code silently... but I think that's OK, it would be insane to do that.

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -220,6 +220,78 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+	}
+
+	[Test]
+	public void TaskToVoidReturnTypeAssignment() {
+		var actual = Transform( @"[GenerateSync] Task FooAsync() { return m_baz = QuuxAsync(); }" );
```
j3parker, line 279: I was going to say that I didn't think this was valid syntax in C# (it is in C/C++) but uh.... [it is!](https://sharplab.io/#v2:EYLgtghglgdgNAFxAJwK4wD4AEBMBGAWAChis8BOACgAIAxAe3soEprmBuY42BOxl6gG9i1UdQBuEZNQAe1ALzUcnImOpYA7NUpzFAZg7EAvkA==) lol.

j3parker (dismissed):

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -220,6 +220,78 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+		Assert.AreEqual( "[Blocking] void Foo() { { m_baz = Quux(); return; } }", actual.Value.ToFullString() );
+	}
+
+	[Test]
+	public void TaskToVoidReturnTypeNewObject() {
+		var actual = Transform( @"[GenerateSync] Task FooAsync() { return new Task(QuuxAsync); }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Foo() { { new Task(Quux); return; } }", actual.Value.ToFullString() );
```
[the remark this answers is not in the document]
j3parker: Doesn't have to be in this PR, could be a follow-up

j3parker (dismissed):

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -121,21 +121,21 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
 
 	[Test]
 	public void FromResult() {
-		var actual = Transform( @"[GenerateSync] Task BarAsync() { return Task.FromResult( baz ); }" );
+		var actual = Transform( @"[GenerateSync] Task<Baz> BarAsync() { return Task.FromResult( baz ); }" );
 
 		Assert.IsTrue( actual.Success );
 		Assert.IsEmpty( actual.Diagnostics );
-		Assert.AreEqual( "[Blocking] void Bar() { return ( baz ); }", actual.Value.ToFullString() );
+		Assert.AreEqual( "[Blocking] Baz Bar() { return ( baz ); }", actual.Value.ToFullString() );
```
omsmith, line 138: I know this is existing, but was leaving the parentheses intentional?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -220,6 +220,80 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+		var actual = Transform( @"[GenerateSync] Task FooAsync() { return null; }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Foo() { return; }", actual.Value.ToFullString() );
+	}
+
+	[Test]
+	public void TaskToVoidReturnTypeDoubleInvocation() {
+		var actual = Transform( @"[GenerateSync] Task FooAsync() { if( QuuxAsync() ) { return BarAsync(2); } return BarAsync(4); }" );
```
omsmith, line 243: nit: `if( QuuxAsync() )` doesn't really make sense.

```suggestion
		var actual = Transform( @"[GenerateSync] async Task FooAsync() { if( await QuuxAsync() ) { return await BarAsync(2); } return await BarAsync(4); }" );
```

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -220,6 +220,80 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+		var actual = Transform( @"[GenerateSync] Task FooAsync() { return Baz.bar; }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Foo() { return; }", actual.Value.ToFullString() );
+	}
+
+	[Test]
+	public void TaskToVoidReturnTypeAssignment() {
+		var actual = Transform( @"[GenerateSync] Task FooAsync() { return m_baz = QuuxAsync(); }" );
```
omsmith, line 279: `m_baz = QuuxAsync();` implies that `m_baz` is of type `Task`, which means `m_baz = Quux()` in the transformation doesn't make sense.

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
AnaCoda: Not exactly intentional but seems like a small price to pay to keep 
`Transform( MemberAccessExpressionSyntax memberAccessExpr )`
relatively clean

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
AnaCoda: Maybe `AssignmentExpressionSyntax` should be removed from `IsStatementCompatibleExpression`, because a Task returning function that returns an AssignmentExpression would always have to assign Task to some variable (which as you said, doesn't really make sense when we're transforming to sync). Thoughts?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -220,6 +220,80 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+		var actual = Transform( @"[GenerateSync] Task FooAsync() { return null; }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Foo() { return; }", actual.Value.ToFullString() );
+	}
+
+	[Test]
+	public void TaskToVoidReturnTypeDoubleInvocation() {
+		var actual = Transform( @"[GenerateSync] async Task FooAsync() { if( await QuuxAsync() ) { return await BarAsync(2); } return await BarAsync(4); }" );
```
omsmith, line 243: Err sorry, that was a bad suggestion as well :facepalm:. This is what actually makes sense, but it doesn't really show what you were trying to test now :smile:. I'll comment with another suggestion which maintains what you were trying to go for (which is just, don't have the `if` predicate be Async, since it's getting in the way).

```suggestion
		var actual = Transform( @"[GenerateSync] async Task FooAsync() { if( await QuuxAsync() ) { await BarAsync(2); return; } await BarAsync(4); }" );
```

omsmith, line 243:
```suggestion
		var actual = Transform( @"[GenerateSync] Task FooAsync() { if( Quux() ) { return BarAsync(2); } return BarAsync(4); }" );
```

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
j3parker: I think it is OK to be conservative and not support things in the generator that are super niche.

omsmith: Which also means bailing out entirely, not just eliding the statement/expression, right?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
j3parker: What's happening is that we are transforming `Task.FromResult` into `Parse( "" )` (which I think turns into an `IdentifierNameSyntax` with "missing token" placeholder) and the `(` and `)` are from the left over `ArgumentListSyntax`.

I think what we want to do is transform `InvocationExpression` and peek down into the `.Expression` to confirm that it is exactly `Task.FromResult` (if you do things funky like `var x = Task.FromResult` then its on you), and if so reach down into the `.Arguments[0]` and lift that up.

This has come up a couple of times now, I think we ought to have the tests validate that the generated parse tree is valid, or that ToString/Parse round trip in some reasonable way (like that we get an equivalent SyntaxTree from `Parse( output.ToString() )`.

It's not the end of the world that this is happening, but the (false) belief that we are generating valid parse trees gives me some peace of mind lol

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -185,6 +193,44 @@ private StatementSyntax Transform( StatementSyntax stmt )
...
+		if( IsStatementCompatibleExpression( expr ) == Compatibility.Compatible ) {
+			return SyntaxFactory.Block(
+					SyntaxFactory.ExpressionStatement( Transform( expr ) ).WithLeadingTrivia( SyntaxFactory.Space ),
+					SyntaxFactory.ReturnStatement().WithLeadingTrivia( SyntaxFactory.Space ).WithTrailingTrivia( SyntaxFactory.Space )
+			).WithTriviaFrom( returnStmt );
+		} else if (IsStatementCompatibleExpression( expr ) == Compatibility.Incompatible ) {
+			return SyntaxFactory.ReturnStatement().WithTriviaFrom( returnStmt );
+		}
+
+		return UnhandledSyntax( returnStmt );
```
j3parker, line 211:
```suggestion
		return IsStatementCompatibleExpression( expr ) switch {
			Compatibility.Compatible => SyntaxFactory.Block(
					SyntaxFactory.ExpressionStatement( Transform( expr ) ).WithLeadingTrivia( SyntaxFactory.Space ),
					SyntaxFactory.ReturnStatement().WithLeadingTrivia( SyntaxFactory.Space ).WithTrailingTrivia( SyntaxFactory.Space )
			).WithTriviaFrom( returnStmt ),
			Compatibility.Incompatible => SyntaxFactory.ReturnStatement().WithTriviaFrom( returnStmt ),
			Compatibility.Unsupported => UnhandledSyntax( returnStmt ),
			_ => throw new NotImplementedException()
		};
```

(switch makes it easier to see how we handle each case. The `_` case should never get hit in practice but enums have some weird sharp corners so crashing here is reasonable.)

j3parker (dismissed):

j3parker (approved):

PR #900, opened 2023-07-26
Add generator support and tests for GenericNameSyntax to sync

AnaCoda:


src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -392,6 +396,14 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
 		);
 	}
 
+	private SeparatedSyntaxList<TypeSyntax> TransformTypes( SeparatedSyntaxList<TypeSyntax> typeDecls ) {
+		return SyntaxFactory.SeparatedList(
+			typeDecls.Select( typeDecl =>
+				TransformType( typeDecl )
+			)
```
j3parker, line 403: nit, I think this is sufficient?

```suggestion
			typeDecls.Select( TransformType )
```

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -404,6 +416,9 @@ private ElseClauseSyntax Transform( ElseClauseSyntax clause )
 	private ArgumentListSyntax Transform( ArgumentListSyntax argList )
 		=> TransformAll( argList, Transform );
 
+	private TypeArgumentListSyntax Transform ( TypeArgumentListSyntax typeArgList )
+		=> SyntaxFactory.TypeArgumentList( TransformTypes( typeArgList.Arguments ) );
```
j3parker, line 420: does `typeArgList` have a `.WithArguments` or something? Just wondering because those `.WithX` buddies preserve syntax trivia (vs creating a new thing with `SyntaxFactory.TypeArgumentList`). There's `.WithTriviaFrom` as well I think for cases where you can't get by with a `.WithX` (either because it doesn't exist or you are changing the node type)

j3parker (dismissed):

j3parker (approved):

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
j3parker: (To be clear these are usually equivalent, just less verbose.)

PR #903, opened 2023-07-28
Support SyncGen of Parenthesized Anonymous Object Creation Lambda

AnaCoda:


tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -319,6 +319,15 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
 	}
 
+	[Test]
+	public void ParenthesizedAnonymousCreationLambda() {
+		var actual = Transform( @"[GenerateSync] Task BarAsync() { () => new { BazAsync = TaskCreator.NewTask, Quux = id.ToStringAsync( ""N"" ) } }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( @"[Blocking] void Bar() { () => new { Baz = TaskCreator.NewTask,Quux = id.ToString( ""N"" ) } }", actual.Value.ToFullString() );
+	}
```
omsmith, line 329: I don't believe `BarAsync` is currently a valid example? It's not returning a task anywhere.

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -319,6 +319,15 @@ internal sealed class AsyncToSyncMethodTransformerTests {
 		Assert.AreEqual( @"[Blocking] Foo Bar() { var ct = new CancellationToken(); return Foo( 1,""test"" ); }", actual.Value.ToFullString() );
 	}
 
+	[Test]
+	public void ParenthesizedAnonymousCreationLambda() {
```
omsmith, line 323: Are parenthesized lambdas and anonymous objects creations actually directly related that we're testing them together like this?

AnaCoda: Pretty much the only ~500 uses of anonymous object creations in the LMS are in parenthesized lambda expressions to do `m_logger.Debug()` or similar operations

omsmith: That's true, and we do need to test combinations of things, but this is the only test for parenthesized lambdas, and the only test for anonymous object creations. I think we could use more tests for both so we're describing our target behaviour, and to show that they aren't actually tied together (while anonymous object creations might mostly be used in logging, we use parenthesized lambdas in many places).

Not going into it in this comment, but maybe lets brainstorm a number of examples of parenthesized lambda usages and anonymous object creations, and go from there?

(Also, I see something more like 2000+ hits. Did you mean something else by ~500: https://search.d2l.dev/search?project=lms&full=%22new+%7B%22&defs=&refs=&path=-*Test*&hist=&type=csharp&xrd=&nn=1&n=25&start=2325).

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -321,11 +321,11 @@ internal sealed class AsyncToSyncMethodTransformerTests {
 
 	[Test]
 	public void ParenthesizedAnonymousCreationLambda() {
-		var actual = Transform( @"[GenerateSync] Task BarAsync() { () => new { BazAsync = TaskCreator.NewTask, Quux = id.ToStringAsync( ""N"" ) } }" );
+		var actual = Transform( @"[GenerateSync] Task BarAsync() { await BazAsync(() => new { BazAsync = 5, Quux = ""test"" } ); }" );
```
omsmith, line 324: Note, this `await` isn't valid unless the function is `async`

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -402,6 +408,15 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
 		);
 	}
 
+	private SeparatedSyntaxList<AnonymousObjectMemberDeclaratorSyntax> TransformAnonDecls( SeparatedSyntaxList<AnonymousObjectMemberDeclaratorSyntax> anonDecls ) {
+		return SyntaxFactory.SeparatedList(
+			anonDecls.Select( anonDecl => anonDecl
+				.WithNameEquals( anonDecl.NameEquals?.WithName( anonDecl.NameEquals.Name.WithIdentifier( RemoveAsyncSuffix( anonDecl.NameEquals.Name.Identifier, optional: true ) ) ) )
```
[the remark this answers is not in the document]
AnaCoda: Sure

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
AnaCoda: Yeah the 500 was a different search I made

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -299,12 +299,21 @@ private ExpressionSyntax Transform( ExpressionSyntax expr )
 			SimpleLambdaExpressionSyntax lambExpr => lambExpr
 				.WithModifiers( RemoveAsyncModifier( lambExpr.Modifiers ) )
 				.WithParameter( Transform( lambExpr.Parameter ) )
-				.WithExpressionBody( lambExpr.ExpressionBody != null ? Transform( lambExpr.ExpressionBody ) : null ),
+				.WithExpressionBody( lambExpr.ExpressionBody != null ? Transform( lambExpr.ExpressionBody ) : null )
+				.WithBlock( lambExpr.Block != null ? Transform( lambExpr.Block ) : null ),
```
AnaCoda, line 303: Let me know if you think I should move this change to another PR, it was a small miss

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -202,6 +202,15 @@ internal sealed class AsyncToSyncMethodTransformerTests {
 		Assert.AreEqual( "[Blocking] void Bar() { Func<int, Task> baz = quux => fred.Delay( 2*y ); }", actual.Value.ToFullString() );
 	}
 
+	[Test]
+	public void SimpleLambdaBlock() {
+		var actual = Transform( @"[GenerateSync] async Task BarAsync() { Func<int, Task> bazAsync = async quuxAsync => { await fredAsync.Delay( 2*y ); await zoopAsync(); } }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] void Bar() { Func<int, Task> baz = quux => { fred.Delay( 2*y ); zoop(); } }", actual.Value.ToFullString() );
```
omsmith, line 211: I believe the type of `baz` is `Action<int>` now

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -319,6 +328,43 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		// For now we are not transforming the variable names
+		Assert.AreEqual( @"[Blocking] void Bar() { Baz(() => new { BazAsync = 5,Quux = ""test"" } ); }", actual.Value.ToFullString() );
+	}
+
+	[Test]
+	public void AnonymousCreation() {
+		var actual = Transform( @"[GenerateSync] async Task BarAsync() { var options = new { size = new { width = GetCurrentWidthAsync(), height = 200 } }; await JsonSerializer.SerializeAsync(options); }" );
```
j3parker, line 343: There is no `await` on `GetCurrentWidthAsync()` here, which would presumably have some `Task<T>` type from the context.

The rewriter is turning it into a call to `GetCurrentWidth()` which you'd imagine would have type `T`, at least not `Task<T>`.... so the type of `size` would be different in the rewritten code.

I'm just observing here, not implying something should be different. It's OK but I think this is a case where the generator is doing something that will generate compiler errors... which is OK, not the worst thing.

If we put an `await` in the test then the test might be demonstrating the utility of the generator better, even if it would obscure the quirk.

AnaCoda: Good observation. I think I'll leave it as is for now

PR #905, opened 2023-08-02
Rm "async" anywhere from IAsyncEnumerable & methods returning it

AnaCoda:
[TA177961](https://rally1.rallydev.com/#/?detail=/task/718746909081&fdp=true)

Motivation is that the sync generator currently can't convert IAsyncEnumerable and most related functions because the Async keyword isn't at the end, for example here:
https://github.com/Brightspace/lms/blob/c4f2338b4d7f6b582b8c56ac959fd953f0846ec3/lp/framework/core/D2L.LP/LP/Security/OAuth2/AccessTokens/Keys/MsSqlKeyDataProvider.cs#L63

In general, there is no clear naming convention for methods that return IAsyncEnumerable except for adding the keyword async somewhere: https://stackoverflow.com/questions/59439281/is-there-a-definitive-naming-convention-for-methods-returning-iasyncenumerable/59452285#59452285

So for converting these we will change all variables/return types defined IAsyncEnumerable to IEnumerable, as well as removing "async" from any method that returns an IAsyncEnumerable.

I also had to make new test methods for this because we needed another method in the class that returns an IAsyncEnumerable

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -125,6 +125,15 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 			}
 		}
 
+		switch( returnTypeInfo.Type.MetadataName ) {
+			case "IAsyncEnumerable":
+				return SyntaxFactory.ParseTypeName( "IEnumerable" ).WithTriviaFrom( typeSynt );
+			case "IAsyncEnumerable`1":
+				return ( (GenericNameSyntax)typeSynt )
+					.WithIdentifier( SyntaxFactory.Identifier( "IEnumerable" ) )
+					.WithTriviaFrom( typeSynt );
+		}
+
 		if( isReturnType ) { ReportDiagnostic( Diagnostics.NonTaskReturnType, typeSynt.GetLocation() ); }
 		return typeSynt;
 	}
@@ -329,6 +338,11 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 	}
 
 	private ExpressionSyntax Transform( InvocationExpressionSyntax invocationExpr) {
+		if( Model.GetTypeInfo( invocationExpr ).ConvertedType?.Name == "IAsyncEnumerable" ) {
+			// Remove all "Async" in the expression
+			return SyntaxFactory.ParseExpression( invocationExpr.ToString().Replace( "Async", "" ) );
+		}
+
 		ExpressionSyntax newExpr = invocationExpr;
 		var memberAccess = invocationExpr.Expression as MemberAccessExpressionSyntax;
 
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -339,7 +339,6 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 
 	private ExpressionSyntax Transform( InvocationExpressionSyntax invocationExpr) {
 		if( Model.GetTypeInfo( invocationExpr ).ConvertedType?.Name == "IAsyncEnumerable" ) {
-			// Remove all "Async" in the expression
 			return SyntaxFactory.ParseExpression( invocationExpr.ToString().Replace( "Async", "" ) );
 		}
 
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -333,7 +333,8 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 	}
 
 	private ExpressionSyntax Transform( InvocationExpressionSyntax invocationExpr) {
-		if( Model.GetTypeInfo( invocationExpr ).ConvertedType?.Name == "IAsyncEnumerable" ) {
+		ITypeSymbol? returnTypeInfo = Model.GetTypeInfo( invocationExpr ).Type;
+		if( returnTypeInfo?.MetadataName == "IAsyncEnumerable`1" && returnTypeInfo.ContainingNamespace.ToString() == "System.Collections.Generic" ) {
 			return invocationExpr.WithExpression( SyntaxFactory.ParseExpression( invocationExpr.Expression.ToString().Replace( "Async", "" ) ) );
 		}
 
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
@@ -123,9 +123,7 @@ internal sealed class AsyncToSyncMethodTransformer : SyntaxTransformer {
 					);
 					return typeSynt;
 			}
-		}
-
-		if( returnTypeInfo.Type.MetadataName == "IAsyncEnumerable`1" ) {
+		} else if( returnTypeInfo.Type.MetadataName == "IAsyncEnumerable`1" && returnTypeInfo.Type.ContainingNamespace.ToString() == "System.Collections.Generic" ) {
 			return ( (GenericNameSyntax)typeSynt )
 				.WithIdentifier( SyntaxFactory.Identifier( "IEnumerable" ) )
 				.WithTriviaFrom( typeSynt );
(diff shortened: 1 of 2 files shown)


src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -125,6 +125,15 @@ AttributeSyntax attribute
 			}
 		}
 
+		switch( returnTypeInfo.Type.MetadataName ) {
```
JeffAshton, line 128: Are both of these cases possible?  There is only one `IAsyncEnumerable<T>` from my understanding.  Can we check the containing namespace as well?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -329,6 +338,10 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
 	}
 
 	private ExpressionSyntax Transform( InvocationExpressionSyntax invocationExpr) {
+		if( Model.GetTypeInfo( invocationExpr ).ConvertedType?.Name == "IAsyncEnumerable" ) {
+			return SyntaxFactory.ParseExpression( invocationExpr.ToString().Replace( "Async", "" ) );
```
JeffAshton, line 342: Would this replace parameter names as well if they contain `Async`?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -328,6 +328,15 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
 
+	[Test]
+	public void IAsyncEnumerable() {
+		var actual = TransformWithIAsyncEnumerator( @"[GenerateSync] async Task BarAsync() { IAsyncEnumerable<string> m_enum = MethodReturningIAsyncEnumerable(); }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( @"[Blocking] void Bar() { IEnumerable<string> m_enum = MethodReturningIEnumerable(); }", actual.Value.ToFullString() );
+	}
+
```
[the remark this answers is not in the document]
AnaCoda: You are right! Good catch and I'll add the test and fix it

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
AnaCoda: The debugger just says global namespace, not System.Collections.Generic... does that sound right and is it specific enough to be checked?

AnaCoda: Nevermind, figured it out thanks!

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -329,6 +333,10 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
 	}
 
 	private ExpressionSyntax Transform( InvocationExpressionSyntax invocationExpr) {
+		if( Model.GetTypeInfo( invocationExpr ).ConvertedType?.Name == "IAsyncEnumerable" ) {
```
JeffAshton, line 336: Could this check the full type instead of just the name?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -475,6 +515,28 @@ class TestType{{{methodSource}}}" );
...
+	public static (Compilation, MethodDeclarationSyntax) ParseMethodWithIAsyncEnumerable( string methodSource ) {
+		var wrappedAndParsed = CSharpSyntaxTree.ParseText( @$"
+using System.Threading.Tasks;
+using System.Threading;
+using System.Collections.Generic;
+using D2L.CodeStyle.Annotations;
+
+class TestType{{
+{methodSource}async IAsyncEnumerable<string> MethodReturningIAsyncEnumerable() {{
```
JeffAshton, line 527: For completeness so that both cases above are valid code:

```suggestion
{methodSource}async IAsyncEnumerable<string> MethodReturningIAsyncEnumerable( int p = 0 ) {{
```

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -329,6 +333,11 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
 	}
 
 	private ExpressionSyntax Transform( InvocationExpressionSyntax invocationExpr) {
+		ITypeSymbol? returnTypeInfo = Model.GetTypeInfo( invocationExpr ).Type;
+		if( returnTypeInfo?.MetadataName == "IAsyncEnumerable`1" && returnTypeInfo.ContainingNamespace.ToString() == "System.Collections.Generic" ) {
+			return invocationExpr.WithExpression( SyntaxFactory.ParseExpression( invocationExpr.Expression.ToString().Replace( "Async", "" ) ) );
```
j3parker, line 338: I'm a little nervous about calling `ParseExpression` here... there are unimportant amusing consequences like how `var x = SomethingAsyncSomething() /* blah Async blah */;` would get transformed into `var x = SomethingSomething() /* blah blah */;` I think? (like substitution within the comment)

I wonder if we should just have a state variable on the transformer that tells `RemoveAsyncSuffix` to not just do suffixes but do `.Replace` instead. Then the inside of this if would just set that to true, and we would restore the previous state of that variable on the way out of this function... does that make sense?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -73,6 +75,13 @@ AttributeSyntax attribute
 		);
 
 	private SyntaxToken RemoveAsyncSuffix( SyntaxToken ident, bool optional = false ) {
+		if( m_removeAsyncAnywhere && ident.ValueText.Contains( "Async" ) ) {
```
j3parker, line 78: Should this check `optional`?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -338,6 +351,11 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
 	}
 
 	private ExpressionSyntax Transform( InvocationExpressionSyntax invocationExpr) {
+		ITypeSymbol? returnTypeInfo = Model.GetTypeInfo( invocationExpr ).Type;
+		if( returnTypeInfo?.MetadataName == "IAsyncEnumerable`1" && returnTypeInfo.ContainingNamespace.ToString() == "System.Collections.Generic" ) {
+			m_removeAsyncAnywhere = true;
```
j3parker, line 356: I think we probably want to store the previous value of `m_removeAsyncAnywhere` (it could be true or false here) and restore it on the way back out of the function?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
AnaCoda: I don't think so, because there are cases where a function may return IAsyncEnumerable and not end in Async and we want the sync generator to transform it without a diagnostic

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -369,9 +371,13 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
 			return SyntaxFactory.ParseExpression( $"Task.Run(() => {invocationExpr}).Result" );
 		}
 
-		return invocationExpr
+		invocationExpr = invocationExpr
 			.WithExpression( Transform( invocationExpr.Expression ) )
 			.WithArgumentList( TransformAll( invocationExpr.ArgumentList, Transform ) );
+
+		m_removeAsyncAnywhere = prevRemoveAsyncAnywhereState;
```
j3parker, line 378: There are a few exit points in this function where it doesn't get restored.
One way to deal with that is `try`+`finally`:

```
bool prevRemoveAsyncAnywhereState = m_removeAsyncAnywhere;

try {
  // ...
finally {
  m_removeAsyncAnywhere = prevRemoveAsyncAnywhereState;
}
```

no matter how you get out of the `try` block (even exceptions, which is maybe the primary use case for `finally` admittedly (but we don't care)) the `finally` block will execute.

j3parker (approved):

PR #907, opened 2023-08-09
Remove CancellationToken from declaration parameters

AnaCoda:


src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -466,15 +467,35 @@ private ArgumentListSyntax Transform( ArgumentListSyntax argList )
...
 
-	private ParameterSyntax Transform( ParameterSyntax parameter )
-		=> parameter.WithIdentifier(RemoveAsyncSuffix( parameter.Identifier, optional: true ) );
+	private SeparatedSyntaxList<ParameterSyntax> TransformParams( SeparatedSyntaxList<ParameterSyntax> paramSynts ) {
+		for( int i = 0; i < paramSynts.Count; i++ ) {
+			ParameterSyntax? newParamSynt = Transform( paramSynts[i] );
+			if( newParamSynt is not null ) {
+				paramSynts = paramSynts.Replace( paramSynts[i], newParamSynt );
+			} else {
+				paramSynts = paramSynts.RemoveAt( i );
```
j3parker, line 486: I think this might not work when the CancellationToken is in the middle? because we are modifying the size of `paramSynts` while we are iterating over it, so if we remove the ith element the next run of the loop will be `i+1` and miss an element of `paramSynts`.

There are a few solutions. One is to do `i--` here.

Another is to build a new `List<ParameterSyntax>` to pass into `SyntaxFactory.SeparatedList`. Instead of calling `Replace` above you would `Add` to the list, and instead of `RemoveAt` you would just not do anything.

One thing to think about is how (I think) it's probably a not-uncommon case for the parameter list to not be modified. The ideal thing in a situation like that would be if we ultimately did `return paramSynts` (on the original/unmodified value). To make that happen you could lazily allocate that `List<ParameterSyntax>` (make it null to start, only allocate it with `new` once you find something that `newParamSynt != paramSynts[i]` for the first time, but then you also have to `Array.Copy` all the things from 0 to i-1 to "catch up"). But since this is a parameter list that only happens once per `[GenerateSync]` I probably wouldn't worry about it (if it was argument lists instead I might -- there could be many argument lists per `[GenerateSync]` and its probably much more common for them to go unmodified.)

j3parker: Can you add a test for

```cs
[GenerateSync]
Task FooAsync( int w, int x, CancellationToken c, int y, int z );
```
?

tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -337,6 +337,15 @@ internal sealed class AsyncToSyncMethodTransformerTests {
...
 	}
 
+	[Test]
+	public void CancellationTokenParameter() {
+		var actual = Transform( @"[GenerateSync] Task<Foo> BarAsync( int Baz, CancellationToken ct ) { return FooAsync( Baz + 1, ct ); }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( "[Blocking] Foo Bar( int Baz) { return Foo( Baz + 1); }", actual.Value.ToFullString() );
+	}
```
j3parker, line 347: This is good 👍 
We're going to have a separate PR to address invocation argument lists right?
Like a call to `await FooAsync( x, y, ct );` currently becomes `Foo( x, y, ct );` (I think) but we want it to be `Foo( x, y )`, right?
(Separate PR for that is good.)

j3parker: Oh wait nvm I see that's happening in this test, I'll read the PR more closely :)

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
AnaCoda: I tried passing into SyntaxFactory.SeparatedList first but there was no way to preserve the comma whitespace and I didn't want the params to get squished together, if that makes sense

AnaCoda: Ok, I used the i--; fix and added the test but with another adjacent CancellationToken because the test you proposed was still working since it just skipped removing int y haha

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -311,7 +312,7 @@ private ExpressionSyntax Transform( ExpressionSyntax expr )
 
 			SimpleLambdaExpressionSyntax lambExpr => lambExpr
 				.WithModifiers( RemoveAsyncModifier( lambExpr.Modifiers ) )
-				.WithParameter( Transform( lambExpr.Parameter ) )
+				.WithParameter( Transform( lambExpr.Parameter, removeCancellationToken: false ) ?? lambExpr.Parameter )
```
omsmith, line 315: Can we add a test showing why/the result of passing `false` here? Or do we already have that?

omsmith: Especially given the null coalescing operator. When does that come in to play?

AnaCoda: Added the test. The null coalescing operator shouldn't actually ever come into use if removeCancellationToken: false  but the compiler can't tell that the function can't return null if it's false so I had to add it

omsmith: Perhaps a separate function that can't return null would be better then? Or perhaps the logic should live in `TransformParams`?

AnaCoda: Thanks! Moving the logic to TransformParams seems cleaner

omsmith (approved):

PR #909, opened 2023-08-10
Disable CS1572 at top of generated files

AnaCoda:
Workaround so we don't have to selectively remove XML parameters for now


j3parker (approved):

j3parker:
I think there was another warning we suppressed but we did it a harder way? IMO we shouldn't be too worried about doing this in the generator (in this case its temporary until a minor defect is fixed, but still -- easy, cheap)

PR #913, opened 2023-08-16
Remove catch TaskCanceledException when Generating

AnaCoda:


tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
```diff
@@ -448,6 +448,15 @@ internal sealed class AsyncToSyncMethodTransformerTests {
 		Assert.AreEqual( @"[Blocking] void Bar() { Console.WriteLine($""Hello, {GetUsername()}! Welcome back to {m_baz}, it is {GetTime()} {m_ampm}.""); }", actual.Value.ToFullString() );
 	}
 
+	[Test]
+	public void TaskCanceledException() {
+		var actual = Transform( @"[GenerateSync] async Task BarAsync() { try { response = await GetBazAsync().ConfigureAwait( false ); } catch( TaskCanceledException exception ) { Console.WriteLine( ""Failed"" ); } }" );
+
+		Assert.IsTrue( actual.Success );
+		Assert.IsEmpty( actual.Diagnostics );
+		Assert.AreEqual( @"[Blocking] void Bar() { try { response = GetBaz(); } catch( OperationCanceledException exception ) { Console.WriteLine( ""Failed"" ); } }", actual.Value.ToFullString() );
```
j3parker, line 457: Hmmmm.... I'm not sure about this one? Like maybe we should just exclude `catch( TaskCanceledException )` guards entirely? I'm not sure `OperationCanceledException` is really an equivalent.

@omsmith wdyt?

AnaCoda: I based this off of the fact that TaskCanceledException derives from OperationCanceledException: https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskcanceledexception?view=net-7.0

omsmith: Yeah, I would generally expect TaskCanceledExceptions to disappear in entirely sync flows.

Anywhere we're just doing `Task.Run( AsyncThing() ).Result` is still going to have them though.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -124,6 +127,9 @@ AttributeSyntax attribute
 					return ( (GenericNameSyntax)typeSynt )
 						.TypeArgumentList.Arguments.First()
 						.WithTriviaFrom( typeSynt );
+				case "TaskCanceledException":
+					m_removeCatchClause = !m_disableTaskRunWarningFlag;
```
AnaCoda, line 131: This has the unfortunate side effect that if anything is wrapped in Task.Run() in the entire method by the generator, then we will not remove a TaskCanceledException catch clause.

I think this is acceptable because it would be pretty rare, and we're doing it in the safer direction (it's fine if we still have a catch even if it's never reached, not fine if we remove a catch that could be reached). It also reduces generator complexity.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -435,6 +437,17 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
 			.WithVariables( TransformVariables( varDecl.Variables ) );
 	}
 
+	private StatementSyntax Transform( TryStatementSyntax tryStmt ) {
+		m_catchClauseCount = tryStmt.Catches.Count;
```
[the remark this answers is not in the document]
AnaCoda: Thanks! Good idea

j3parker (approved):

PR #915, opened 2023-08-17
Adding support for conversion operators to the ConstantAttributeAnalyzer

JeffAshton:


omsmith:
Okay cool so this is adding support for say `string` to  `Constant<string>`, but not the reverse, right?

omsmith (approved):

JeffAshton:
> Okay cool so this is adding support for say `string` to `Constant<string>`, but not the reverse, right?

Correct.  I really want to avoid the other direction.

PR #917, opened 2023-08-18
Generate finally clause and add tests including catch removal

AnaCoda:


tests/D2L.CodeStyle.Analyzers.Test/Async/Generator/AsyncToSyncMethodTransformerTests.cs
omsmith, line 1: Just for completeness, can you add a test for `try {} finally { ... }` (an empty `try` block with a non-empty `finally` block).

omsmith (approved):

PR #921, opened 2023-09-29
Moving pinning attributes and the pinned analyzer

alamarre:
This could be broken up into a couple PRs. Either way we'll want bring in the analyzer before switching the attributes and removing the analyzers from the LMS. The recursive pinned analyzer in the LMS isn't where we'll want it to be, due to not incorporating changes that prevent it from erroring immediately so we can remove that from the LMS before bring in the new one.

I also refactored what's in the LMS repo to clean it up a bit and start from the symbol rather than the syntax node.

### DIFF
--- a/src/D2L.CodeStyle.Annotations/D2L.CodeStyle.Annotations.csproj
+++ b/src/D2L.CodeStyle.Annotations/D2L.CodeStyle.Annotations.csproj
@@ -21,8 +21,4 @@
     <IncludeSymbols>True</IncludeSymbols>
   </PropertyGroup>
 
-  <ItemGroup>
-    <Folder Include="Pinning\" />
-  </ItemGroup>
-
 </Project>
--- a/src/D2L.CodeStyle.Annotations/D2L.CodeStyle.Annotations.csproj
+++ b/src/D2L.CodeStyle.Annotations/D2L.CodeStyle.Annotations.csproj
@@ -21,4 +21,8 @@
     <IncludeSymbols>True</IncludeSymbols>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Folder Include="Pinning\" />
+  </ItemGroup>
+
 </Project>
--- a/src/D2L.CodeStyle.Annotations/Pinning/MustBeDeserializableAttribute.cs
+++ b/src/D2L.CodeStyle.Annotations/Pinning/MustBeDeserializableAttribute.cs
@@ -2,9 +2,9 @@
 
 namespace D2L.CodeStyle.Annotations.Pinning {
 	/// <summary>
-	/// Represents a parameter in a dangerous call chain (e.g. serialization using the assembly qualified name) that must be pinned recursively
+	/// Represents a parameter in the call chain of serialization in order to enforce pinning or otherwise set limitations on serialized types
 	/// </summary>
 	[AttributeUsage( AttributeTargets.Parameter | AttributeTargets.GenericParameter, AllowMultiple = false, Inherited = false )]
-	public sealed class MustBeRecursivelyPinnedAttribute : Attribute {
+	public sealed class MustBeDeserializableAttribute : Attribute {
 	}
 }
--- a/src/D2L.CodeStyle.Annotations/Pinning/MustBePinnedAttribute.cs
+++ b/src/D2L.CodeStyle.Annotations/Pinning/MustBePinnedAttribute.cs
@@ -1,0 +1,10 @@
+﻿using System;
+
+namespace D2L.CodeStyle.Annotations.Pinning {
+	/// <summary>
+	/// Represents a parameter in a dangerous call chain (e.g. serialization using the assembly qualified name) that must be pinned
+	/// </summary>
+	[AttributeUsage( AttributeTargets.Parameter | AttributeTargets.GenericParameter, AllowMultiple = false, Inherited = false )]
+	public sealed class MustBePinnedAttribute : Attribute {
+	}
+}
--- a/src/D2L.CodeStyle.Annotations/Pinning/MustBePinnedRecursivelyPinnedAttribute.cs
+++ b/src/D2L.CodeStyle.Annotations/Pinning/MustBePinnedRecursivelyPinnedAttribute.cs
@@ -1,0 +1,10 @@
+﻿using System;
+
+namespace D2L.CodeStyle.Annotations.Pinning {
+	/// <summary>
+	/// Represents a parameter in a dangerous call chain (e.g. serialization using the assembly qualified name) that must be pinned recursively
+	/// </summary>
+	[AttributeUsage( AttributeTargets.Parameter | AttributeTargets.GenericParameter, AllowMultiple = false, Inherited = false )]
+	public sealed class MustBeRecursivelyPinnedAttribute : Attribute {
+	}
+}
--- a/src/D2L.CodeStyle.Annotations/Pinning/PinnedAttribute.cs
+++ b/src/D2L.CodeStyle.Annotations/Pinning/PinnedAttribute.cs
@@ -1,0 +1,22 @@
+﻿using System;
+
+namespace D2L.CodeStyle.Annotations.Pinning {
+	/// <summary>
+	/// The type should not be moved from its assembly due to risk of deserialization based on assembly qualified name or other dangerous call chains
+	/// </summary>
+	[AttributeUsage( AttributeTargets.Class
+		| AttributeTargets.Struct
+		| AttributeTargets.Interface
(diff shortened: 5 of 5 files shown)


src/D2L.CodeStyle.Analyzers/Diagnostics.cs
```diff
@@ -757,5 +757,15 @@ public static class Diagnostics {
 			defaultSeverity: DiagnosticSeverity.Error,
 			isEnabledByDefault: true
 		);
+
+		public static readonly DiagnosticDescriptor PinnedTypesMustNotMove = new DiagnosticDescriptor(
+			id: "D2L0102",
```
alamarre, line 762: This will likely end up in conflict and I'll bump it closer to when we want to land it.

src/D2L.CodeStyle.Analyzers/Pinning/PinnedAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+				return;
+			}
+
+			Location? location = classSymbol.Locations.FirstOrDefault();
+			if( !PinnedAnalyzerHelper.TryGetPinnedAttribute( classSymbol, pinnedAttributeSymbol, out AttributeData? attribute ) ) {
+				return;
+			}
+
+			string? fqName = attribute?.ConstructorArguments[0].Value?.ToString();
+			string? assembly = attribute?.ConstructorArguments[1].Value?.ToString();
```
cesar-d2l, line 50: we're not doing anything with the 3rd argument. Is this coming in a follow-up?

tests/D2L.CodeStyle.Analyzers.Test/Pinning/PinnedTests.cs
```diff
@@ -0,0 +1,95 @@
+﻿using System.Collections.Immutable;
+using D2L.CodeStyle.Analyzers.Pinning;
+using Microsoft.CodeAnalysis;
+using NUnit.Framework;
+
+namespace D2L.CodeStyle.Analyzers.Tests.Pinning {
```
j3parker, line 6: Could these be done as spec tests instead? See the "spec" folder in this project (spec = specification 🤷 )

src/D2L.CodeStyle.Analyzers/Pinning/PinnedAnalyzer.cs
alamarre: Yes, there are analyzers for recursive pinning and the must be pinned that are separate.

tests/D2L.CodeStyle.Analyzers.Test/Pinning/PinnedTests.cs
```diff
@@ -0,0 +1,95 @@
...
+namespace D2L.CodeStyle.Annotations.Pinning;
+
+[Pinned( assembly: ""D2L.Bar"", fullyQualifiedName: ""D2L.Bar.X"", pinnedRecursively: false )]
+public {type} X {{
+}}
+";
+			ImmutableArray<Diagnostic> diagnostics = await GetDiagnostics( input );
+			IEnumerable<Diagnostic> matching = diagnostics.Where( d => d.Id == Diagnostics.PinnedTypesMustNotMove.Id );
+			Assert.AreEqual( 1, matching.Count() );
+		}
```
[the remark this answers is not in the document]
alamarre: I think what you're referring to is handled by the other analyzers.

tests/D2L.CodeStyle.Analyzers.Test/Pinning/PinnedTestsBase.cs
```diff
@@ -0,0 +1,72 @@
...
+				MetadataReference.CreateFromFile(typeof(CSharpCompilation).Assembly.Location),
+				MetadataReference.CreateFromFile(typeof(object).Assembly.Location),
+				MetadataReference.CreateFromFile(typeof(PinnedAttribute).Assembly.Location),
+				MetadataReference.CreateFromFile(typeof(PinnedAttributeAnalyzer).Assembly.Location, MetadataReferenceProperties.Assembly),
+			};
+			m_metadataReferences = references.ToImmutableArray();
+
+			m_analyzers = analyzers.ToImmutableArray();
+
+			AdditionalText allowedList = new StringAdditionalText( "UnpinnedAllowedList.txt", $@"System.Collections.Generic.Dictionary`2" );
```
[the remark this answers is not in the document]
alamarre: The other analyzers use it. I removed the other methods that aren't used, but left minor changes like this so that it was less work to do it as 3 PRs without missing anything, since some of the changes came from working on the must be pinned analyzer, but were for the recursive pinning.

tests/D2L.CodeStyle.Analyzers.Test/Pinning/PinnedTests.cs
alamarre: Is there documentation on them?

tests/D2L.CodeStyle.Analyzers.Test/Pinning/PinnedTests.cs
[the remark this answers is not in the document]
cesar-d2l: ah just to be clear on what I referring to, something like:

```c#
public class X {
    [Pinned( ... )]
    public class Y {}
}
```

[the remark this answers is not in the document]
alamarre: I've added tests for them now.

alamarre:
> This could be broken up into a couple PRs. 

Doing so after discussion of changing must be recursively pinned to must be deserializable.

src/D2L.CodeStyle.Annotations/D2L.CodeStyle.Annotations.csproj
```diff
@@ -21,4 +21,8 @@
     <IncludeSymbols>True</IncludeSymbols>
   </PropertyGroup>
 
+  <ItemGroup>
```
cesar-d2l, line 24: Is this needed? This seems unusual. I only seen this when adding an empty directory, but the folder looks populated

cesar-d2l (dismissed):

DaleHennessey (approved):

omsmith:
Just for reference - curious why we're moving these here from the LMS? I'm assuming its so they can be used in D2L.Core or something?

alamarre:
> Just for reference - curious why we're moving these here from the LMS? I'm assuming its so they can be used in D2L.Core or something?

That is the primary driver. It could be handled in another way, but it also does enforce that the analyzers run on all the LMS libraries in a way that's not easily / accidentally undone.

src/D2L.CodeStyle.Annotations/D2L.CodeStyle.Annotations.csproj
```diff
@@ -21,4 +21,8 @@
     <IncludeSymbols>True</IncludeSymbols>
   </PropertyGroup>
 
+  <ItemGroup>
+    <Folder Include="Pinning\" />
+  </ItemGroup>
+
```
mthjones, line 27:
```suggestion
```

mthjones (approved):

cesar-d2l (approved):

j3parker:
Has this PR been descoped? It looks like the analyzer is no longer included

alamarre:
> Has this PR been descoped? It looks like the analyzer is no longer included

Yes, as per my comment on the 11th due in part to the renaming of the one attribute. 

I also wanted to split it if moving to spec tests since it's a further digression from what's in the LMS. I need to do some cleanup, but the analyzer is using spec tests on a branch. https://github.com/Brightspace/D2L.CodeStyle/compare/main...alamarre/add-pinned-analyzer

I had also noticed issues with running the spec tests cross-platform and in .NET 7 so I also did a POC to get those working to be more confident in the spec tests before switching to them. https://github.com/Brightspace/D2L.CodeStyle/compare/main...alamarre/wip-test

tests/D2L.CodeStyle.Analyzers.Test/Pinning/PinnedTests.cs
alamarre: I've moved the analyzer to it's own branch and implemented spec tests in there. You can preview them if you want. https://github.com/Brightspace/D2L.CodeStyle/compare/main...alamarre/add-pinned-analyzer

mthjones (approved):

omsmith (approved):
Don't really have any context for the plan here, but adding some attributes to Annotations without corresponding analyzers won't hurt anything, so cool.

I'll bump the version after you merge.

PR #922, opened 2023-10-23
Add SerializationFrameworkAnalyzer

Matthew-nop:
This analyzer restricts the use of these interfaces and their implementations:
- `ISerializationWriter`
- `ISerializationWriterExtensions`
- `ISerializer`
- `ITrySerializer`

Invocation, property, and field references are only permitted in classes annotated with the `[SerializationFramework]` attribute

There is one small problem/catch to this analyzer though - implementing types may still be referenced through another safe interface. Given some safe interface `a` and a dangerous interface `d`, a serialiser `s` such that `s : a, d` may invoke dangerous behaviour through `a`. Although, I don't _think_ this can be solved without a noticeable/significant performance hit.

I built the LMS with this analyzer to roughly gauge the performance. It seems to be a bit worse than the OBSL analyser which was the basis for a lot of this code. I figure this is a because for each type `t` we sometimes check `t.AllInterfaces.Any( dangerousInterface.Contains )` in addition to the `dangerousInterfaces.Contains( t )` that the OBSL analyser checks
image

Before merging, the annotations should be added in the LMS:
- https://github.com/Brightspace/lms/pull/42303
- https://github.com/Brightspace/lms/pull/42304
- https://github.com/Brightspace/lms/pull/42305
- https://github.com/Brightspace/lms/pull/42306

ref: https://github.com/Brightspace/lms-dotnet-migration-issues/issues/135

### DIFF
--- a/tests/D2L.CodeStyle.Analyzers.Test/Specs/SerializationFrameworkAnalyzer.cs
+++ b/tests/D2L.CodeStyle.Analyzers.Test/Specs/SerializationFrameworkAnalyzer.cs
@@ -1,0 +1,157 @@
+﻿// analyzer: D2L.CodeStyle.Analyzers.ApiUsage.Serialization.SerializationFrameworkAnalyzer
+
+using System;
+using D2L.LP.Serialization
+
+namespace D2L.LP.Serialization {
+
+	public sealed class SerializationFrameworkAttribute : Attribute { }
+
+	[SerializationFramework]
+	public class DangerousExplicitClass : ISerializer, ITrySerializer, ISerializationWriter {
+		public void Foo() { }	
+		void ISerializer.Serialize() { }
+		bool ITrySerializer.TrySerialize() { return true; }
+		void ISerializationWriter.Write() { }
+	}
+
+	[SerializationFramework]
+	public class DangerousImplicitClass : ISerializer, ITrySerializer, ISerializationWriter {
+		public void Serialize() { }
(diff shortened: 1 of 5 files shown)


tests/D2L.CodeStyle.Analyzers.Test/Specs/SerializationFrameworkAnalyzer.cs
```diff
@@ -0,0 +1,157 @@
...
+	[SerializationFramework]
+	public class DangerousExplicitClass : ISerializer, ITrySerializer, ISerializationWriter {
+		public void Foo() { }	
+		void ISerializer.Serialize() { }
+		bool ITrySerializer.TrySerialize() { return true; }
+		void ISerializationWriter.Write() { }
+	}
+
+	[SerializationFramework]
+	public class DangerousImplicitClass : ISerializer, ITrySerializer, ISerializationWriter {
```
cesar-d2l, line 19: I expect the analyzer to catch this, but you may want a class that inherits `DangerousImplicitClass` and checking that the analyzer catches that.

Matthew-nop:
Putting this on hold for now - investigating potentially adding a safe alternative interface to `ISerializationWriter` that serialisers can call

Matthew-nop:
Closing this for now. This may be used as a plan B, but we've opted to go with https://github.com/Brightspace/lms/pull/42621 to enforce pinned type safety within serialization internals.

PR #923, opened 2023-10-30
Add the Pinned Analyzer to prevent pinned types from moving accidentally

alamarre:


src/D2L.CodeStyle.Analyzers/Pinning/PinnedAttributeAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+
+		private void OnCompilationStart(
+			CompilationStartAnalysisContext context
+		) {
+			context.RegisterSymbolAction( AnalyzeSymbol,
+				SymbolKind.NamedType );
+		}
+
+		private static void AnalyzeSymbol( SymbolAnalysisContext context ) {
+			INamedTypeSymbol? pinnedAttributeSymbol = context.Compilation.GetTypeByMetadataName( PinnedAnalyzerHelper.PinnedAttributeName );
```
mthjones, line 30: nit: it's probably cached by Roslyn, but consider doing this in `OnCompilationStart` and passing it in here so it doesn't need to be called for every symbol.

src/D2L.CodeStyle.Analyzers/Pinning/PinnedAttributeAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.CSharp;
+using Microsoft.CodeAnalysis.CSharp.Syntax;
+using Microsoft.CodeAnalysis.Diagnostics;
+
+namespace D2L.CodeStyle.Analyzers.Pinning {
+	[DiagnosticAnalyzer( LanguageNames.CSharp )]
+	internal sealed class PinnedAttributeAnalyzer : DiagnosticAnalyzer {
+
+		private static readonly TypeKind[] TypeKindsSupported = new[] { TypeKind.Class, TypeKind.Interface };
```
mthjones, line 11: The attribute also supports being applied to `struct`. Is this check necessary given that we've already cast to `INamedTypeSymbol` at the point we check this? Can we just trust the `AttributeUsage` definition?

src/D2L.CodeStyle.Analyzers/Pinning/PinnedAttributeAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+		private static void AnalyzeSymbol( SymbolAnalysisContext context ) {
+			INamedTypeSymbol? pinnedAttributeSymbol = context.Compilation.GetTypeByMetadataName( PinnedAnalyzerHelper.PinnedAttributeName );
+			if( pinnedAttributeSymbol == null ) {
+				return;
+			}
+
+			INamedTypeSymbol? classSymbol = context.Symbol as INamedTypeSymbol;
+			if( classSymbol == null ) {
+				return;
+			}
```
mthjones, line 38: Consider hard-casting this rather than giving the option for a silent failure. This action only runs on `SymbolKind.NamedType` so a cast to `INamedTypeSymbol` should never fail, and if it ever does, we'd want to know very loudly rather than having this silently skip analysis.

src/D2L.CodeStyle.Analyzers/Pinning/PinnedAnalyzerHelper.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+
+namespace D2L.CodeStyle.Analyzers.Pinning {
+	public static class PinnedAnalyzerHelper {
+
+		public const string PinnedAttributeName = "D2L.CodeStyle.Annotations.Pinning.PinnedAttribute";
+
+		public static bool TryGetPinnedAttribute( ISymbol classSymbol, INamedTypeSymbol pinnedAttributeSymbol, out AttributeData? attribute ) {
+			attribute = null;
```
mthjones, line 10: nit: prefer setting this value just before returning `false`. Setting it here is both less performant (unused assignment on `true` path), and can mask logic errors where the value otherwise should have been set to something different.

cesar-d2l: alternatively, you can change the definition to `AttributeData? GetPinnedAttribute(...)`. I'm usually a fan of the `TryABC()` pattern, but with the C# nullablity types it's a bit odd because returning `false` already tells you that, so the caller has to check both the return type and the `out` parameter (or use `attribute!.something()` everywhere)

src/D2L.CodeStyle.Analyzers/Pinning/PinnedAttributeAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+
+			Location? location = classSymbol.Locations.FirstOrDefault();
+			if( !PinnedAnalyzerHelper.TryGetPinnedAttribute( classSymbol, pinnedAttributeSymbol, out AttributeData? attribute ) ) {
+				return;
+			}
+
+			string? fqName = attribute?.ConstructorArguments[0].Value?.ToString();
+			string? assembly = attribute?.ConstructorArguments[1].Value?.ToString();
+			if( fqName == null || assembly == null ) {
```
cesar-d2l, line 51: I think for these to be `null` it would be a compiler error. I guess it doesn't hurt to report the diagnostic, and it's still technically correct

src/D2L.CodeStyle.Analyzers/Pinning/PinnedAttributeAnalyzer.cs
```diff
@@ -0,0 +1,70 @@
...
+				context.ReportDiagnostic( Diagnostic.Create( Diagnostics.PinnedTypesMustNotMove, location, classSymbol.Name ) );
+				return;
+			}
+
+			SymbolDisplayFormat fullyQualifiedNameFormat = new SymbolDisplayFormat(
```
cesar-d2l, line 56: nit: consider making this `private static` class field

src/D2L.CodeStyle.Analyzers/Pinning/PinnedAttributeAnalyzer.cs
cesar-d2l: We also support `record` types. That might be a class/struct because I don't see a record enum

mthjones (commented):
LGTM apart from the missing `return null;`!

mthjones (approved):

cesar-d2l (approved):

alamarre:
Superseded by #945 and may not be necessary due to changes in the plan.

PR #925, opened 2023-11-08
Allow OnlyVisibleToType on interfaces.

TimothyJCowen:


tests/D2L.CodeStyle.Analyzers.Test/Specs/OnlyVisibleToAnalyzer.cs
```diff
@@ -9,7 +9,7 @@
 namespace D2L.CodeStyle.Annotations.Contract {
 
 	[AttributeUsage(
-		validOn: AttributeTargets.Method | AttributeTargets.Property,
+		validOn: AttributeTargets.Interface | AttributeTargets.Method | AttributeTargets.Property,
```
TimothyJCowen, line 12: I'm noticing the lack of the `AttributeTargets.Constructor` here, which is on the main attribute. I'm not sure if that should be here?

j3parker: Do we have tests with constructors? If not I guess I can see why it works. Maybe we should have tests for that... 🤷 Maybe not a big deal

j3parker (approved):

omsmith:
This seems likely to make this analyzer quite a bit more expensive.

What benefit are we looking for by blocking interacting with the type in any way?

The added tests only show declaring variables, could we use `OperationKind.VariableDeclaration`?

PR #928, opened 2023-11-10
Added PoC tests for adding OnlyVisibleToType to interfaces.

TimothyJCowen:


tests/D2L.CodeStyle.Analyzers.Test/Specs/OnlyVisibleToAnalyzer.cs
```diff
@@ -276,88 +286,223 @@ public static class TypeOfTargets {
...
+		public RestrictedWithInheritance() { }
+		string SomeProperty => "SomeProperty";
+		void SomeMethod( string someValue ) { }
+	}
+
+	public class RestrictedWithDoubleInheritance : RestrictedWithInheritance {
+		public RestrictedWithDoubleInheritance() { }
+	}
+
+	[ReleaseVisibilityConstraints( typeof( IInheritanceInterface ) )]
```
omsmith, line 396: I feel think this feature becomes a pretty big footgun? If I'm restricting a type, I'm not too worried about what properties I'm adding to it, but then something else might be re-exposing those down the road.

Do we have an example of when we'd use this?

TimothyJCowen: I ended up pivoting from having a secondary attribute for precisely this reason - I've created [a whole new PoC](https://github.com/Brightspace/D2L.CodeStyle/pull/929) which leaves it up to the original definition to decide which restrictions are inherited by derived types (all, by default) and which only apply to the class/interface that the restriction was explicitly added to.

As far as an example, yes. This would be used to create a "bridge" between the `OrgUnitFileType` and `UserFileType` in the form of an interface (`IFileType`) which can only be used by explicitly specified consumers. Essentially it is a way for Storageable code to handle files at the very base level without caring about the different entity types (and having to implement code for each type separately). We still want consumers up the chain to be able to use the separate `...FileType` classes however, including referencing properties on them that originate from the interface itself, so we don't want the restrictions to be inherited by them.

TimothyJCowen:
Replaced by https://github.com/Brightspace/D2L.CodeStyle/pull/929.

PR #929, opened 2023-11-13
Added PoC for adding OnlyVisibleToType to interfaces.

TimothyJCowen:


src/D2L.CodeStyle.Analyzers/Language/OnlyVisibleToAnalyzer.Model.cs
```diff
@@ -25,19 +25,20 @@ INamedTypeSymbol onlyVisibleToTypeAttribute
 			}
 
 			public static Model? TryCreate( Compilation compilation ) {
+				INamedTypeSymbol? onlyVisibleToTypeAttribute = compilation
+					.GetTypeByMetadataName( OnlyVisibleToTypeAttributeMetadataName );
 
-				INamedTypeSymbol? onlyVisibleToTypeAttribute = compilation.GetTypeByMetadataName( OnlyVisibleToTypeAttributeMetadataName );
```
omsmith, line 29: Aside (since I'm not looking at the implementation right now), extra non-changes like this make things a lot harder to review.

omsmith:
So the way I would expect `OnlyVisibleToType` to work, especially applied to types instead of members, is that we're essentially creating a pseudo-assembly with just those types in them and then setting up references. We could potentially do the same work by refactoring things into different assemblies. As such, I would expect "visibility" to work about the same way.

2023-11-15T15:17:23+00:00

* `[OnlyVisibleToType]` on an interface
  * forces any inheriting interfaces to have at most the same restricted types applied.
  * An inheriting interface could have a subset of the types listed (allowed to be more restrictive - equivalent to `IPrivateInternal` and similar)
  * classes/structs extending the interface (if they were allowed to in the first place) do not inherit the visibility restrictions themselves (`CPublicPrivateI`)

* `[OnlyVisibleToType]` on a class
 * forces any inheriting classes to have at most the same restricted types applied.
  * An inheriting classes could have a subset of the types listed (allowed to be more restrictive - equivalent to `CPrivateInternal` and similar)

* `[OnlyVisibleToType]` on a struct
  * does what we'd expect, but there's no inheritance involved

* Restricted types in members
  * Normally you wouldn't be allowed to define a member that exposes a type outside of its visibility (you cannot define a public property as an internal type). We can't apply this approach since our visibility applies to our pseudo-assembly
  * Instead we need to apply the restriction on the consumption end. A type should not be able to interact with a member that uses the type (return type or parameter) unless its in the pseudo-assembly

* Further restrictions
  * Shouldn't be able to refer to the type in general (variables, parameters, etc), behaving as though there's no references to it

PR #942, opened 2023-11-30
ReadOnlyParameterAnalyzer: handle partial methods

omsmith:


src/D2L.CodeStyle.Analyzers/Immutability/ReadOnlyParameterAnalyzer.cs
```diff
@@ -32,77 +31,101 @@ internal sealed class ReadOnlyParameterAnalyzer : DiagnosticAnalyzer {
...
-				 */
-				ctx.ReportDiagnostic(
-					Diagnostics.ReadOnlyParameterIsnt,
-					syntax.GetLocation(),
-					messageArgs: new[] { "is an in/ref/out parameter" }
-				);
-			}
-
-			IMethodSymbol method = parameter.ContainingSymbol as IMethodSymbol;
-			SyntaxNode methodSyntaxNode = method.DeclaringSyntaxReferences[ 0 ].GetSyntax( ctx.CancellationToken );
```
omsmith, line 76: `method.DeclaringSyntaxReferences` is always the declaration (`partial void Foo();`), never the implementation part (`partial void Foo() {}`), so getting the right syntax can be tricky. Plus we don't want duplicates, so would only want to raise the diagnostic (or do the work) on one ParameterSyntax or the other.

Acting on the operation gets us the right syntax and symbol, because we're starting from the _block_ rather than the declaration of a method.

This should be more performant because we're at the operation level rather than going from syntax to symbols, and we're doing the dataflow analysis once per method instead of once per readonly-parameter.

PR #945, opened 2024-01-18
Add analysis to ensure deserializability

alamarre:


src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/MustBeDeserializableAnalyzer.cs
```diff
@@ -0,0 +1,574 @@
...
+			}
+
+			SemanticModel? model = assignmentOperation.SemanticModel;
+			ITypeSymbol? type = assignmentOperation.Value.Type;
+
+
+			ISymbol? definitionSymbol = model?.GetSymbolInfo( assignmentOperation.Value.Syntax ).Symbol;
+			if( definitionSymbol == null ) {
+				return;
+			}
```
mthjones, line 103: Not sure I'm fully understanding this correctly, but is this equivalent to just calling `operation.Field` to get the field symbol?

src/D2L.CodeStyle.Analyzers/ApiUsage/Serialization/MustBeDeserializableAnalyzer.cs
mthjones, line 1: I noticed most of these actions start working with operations, but then drop down to syntax and semantic models. This is really inefficient (there's a reason all those SemanticModel methods take CancellationToken), and also makes for a lot more complex and fragile code. We should be able to do everything here with the existing operations, or by working with symbol actions. Let me know if you want to work through one together.

mthjones (changes requested):
Still working to grok the analyzers here, but as a general comment I'd like to see all of the syntax calls replaced. Syntax trees are the wrong abstraction level here which is leading to much more complex code to review and, in the future, extend and maintain. Syntax is also more fragile to language changes (e.g. when `=>` was added for property getters). Happy to work with you to figure out symbol/operation level replacements!

alamarre:
Moving to internal project

PR #947, opened 2024-02-13
Emit a nullable directive in generated sync code

j3parker:


src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -80,6 +80,9 @@ internal sealed class FileCollector {
 		}
 
 		public string CollectSource() {
+			// This allows us to copy+paste annotations but otherwise does not emit diagnostics
+			// otherwise we get "CS8669: The annotation for nullable reference types should only be used in code within a '#nullable' annotations context. Auto-generated code requires an explicit '#nullable' directive in source."
+			m_out.AppendLine( "#nullable annotations" );
```
j3parker, line 85: list of options: https://learn.microsoft.com/en-us/dotnet/csharp/nullable-references#nullable-contexts

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
```diff
@@ -82,6 +82,9 @@ internal sealed class FileCollector {
 		public string CollectSource() {
 			// TODO: Remove this and modify XML param elements in generator when changed/removed
 			m_out.AppendLine( "#pragma warning disable CS1572" );
+			// This allows us to copy+paste annotations but otherwise does not emit diagnostics
+			// otherwise we get "CS8669: The annotation for nullable reference types should only be used in code within a '#nullable' annotations context. Auto-generated code requires an explicit '#nullable' directive in source."
+			m_out.AppendLine( "#nullable annotations" );
```
mthjones, line 87: If a project specifies its own nullable context explicitly at the project level, how does this interact?

mthjones (approved):

src/D2L.CodeStyle.Analyzers/Async/Generator/FileCollector.cs
j3parker: This only applies to this file. It seems that we need to expliclitly set it per-file for generated code regardless, otherwise its equivalent to the status quo.

PR #951, opened 2024-07-31
Allow users to suppress the sync generator

j3parker:
This will allow .NET core projects to not generate sync stuff (which will often not want to support sync)

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
+++ b/src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
@@ -8,6 +8,9 @@ namespace D2L.CodeStyle.Analyzers.Async.Generator;
 [Generator]
 internal sealed partial class SyncGenerator : IIncrementalGenerator {
 	public void Initialize( IncrementalGeneratorInitializationContext context ) {
+		var options = context.AnalyzerConfigOptionsProvider
+			.Select( ParseConfig );
+
 		// Collect all the methods we want to generate sync versions for
 		// individually (for better incremental builds)
 		IncrementalValuesProvider<(MethodDeclarationSyntax, Compilation)> methodsToGenerate =
@@ -35,7 +38,7 @@ internal sealed partial class SyncGenerator : IIncrementalGenerator {
 			generatedMethods.Collect()
 				.SelectMany( GenerateFiles );
 
-		context.RegisterSourceOutput( generatedFiles, WriteFiles );
+		context.RegisterSourceOutput( generatedFiles.Combine( options ), WriteFiles );
 	}
 
 	private static bool IsInterestingLookingMethod(
@@ -167,8 +170,15 @@ internal sealed partial class SyncGenerator : IIncrementalGenerator {
 
 	private static void WriteFiles(
 		SourceProductionContext context,
-		FileGenerationResult result
+		(FileGenerationResult, SyncGeneratorOptions) args 
 	) {
+		var (result, options) = args;
+
+		// TODO: can we read options earlier and not spend time generating?
+		if( options.SuppressSyncGenerator ) {
+			return;
+		}
+
 		foreach( var diagnostic in result.Diagnostics ) {
 			context.ReportDiagnostic( diagnostic );
 		}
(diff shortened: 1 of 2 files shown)


j3parker:
The first place I'd like to use this is D2L.Security.OAuth2, to remove sync from the .NET Core build.

It's not so much about the implementation of the sync stuff, but the fact that the public interfaces gain sync methods and thus we're forcing users of the library to maintain sync implementations (which may not be feasible in .NET core)

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -167,8 +171,15 @@ CancellationToken cancellationToken
...
 		SourceProductionContext context,
-		FileGenerationResult result
+		(FileGenerationResult, SyncGeneratorOptions) args 
 	) {
+		var (result, options) = args;
+
+		// TODO: can we read options earlier and not spend time generating?
+		if( options.SuppressSyncGenerator ) {
+			return;
+		}
```
omsmith, line 181: I think the easiest spot you could do is `methodsToGenerate.Combine( options ).Select( GenerateSyncMethod )` and update `GenerateSyncMethod` to take `(MethodDeclarationSyntax MethodDeclaration, Compilation Compilation, SyncGeneratorOptions options)`?

j3parker: Ah whoops meant to deal with this before opening the PR.

Ya haha I was writing this in github.dev without IDE support and figured the exact signature for `GenerateSyncMethod` would be nasty so dodged it.

I also wanted to see if I could make it happen earlier, like before `IsInterestingLookingMethod`/`ExtractSyntax` etc. but even with the IDE it wasn't clear, oh well.

src/D2L.CodeStyle.Analyzers/Async/Generator/SyncGenerator.cs
```diff
@@ -20,7 +24,8 @@ internal sealed partial class SyncGenerator : IIncrementalGenerator {
 		// Do the generation per method
 		IncrementalValuesProvider<MethodGenerationResult> generatedMethods =
 			methodsToGenerate
-				.Select( GenerateSyncMethod )
+				.Combine( options )
+				.Select( (x, ct) => GenerateSyncMethod( x.Left.Item1, x.Left.Item2, x.Right, ct ) )
```
j3parker, line 28: `x.Left.Item1, x.Left.Item2, x.Right` pretty ugly lol

omsmith: Ah, I thought Combine magiced that. And you can't destructure into lambda variables yet so that's also sad

omsmith (approved):

src/D2L.CodeStyle.Analyzers/D2L.CodeStyle.Analyzers.csproj
```diff
@@ -46,4 +46,8 @@
     <PackageReference Include="System.Collections.Immutable" Version="7.0.0" PrivateAssets="All" />
   </ItemGroup>
 
+  <ItemGroup>
+    <CompilerVisibleProperty Include="SuppressSyncGenerator" />
+  </ItemGroup>
```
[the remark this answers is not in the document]
omsmith: Should we prefix that with `D2LCodeStyle` or something then?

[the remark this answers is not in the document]
j3parker: 🤷‍♂️ I've seen microsoft not prefix as long as its unambiguous, but I have no strong feelings.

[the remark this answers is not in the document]
j3parker: Just noticed they do have a `MyGenerator_` prefix in these docs (which I just found): https://github.com/dotnet/roslyn/blob/2946c619d91c7279d69d71bbb521ea4647e29fc1/docs/features/source-generators.cookbook.md#consume-msbuild-properties-and-metadata

PR #954, opened 2024-08-07
Enable nullable everywhere

j3parker:
I'd like to add more csproj's soon, and to have nullable enabled in those places, so for simplicity I'm enabling it globally first.

### DIFF
--- a/src/D2L.CodeStyle.TestAnalyzers/Common/NotNullWhenAttribute.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/Common/NotNullWhenAttribute.cs
@@ -1,0 +1,14 @@
+﻿﻿#if !NETSTANDARD2_1
+
+namespace System.Diagnostics.CodeAnalysis;
+
+[AttributeUsage( AttributeTargets.Parameter )]
+internal sealed class NotNullWhenAttribute : Attribute {
+	public bool ReturnValue { get; }
+
+	public NotNullWhenAttribute( bool returnValue ) {
+		ReturnValue = returnValue;
+	}
+}
+
+#endif
--- a/src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
 using System.Text;
 using Microsoft.CodeAnalysis;
 
@@ -9,7 +10,9 @@ namespace D2L.CodeStyle.TestAnalyzers.Extensions {
 
 		// Copied from the non-test assembly because we do not reference it.
 
-		public static bool IsNullOrErrorType( this ITypeSymbol? symbol ) {
+		public static bool IsNullOrErrorType(
+			[NotNullWhen( false )] this ITypeSymbol? symbol
+		) {
 
 			if( symbol == null ) {
 				return true;
@@ -26,20 +29,9 @@ namespace D2L.CodeStyle.TestAnalyzers.Extensions {
 			return false;
 		}
 
-		public static bool IsErrorType( this ITypeSymbol symbol ) {
-
-			if( symbol.Kind == SymbolKind.ErrorType ) {
-				return true;
-			}
-
-			if( symbol.TypeKind == TypeKind.Error ) {
-				return true;
-			}
-
-			return false;
-		}
-
-		public static bool IsNullOrErrorType( this ISymbol symbol ) {
+		public static bool IsNullOrErrorType(
+			[NotNullWhen( false )] this ISymbol? symbol
+		) {
 
 			if( symbol == null ) {
 				return true;
--- a/src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
@@ -9,11 +9,24 @@ namespace D2L.CodeStyle.TestAnalyzers.Extensions {
 
 		// Copied from the non-test assembly because we do not reference it.
 
-		public static bool IsNullOrErrorType( this ITypeSymbol symbol ) {
+		public static bool IsNullOrErrorType( this ITypeSymbol? symbol ) {
 
 			if( symbol == null ) {
 				return true;
 			}
+
+			if( symbol.Kind == SymbolKind.ErrorType ) {
+				return true;
+			}
+
+			if( symbol.TypeKind == TypeKind.Error ) {
+				return true;
+			}
+
+			return false;
+		}
+
+		public static bool IsErrorType( this ITypeSymbol symbol ) {
 
 			if( symbol.Kind == SymbolKind.ErrorType ) {
 				return true;
--- a/src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
@@ -54,7 +54,7 @@ namespace D2L.CodeStyle.TestAnalyzers.NUnit {
 					context: ctx,
 					bannedCategories: bannedCategories,
 					types: types,
-					syntax: (ctx.Node as MethodDeclarationSyntax)!
+					syntax: (MethodDeclarationSyntax)ctx.Node
 				),
 				SyntaxKind.MethodDeclaration
 			);
--- a/src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
+++ b/src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using D2L.CodeStyle.TestAnalyzers.Common;
@@ -42,7 +43,7 @@ namespace D2L.CodeStyle.TestAnalyzers.NUnit {
 		}
 
 		private static void OnCompilationStart( CompilationStartAnalysisContext context ) {
-			if( !TryLoadNUnitTypes( context.Compilation, out NUnitTypes types ) ) {
+			if( !TryLoadNUnitTypes( context.Compilation, out NUnitTypes? types ) ) {
 				return;
 			}
 
@@ -53,7 +54,7 @@ namespace D2L.CodeStyle.TestAnalyzers.NUnit {
 					context: ctx,
 					bannedCategories: bannedCategories,
 					types: types,
-					syntax: ctx.Node as MethodDeclarationSyntax
+					syntax: (ctx.Node as MethodDeclarationSyntax)!
 				),
 				SyntaxKind.MethodDeclaration
 			);
@@ -69,9 +70,16 @@ namespace D2L.CodeStyle.TestAnalyzers.NUnit {
 					return;
 				}
 
+				var syntaxReference = attribute.ApplicationSyntaxReference;
+
+				if( syntaxReference == null ) {
+					// ??
+					return;
+				}
+
 				context.ReportDiagnostic( Diagnostic.Create(
 					Diagnostics.NUnitCategory,
-					attribute.ApplicationSyntaxReference.GetSyntax( context.CancellationToken ).GetLocation(),
+					syntaxReference.GetSyntax( context.CancellationToken ).GetLocation(),
 					$"Assemblies cannot be categorized as any of [{string.Join( ", ", ProhibitedAssemblyCategories )}], but saw '{category}'."
 				) );
 			} );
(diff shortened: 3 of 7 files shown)


src/D2L.CodeStyle.TestAnalyzers/Common/NotNullWhenAttribute.cs
```diff
@@ -0,0 +1,14 @@
...
+[AttributeUsage( AttributeTargets.Parameter )]
+internal sealed class NotNullWhenAttribute : Attribute {
+	public bool ReturnValue { get; }
+
+	public NotNullWhenAttribute( bool returnValue ) {
+		ReturnValue = returnValue;
+	}
+}
+
+#endif
```
j3parker, line 14: The other project brings in a package for this but I'm pretty meh on that.

In the future I'm going to make a folder for common source for analyzers and some Directory.Build.targets magic to `<Compile Include="..." LinkBase="Common" />`-in some shared analyzer code and this can go in there.

src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -26,6 +26,19 @@ public static class RoslynExtensions {
 			return false;
 		}
 
+		public static bool IsErrorType( this ITypeSymbol symbol ) {
```
j3parker, line 29: In some places I lifted the null check out because it was convenient.
For ISymbol I just replaced any problematic usages with the inline checks. I kept this one because there is an extra check (Kind and TypeKind).

I don't really like these overloads much, it seems pretty imprecise, but I didn't want to look too closely to see if all the cases were necessary.

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -69,9 +70,16 @@ internal sealed class CategoryAnalyzer : DiagnosticAnalyzer {
 					return;
 				}
 
+				var syntaxReference = attribute.ApplicationSyntaxReference;
+
+				if( syntaxReference == null ) {
+					// ??
```
j3parker, line 76: I don't know if this is possible but also don't care much. I changed the behaviour to just silently continue if it happens

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -268,15 +291,19 @@ AnalyzerOptions options
...
 
 			if( bannedListFile == null ) {
 				return bannedList.ToImmutableHashSet();
 			}
 
-			SourceText allowedListText = bannedListFile.GetText();
+			SourceText? allowedListText = bannedListFile.GetText();
+
+			if( allowedListText == null ) {
+				throw new Exception( "Couldn't read config" );
```
j3parker, line 305: I'm not sure how this would happen so I didn't try hard on this exception; I figure its better than the NullReferenceException we'd otherwise have. The IDE/CLI will tell you which analyzer threw an exception so I figure this is enough hint for now.

src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
```diff
@@ -9,7 +9,7 @@ public static class RoslynExtensions {
 
 		// Copied from the non-test assembly because we do not reference it.
 
-		public static bool IsNullOrErrorType( this ITypeSymbol symbol ) {
+		public static bool IsNullOrErrorType( this ITypeSymbol? symbol ) {
```
mthjones, line 12: nit: could this be [annotated with `[NotNullWhen( false )]`](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis#conditional-post-conditions-notnullwhen-maybenullwhen-and-notnullifnotnull) to avoid needing the new method, or does that not work with extension methods?

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -53,7 +54,7 @@ internal sealed class CategoryAnalyzer : DiagnosticAnalyzer {
 					context: ctx,
 					bannedCategories: bannedCategories,
 					types: types,
-					syntax: ctx.Node as MethodDeclarationSyntax
+					syntax: (ctx.Node as MethodDeclarationSyntax)!
```
mthjones, line 57: nit: `SyntaxNodeAnalysisContext.Node` is already non-nullable. If this were a hard-cast would we be able to avoid needing to force non-null?

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -69,9 +70,16 @@ internal sealed class CategoryAnalyzer : DiagnosticAnalyzer {
 					return;
 				}
 
+				var syntaxReference = attribute.ApplicationSyntaxReference;
+
+				if( syntaxReference == null ) {
+					// ??
+					return;
+				}
```
mthjones, line 78: Since this already would have been throwing an NRE in this case anyways, I think raising an exception or forcing non-null would be fine here. This property is undocumented, so can't really tell when it would ever actually _be_ null. But better to loudly fail if those cases ever come up than silently ignore and potentially let stuff slip though.

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -189,11 +197,16 @@ ISymbol symbol
...
-			if( arg.Type.SpecialType != SpecialType.System_String ) {
+			if( arg.Type?.SpecialType != SpecialType.System_String ) {
+				return;
+			}
+
+			string? category = arg.Value as string;
+
+			if( category == null ) {
 				return;
 			}
```
mthjones, line 208: If Roslyn says it's a string, but it's somehow not actually a string, I feel like that's worth an exception.

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -208,11 +221,16 @@ ISymbol symbol
...
-				if( arg.Type.SpecialType != SpecialType.System_String ) {
+				if( arg.Type?.SpecialType != SpecialType.System_String ) {
+					continue;
+				}
+
+				string? categoryCsv = arg.Value as string;
+
+				if( categoryCsv == null ) {
 					continue;
 				}
```
mthjones, line 232: Likewise.

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -221,24 +239,29 @@ ISymbol symbol
...
-					compilation.GetTypeByMetadataName( "NUnit.Framework.TestCaseAttribute" ),
-					compilation.GetTypeByMetadataName( "NUnit.Framework.TestCaseSourceAttribute" ),
-					compilation.GetTypeByMetadataName( "NUnit.Framework.TheoryAttribute" )
-				);
+			ImmutableHashSet<INamedTypeSymbol> testAttributes = new[] {
+				compilation.GetTypeByMetadataName( "NUnit.Framework.TestAttribute" ),
+				compilation.GetTypeByMetadataName( "NUnit.Framework.TestCaseAttribute" ),
+				compilation.GetTypeByMetadataName( "NUnit.Framework.TestCaseSourceAttribute" ),
+				compilation.GetTypeByMetadataName( "NUnit.Framework.TheoryAttribute" )
+			}.Where( x => x != null )!
```
mthjones, line 256: Is the null-forgiving op necessary on the result of `Where`?

src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
```diff
@@ -73,12 +73,12 @@ INamedTypeSymbol attributeType
...
 			foreach( AttributeSyntax attribute in attributes ) {
 
-				ISymbol symbol = context
+				ISymbol? symbol = context
 					.SemanticModel
 					.GetSymbolInfo( attribute, context.CancellationToken )
 					.Symbol;
 
-				if( symbol.IsNullOrErrorType() ) {
+				if( symbol == null || symbol.Kind == SymbolKind.ErrorType ) {
```
mthjones, line 81: Was losing the `TypeKind` check intentional here?

src/D2L.CodeStyle.TestAnalyzers/Extensions/Microsoft.CodeAnalysis.cs
j3parker: Oh interesting, I didn't realize it worked with non-`out` parameters (and indeed this case is right there in the docs)
I'll revisit it in this PR yep

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
j3parker: Ah yeah sure. Technically a hard cast throws a different exception/earlier but the reason I used `!` is that in the context the cast will always succeed, so yeah I'll switch to hard cast

src/D2L.CodeStyle.TestAnalyzers/NUnit/ConfigTestSetupStringsAnalyzer.cs
j3parker: 👍 I'm going to switch it back to the extension anyway for simplicity (I still dislike them, kind of) using `[NotNullWhen( false )]` as discussed above.

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
j3parker: Yep, well known problem.

They've talked about extending nullable analysis more to handle LINQ stuff but nothing very concrete. Also related to covariance headaches (e.g. `Task<T>` vs `Task<T?>`)

The `Where( x => x is not null )!` thing is pretty common from what I've seen. Sometimes people make their own extensions. I've also seen `.OfType<T>()` but that feels even worse to me (not obvious what kind of "casting" is happening there)

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -69,9 +70,12 @@ internal sealed class CategoryAnalyzer : DiagnosticAnalyzer {
 					return;
 				}
 
+				// TODO: can this actually be null?
+				var syntaxReference = attribute.ApplicationSyntaxReference!;
```
j3parker, line 74: comment is just there as a hint that this `!` is a guess.
Clearly it hasn't caused troubles yet 🤷

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -189,11 +193,16 @@ ISymbol symbol
 
 			TypedConstant arg = args[0];
 
+			if( arg.Type == null ) {
+				return;
+			}
```
j3parker, line 198: Behaviour change that I think is reasonable (ditto below)?

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -189,11 +193,16 @@ ISymbol symbol
...
+			if( arg.Type == null ) {
+				return;
+			}
+
 			if( arg.Type.SpecialType != SpecialType.System_String ) {
 				return;
 			}
 
-			string category = arg.Value as string;
+			string category = (string?)arg.Value!;
```
j3parker, line 204: We know that arg.Type is System.String. I'm not sure if `null` can actually be returned here, I somewhat doubt it -- but this is equivalent behaviour fwiw

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -221,24 +235,29 @@ ISymbol symbol
...
+			}.Where( x => x != null )!
+			.ToImmutableHashSet<INamedTypeSymbol>( SymbolEqualityComparer.Default );
 
-			INamedTypeSymbol testFixtureAttribute = compilation.GetTypeByMetadataName( "NUnit.Framework.TestFixtureAttribute" );
+			INamedTypeSymbol? testFixtureAttribute = compilation.GetTypeByMetadataName( "NUnit.Framework.TestFixtureAttribute" );
+
+			if( testFixtureAttribute == null ) {
+				types = null;
+				return false;
+			}
```
j3parker, line 260: (reasonable imo) behaviour change

j3parker: (clearly its not crashing today, so maybe in practice not, but I suppose NUnit could rename / move these types in the future.)

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
mthjones: Does `[TestCategory( null )]` bail out on an earlier check? I guess it's treated as `object` because the constant is evaluated independently of the context it's used in?

mthjones (approved):

src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
j3parker: I'm not sure

Issue #956, opened 2024-08-08
[ConditionallyImmutable] is unsound


j3parker:
```cs
[ConditionallyImmutable]
interface IOne<[ConditionallyImmutable.OnlyIf]A> {}

[ConditionallyImmutable]
interface ITwo<[ConditionallyImmutable.OnlyIf]A, [ConditionallyImmutable.OnlyIf]B> : IOne<A>{}

record Impl(int X, List<int> Y) : ITwo<int, List<int>>;
```

An `IOne<A>` is immutable when its type argument `A` is immutable (otherwise it is not).
An `ITwo<A, B>` likewise is immutable when both of its type arguments are immutable.

However, `ITwo<A, B>` also extends `IOne<A>` which will cause a problem:

`Impl` implements `ITwo<int, List<int>>`. It is never judged (itself) to be immutable, because it's Y property is a mutable `List<int>`.

If you refer to it as an `ITwo<int, List<int>>` it will not be considered immutable due to the `OnlyIf` on `ITwo`'s `B` type parameter.

However, every `ITwo<int, List<int>>` is an `IOne<int>`, which is always considered to be immutable, which is problematic.

We could detect cases from one immutable type to another to cover many of the consequences of this, but I don't think that would be sufficient. The smuggling could happen outside of code we are able to analyze (e.g. in the .NET standard library), e.g. via covariance on wrapper classes (`ISomeInnocentThing<out T>` being able to turn an `ISomeInnocentThing<ITwo<int, List<int>>>` into an `ISomeInnocentThing<IOne<int>>`) or even worse a method (e.g. `TReturn SomeInnocentThing<TInput, TReturn>( TInput input )` internally having casting).

Prior to `ConditionallyImmutable` we have accepted that we can't scan all code (e.g. the standard library) but we still want reasonable guarantees -- so I worry that hunting for cases like casts takes us far away from that.

One idea is to ban `ITwo<[OnlyIf]A, [OnlyIf]B>` from implementing `IOne<[OnlyIf]A>` on the rationale that it is meaningfully "less immutable". It's immutability is qualified on things that `IOne<A>`'s isn't, and so it makes sense to complain. The exact rules may need some thinking. To make a more concrete and slightly more complicated case:

```cs
[ConditionallyImmutable]
interface IReadOnlyCollection<[ConditionallyImmutable.OnlyIf] T> {}

[ConditionallyImmutable]
[...]

PR #958, opened 2024-08-09
immutability fixes

omsmith:
* require application of all [ConditionallyImmutable.OnlyIf] type paramters
* a [ConditionallyImmutable] type should not be able to extend/implement an [Immutable] one

Fixes: https://github.com/Brightspace/D2L.CodeStyle/issues/956

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
@@ -94,8 +94,38 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 					return;
 
 				case ImmutableTypeKind.Total:
+					bool missingConditionalParameterUsage = false;
 					if( typeInfo.Kind == ImmutableTypeKind.Total ) {
-						break;
+						if( !typeInfo.IsConditional ) {
+							break;
+						}
+
+						if( baseTypeInfo.IsConditional ) {
+							foreach( ITypeParameterSymbol typeParameter in typeInfo.ConditionalTypeParameters ) {
+								bool parameterUsed = false;
+								for( int i = 0; i < baseTypeInfo.Type.TypeArguments.Length && !parameterUsed; i++ ) {
+									ITypeSymbol typeArgument = baseTypeInfo.Type.TypeArguments[ i ];
+									if( !SymbolEqualityComparer.Default.Equals( typeParameter, typeArgument ) ) {
+										continue;
+									}
+
+									ITypeParameterSymbol baseTypeParameter = baseTypeInfo.Type.TypeParameters[ i ];
+									if( baseTypeInfo.ConditionalTypeParameters.Contains( baseTypeParameter, SymbolEqualityComparer.Default ) ) {
+										parameterUsed = true;
+										break;
+									}
+								}
+
+								if( !parameterUsed ) {
+									missingConditionalParameterUsage = true;
+									break;
+								}
+							}
+
+							if( !missingConditionalParameterUsage ) {
+								break;
+							}
+						}
 					}
 
 					// The docs say the only things that can have BaseType == null
@@ -116,17 +146,29 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 						cancellationToken
 					);
 
-					m_diagnosticSink(
-						Diagnostic.Create(
-							Diagnostics.MissingTransitiveImmutableAttribute,
-							syntax.Identifier.GetLocation(),
-							properties: FixArgs,
-							typeInfo.Type.GetFullTypeName(),
-							baseTypeInfo.IsConditional ? " (or [ConditionallyImmutable])" : "",
-							isInterface ? "interface" : "base class",
-							baseTypeInfo.Type.GetFullTypeName()
-						)
-					);
+					if( missingConditionalParameterUsage ) {
+						m_diagnosticSink(
+							Diagnostic.Create(
+								Diagnostics.UnappliedConditionalImmutability,
+								syntax.Identifier.GetLocation(),
+								typeInfo.Type.GetFullTypeName(),
+								isInterface ? "interface" : "base class",
+								baseTypeInfo.Type.GetFullTypeName()
+							)
+						);
+					} else {
+						m_diagnosticSink(
+							Diagnostic.Create(
+								Diagnostics.MissingTransitiveImmutableAttribute,
+								syntax.Identifier.GetLocation(),
+								properties: FixArgs,
+								typeInfo.Type.GetFullTypeName(),
+								baseTypeInfo.IsConditional ? " (or [ConditionallyImmutable])" : "",
+								isInterface ? "interface" : "base class",
+								baseTypeInfo.Type.GetFullTypeName()
+							)
+						);
+					}
 
 					break;
 
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
@@ -92,13 +92,11 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 
 				// If the base type doesn't require us to be immutable then there is nothing to check
 				case ((ImmutableTypeKind.None, _), _):
-
 				case ((ImmutableTypeKind.Instance, _), _):
 					return;
 
 				// When the base type is [Immutable] or [ConditionallyImmutable] we need to be as well
 				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.None, _)):
-
 				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.Instance, _)):
 					RaiseMissingAttribute();
 					return;
--- a/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
+++ b/src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
@@ -90,11 +90,15 @@ namespace D2L.CodeStyle.Analyzers.Immutability {
 				default:
 					throw new NotImplementedException();
 
+				// If the base type doesn't require us to be immutable then there is nothing to check
 				case ((ImmutableTypeKind.None, _), _):
+
 				case ((ImmutableTypeKind.Instance, _), _):
 					return;
 
+				// When the base type is [Immutable] or [ConditionallyImmutable] we need to be as well
 				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.None, _)):
+
 				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.Instance, _)):
 					RaiseMissingAttribute();
 					return;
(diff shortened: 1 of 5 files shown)


src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -94,8 +94,38 @@ CancellationToken cancellationToken
...
+							break;
+						}
+
+						if( baseTypeInfo.IsConditional ) {
+							foreach( ITypeParameterSymbol typeParameter in typeInfo.ConditionalTypeParameters ) {
+								bool parameterUsed = false;
+								for( int i = 0; i < baseTypeInfo.Type.TypeArguments.Length && !parameterUsed; i++ ) {
+									ITypeSymbol typeArgument = baseTypeInfo.Type.TypeArguments[ i ];
+									if( !SymbolEqualityComparer.Default.Equals( typeParameter, typeArgument ) ) {
+										continue;
```
j3parker, line 109: So this would emit a diagnostic for `IMyReadOnlyDictionary` right?

```cs
[ConditionallyImmutable]
MyKeyValuePair<[ConditionallyImmutable.ConditionallyImmutable OnlyIf]TKey, [ConditionallyImmutable.OnlyIf]TValue>

[ConditionallyImmutable]
IMyReadOnlyCollection<[ConditionallyImmutable.ConditionallyImmutable OnlyIf]T>

[ConditionallyImmutable]
IMyReadOnlyDictionary<[ConditionallyImmutable.ConditionallyImmutable OnlyIf]TKey, [ConditionallyImmutable.OnlyIf]TValue> : IMyReadOnlyCollection<KeyValuePair<TKey, TValue>>
```

because `TKey` "isn't used" in the type arguments to `IMyReadOnlyCollection`, right?

omsmith: Yeah, it would

omsmith: You thinking a recursive search?

j3parker: Ok -- you agree that the above is safe right?

Not saying its a problem with this PR, just wrapping my head around the issue

omsmith: > Ok -- you agree that the above is safe/we would like it to be accepted right?

Yep

omsmith:
Hmm, something is broken
image

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
j3parker: > You thinking a recursive search?

Something like that but I think it's a maybe a little fancy.

I think what we want to know is, are we (`IMyReadOnlyDictionary`) "at least" as immutable as all of the interfaces (base class) we implement (derive from)? "At least" meaning that any time the broader type gets judged as immutable we also do...

I think the reason its safe here is `IMyReadOnlyCollection<T>` is only immutable when T is, ok so recursively we need to care about when KeyValuePair<TKey, TValue> is... again recursively we look at that and see we have conditionals on TKey and TValue....

If we accumulate all those conditionals then I think we want to say OK, are the set of conditionals from my equal to (or a superset of?) the conditionals from my type parameters?

I'm not quite sure though, still thinking about it.

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -87,54 +87,103 @@ CancellationToken cancellationToken
...
+							if( !SymbolEqualityComparer.Default.Equals( typeParameter, typeArgument ) ) {
+								continue;
+							}
+
+							ITypeParameterSymbol baseTypeParameter = baseTypeInfo.Type.TypeParameters[ i ];
+							if( baseTypeInfo.ConditionalTypeParameters.Contains( baseTypeParameter, SymbolEqualityComparer.Default ) ) {
+								parameterUsed = true;
+								break;
+							}
+						}
```
j3parker, line 129: We talked on Zoom about how this chunk could be a helper and that helper would change if we implemented the more powerful logic here (recursive/whatever)

j3parker (approved):
We talked on zoom about reorganizing this a bit (to cut down on state/nesting) but didn't get anywhere fantastic
EDIT: we're looking at a big tuple switch and that looks promising; if it pans out this approval still stands

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -86,52 +86,88 @@ CancellationToken cancellationToken
 			ImmutableTypeInfo baseTypeInfo,
 			CancellationToken cancellationToken
 		) {
-			switch( baseTypeInfo.Kind ) {
-				case ImmutableTypeKind.None:
+			switch( ((baseTypeInfo.Kind, baseTypeInfo.IsConditional), (typeInfo.Kind, typeInfo.IsConditional)) ) {
+				default:
+					throw new NotImplementedException();
+
+				case ((ImmutableTypeKind.None, _), _):
```
j3parker, line 93:
```suggestion
				// If the base type doesn't require us to be immutable then there is nothing to check
				case ((ImmutableTypeKind.None, _), _):

```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -86,52 +86,88 @@ CancellationToken cancellationToken
...
+			switch( ((baseTypeInfo.Kind, baseTypeInfo.IsConditional), (typeInfo.Kind, typeInfo.IsConditional)) ) {
+				default:
+					throw new NotImplementedException();
+
+				case ((ImmutableTypeKind.None, _), _):
+				case ((ImmutableTypeKind.Instance, _), _):
 					return;
 
-				case ImmutableTypeKind.Instance:
+				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.None, _)):
```
j3parker, line 97:
```suggestion
				// When the base type is [Immutable] or [ConditionallyImmutable] we need to be as well
				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.None, _)):

```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -86,52 +86,88 @@ CancellationToken cancellationToken
...
-
-					break;
+				// If the base type is [Immutable] then the implementing type should be as well
+				case ((ImmutableTypeKind.Total, false), (ImmutableTypeKind.Total, true)):
+					RaiseMissingAttribute();
+					return;
 
-				default:
-					throw new NotImplementedException();
+				case ((ImmutableTypeKind.Total, true), (ImmutableTypeKind.Total, true)):
```
j3parker, line 112:
```suggestion
				// When we and our base type are [ConditionallyImmutable] we need to consider the compatibility of those conditions
				case ((ImmutableTypeKind.Total, true), (ImmutableTypeKind.Total, true)):
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -86,52 +86,88 @@ CancellationToken cancellationToken
...
-							baseTypeInfo.IsConditional ? " (or [ConditionallyImmutable])" : "",
-							isInterface ? "interface" : "base class",
-							baseTypeInfo.Type.GetFullTypeName()
-						)
-					);
-
-					break;
+				// If the base type is [Immutable] then the implementing type should be as well
+				case ((ImmutableTypeKind.Total, false), (ImmutableTypeKind.Total, true)):
+					RaiseMissingAttribute();
```
j3parker, line 113:
```suggestion
					// TODO: error message could be improved to directly explain that we need [Immutable]
					RaiseMissingAttribute();
```

j3parker: I don't think it's urgent to actually implement but I think it'd be helpful to write down what we talked about in a comment

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -86,52 +86,88 @@ CancellationToken cancellationToken
...
+						m_diagnosticSink(
+							Diagnostic.Create(
+								Diagnostics.UnappliedConditionalImmutability,
+								syntax.Identifier.GetLocation(),
+								typeInfo.Type.GetFullTypeName(),
+								baseTypeInfo.Type.TypeKind == TypeKind.Interface ? "interface" : "base class",
+								baseTypeInfo.Type.GetFullTypeName()
+							)
+						);
+						return;
```
j3parker, line 173: We talked about how we could drop this return if the diagnostic was on `typeParameter` instead, like we could emit multiple diagnostics instead. I dno, consider it since this is a new diagnostic anyway

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -86,52 +86,93 @@ CancellationToken cancellationToken
...
-			switch( baseTypeInfo.Kind ) {
-				case ImmutableTypeKind.None:
+			switch( ((baseTypeInfo.Kind, baseTypeInfo.IsConditional), (typeInfo.Kind, typeInfo.IsConditional)) ) {
+				default:
+					throw new NotImplementedException();
+
+				// If the base type doesn't require us to be immutable then there is nothing to check
+				case ((ImmutableTypeKind.None, _), _):
+
+				case ((ImmutableTypeKind.Instance, _), _):
```
omsmith, line 96:
```suggestion
				// If the base type doesn't require us to be immutable then there is nothing to check
				case ((ImmutableTypeKind.None, _), _):
				case ((ImmutableTypeKind.Instance, _), _):
```

src/D2L.CodeStyle.Analyzers/Immutability/ImmutableAttributeConsistencyChecker.cs
```diff
@@ -86,52 +86,93 @@ CancellationToken cancellationToken
...
 					return;
 
-				case ImmutableTypeKind.Instance:
+				// When the base type is [Immutable] or [ConditionallyImmutable] we need to be as well
+				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.None, _)):
+
+				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.Instance, _)):
```
omsmith, line 102:
```suggestion
				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.None, _)):
				case ((ImmutableTypeKind.Total, _), (ImmutableTypeKind.Instance, _)):
```

j3parker (approved):

PR #961, opened 2024-08-10
RequireNamedArgumentsAnalyzer: add support for exemptions

omsmith:
ExemptSymbolsBuilder is a helper for building sets of exempt symbols without bespoke logic per-analyzer.

The AdditionalFile format is similar to Roslyn's BannedApiAnalyzer: https://github.com/dotnet/roslyn-analyzers/blob/3211f48253bc18560156d90dc5e710d35f7d03fa/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/BannedApiAnalyzers.Help.md

In this case the optional `[;Description Text]` is excluded from the format, as exemptions do not lead to user-facing messages.

Comments are supported for developers to leave notes justifying the exemption.

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/Helpers/ExemptSymbolsBuilder.cs
+++ b/src/D2L.CodeStyle.Analyzers/Helpers/ExemptSymbolsBuilder.cs
@@ -1,0 +1,84 @@
+﻿using System.Collections.Immutable;
+using Microsoft.CodeAnalysis;
+using Microsoft.CodeAnalysis.Diagnostics;
+using Microsoft.CodeAnalysis.Text;
+
+namespace D2L.CodeStyle.Analyzers.Helpers;
+
+internal sealed class ExemptSymbolsBuilder {
+
+	private readonly Compilation m_compilation;
+	private readonly AnalyzerOptions m_analyzerOptions;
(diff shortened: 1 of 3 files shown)


src/D2L.CodeStyle.Analyzers/Helpers/ExemptSymbolsBuilder.cs
```diff
@@ -0,0 +1,84 @@
...
+	)]
+	public ExemptSymbolsBuilder(
+		CompilationStartAnalysisContext context
+	) {
+		m_compilation = context.Compilation;
+		m_analyzerOptions = context.Options;
+		m_cancellationToken = context.CancellationToken;
+	}
+
+	public ImmutableHashSet<ISymbol> Build() => m_exemptions.ToImmutable();
```
j3parker, line 29: `IReadOnlySet<T>`/`HashSet<T>` is better performance

j3parker: (`ImmutableHashSet<T>` uses an implementation that makes functional edits cheaper, at the cost of making lookups slower.)

j3parker (approved):

PR #963, opened 2025-05-28
LFT-1728 - Add support for CollectionExpression in async generator

j3parker:


.github/workflows/build.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - uses: Brightspace/third-party-actions@actions/setup-dotnet
         with:
           dotnet-version: |
-            7.0.x
+            9.0.x
```
j3parker, line 20: Need to update for new C# 12 syntax

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -352,6 +355,17 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
...
 
+	private SeparatedSyntaxList<CollectionElementSyntax> Transform(
+		SeparatedSyntaxList<CollectionElementSyntax> original
+	) => SyntaxFactory.SeparatedList(
+		original.Select( elem => elem switch {
+			ExpressionElementSyntax ee => ee.WithExpression( Transform( ee.Expression ) ),
+			SpreadElementSyntax sp => sp.WithExpression( Transform( sp.Expression ) ),
+			_ => UnhandledSyntax( elem )
+		} ),
+		original.GetSeparators()
```
omsmith, line 366: We're not passing the original separators for the other cases in the transformer, so perhaps something to correct later

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -352,6 +355,17 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
...
 		};
 
+	private SeparatedSyntaxList<CollectionElementSyntax> Transform(
+		SeparatedSyntaxList<CollectionElementSyntax> original
+	) => SyntaxFactory.SeparatedList(
+		original.Select( elem => elem switch {
+			ExpressionElementSyntax ee => ee.WithExpression( Transform( ee.Expression ) ),
+			SpreadElementSyntax sp => sp.WithExpression( Transform( sp.Expression ) ),
+			_ => UnhandledSyntax( elem )
+		} ),
```
omsmith, line 365: Particular reason this isn't just `original.Select( Transform )` or something, with the top-level Dispatch handling Spread?

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
```diff
@@ -352,6 +355,17 @@ private VariableDesignationSyntax Transform( VariableDesignationSyntax des )
...
+		original.GetSeparators()
+	);
```
j3parker, line 367: Actually part of this is already in place with the TransformAll combinator.
It doesn't do original.GetSeparators(), though.

I'll do a follow-up PR to switch to that, but also add the separators. That will incur some noise because we will have to fix up the whitespace in unrelated tests.

src/D2L.CodeStyle.Analyzers/Async/Generator/AsyncToSyncMethodTransformer.cs
j3parker: Yeah -- see note below.

One other reason we're not doing it in the other ones is TransformAll supports the "maybe transform" e.g. so we can delete cancellation token arguments. There needs to be some accounting for that otherwise .GetSeparators() will possibly insert extra separators/commas.

PR #974, opened 2026-05-15
fix ConstantAttributeAnalyzer null reference

omsmith:
Fixes a NullReferenceException during analysis of an InterpolatedStringHandler usage.

Possible to add a test, but requires a couple of package updates that I don't want to verify are safe, at the moment.

```csharp
[InterpolatedStringHandler]
public sealed class MyString {
  public MyString( int a, int b ) {}
  public void AppendLiteral( [Constant] string s ) {}
  public void AppendFormatted( [Constant] string s ) {}
}

public sealed class C {
  public void M( bool x ) {

    R(
      $"""
      FOO
      {( x ? "x" : "y" )}
      """
    );
  }

  public void R( MyString s ) {}
}
```

This desugars to

```csharp
  public void M( bool x ) {
    MyString <> = new MyString( 4, 1 );
    <>.AppendLiteral( "FOO\n" );
    <>.AppendFormatted( x ? "x" : "y" );
    R( <> );
  }
```

However, the syntax for the operation of `<>.AppendFormatted( x ? "x" : "y" );` is still from the original interpolated string; both `argument.Syntax` and `argument.Value.Syntax`  are `ConditionalExpressionSyntax` rather than an `ArgumentSyntax`.

SPLA-3693

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/ConstantAttributeAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/ConstantAttributeAnalyzer.cs
@@ -133,7 +133,7 @@ namespace D2L.CodeStyle.Analyzers.ApiUsage {
 			}
 
 			// Argument was defined as [Constant] already, so trust it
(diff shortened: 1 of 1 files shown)


j3parker (approved):
lol nice

PR #976, opened 2026-06-02
Add test case for class primary constructors

j3parker:
These create mutable fields, so this test should fail but maybe doesn't


j3parker:
#977

PR #983, opened 2026-07-17
LFT-3078 - Apply changes to remove analyzer version mismatch warning

toddlang:


src/D2L.CodeStyle.TestAnalyzers/NUnit/CategoryAnalyzer.cs
```diff
@@ -298,7 +298,7 @@ AnalyzerOptions options
 			SourceText? allowedListText = bannedListFile.GetText();
 
 			if( allowedListText == null ) {
-				throw new Exception( "Couldn't read config" );
+				throw new InvalidOperationException( "Couldn't read config" );
```
toddlang, line 301: So, getting the version mismatch fixed seemed to unlock CA2201 "Exception type System.Exception is not sufficiently specific", that's why this change is in here as well.  It's a straight-up error with these settings and could not be ignored.

omsmith (approved):

PR #989, opened 2026-09-09
move [PatternString] from the LMS

omsmith:
Brings in the `[PatternString]` attribute from the LMS (so it can be used outside the LMS), with edits to the analysis being done.

* Only only application to parameters, thus only analyze arguments (and conversions which are implicit arguments. This works more similarly to `[Constant]`, where holding an analyzed value in a property/field can be done via a wrapper struct. The previous analysis was quite heavy (top 3 analyzers) even without any usage of the attribute.
* Move diagnostic reporting of "invalid regex" to the attribute usage, vs consumers. Previously a bad regex wouldn't be reported unless and API was used, and you would get a diagnostic per usage (including in projects not defining the API).
* Add other familiar checks such as preventing MethodReferences (where we'd lose knowledge of the attribute, and wouldn't be able to enforce the regex), and enforcing consistent application between interface/base class implementation/override.

SPLA-4124

### DIFF
--- a/src/D2L.CodeStyle.Analyzers/ApiUsage/ConsistentParameterAttributesAnalyzer.cs
+++ b/src/D2L.CodeStyle.Analyzers/ApiUsage/ConsistentParameterAttributesAnalyzer.cs
@@ -60,20 +60,26 @@ namespace D2L.CodeStyle.Analyzers.ApiUsage {
 					var implementedParameterUsage = consistentAttributesContext.GetAttributeUsage( implementedMethod.Parameters[ i ] );
 
 					for( int j = 0; j < thisParameterUsage.Length; ++j ) {
-						var thisParameterUsageAttr = thisParameterUsage[ j ];
-						var implementedParameterUsageAttr = implementedParameterUsage[ j ];
+						(string AttributeName, ImmutableArray<AttributeData> Instances) thisParameterUsageAttr = thisParameterUsage[ j ];
+						(string AttributeName, ImmutableArray<AttributeData> Instances) implementedParameterUsageAttr = implementedParameterUsage[ j ];
 
-						if( thisParameterUsageAttr != implementedParameterUsageAttr ) {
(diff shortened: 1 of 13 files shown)


daryl-mcmillan (dismissed):

src/D2L.CodeStyle.Analyzers/ApiUsage/ConsistentParameterAttributesAnalyzer.cs
```diff
@@ -60,20 +60,26 @@ IMethodSymbol methodSymbol
...
+						(string AttributeName, ImmutableArray<AttributeData> Instances) thisParameterUsageAttr = thisParameterUsage[ j ];
+						(string AttributeName, ImmutableArray<AttributeData> Instances) implementedParameterUsageAttr = implementedParameterUsage[ j ];
+
+						if( thisParameterUsageAttr.Instances.IsDefaultOrEmpty && implementedParameterUsageAttr.Instances.IsDefaultOrEmpty ) {
+							continue;
+						}
+
+						if( thisParameterUsageAttr.Instances.SequenceEqual( implementedParameterUsageAttr.Instances, AttributeDataComparer.Instance ) ) {
+							continue;
 						}
```
omsmith, line 72: This is obviously more expensive than comparing `true/false`, but since `[Constant]` and `[StatelessFunc]` dont' have any arguments, their comparisons should still be fairly free.

Other than that, don't expect `[PatternString]` to come up _that_ often, especially in these interface/base class cases (rather, using it in a `struct` constructor/implicit conversion), so don't expect an extra slowdown from that either.

omsmith, line 72: Note the `SequenceEqual`, it is requiring multiple usages of the attribute would be defined in the same order (which, how that gets returned by the compiler, not strictly guaranteed). Dumb, but easy.

